Issue #3, opened 2016-03-23
Server should not panic when using a non-existent auth provider


cheungpat:
When calling `auth:login` with a non-existent auth provider, skygear-server should not panic

- Skygear Server Date/Version: 0.9.0

#### Actual Results

```
time="2016-03-23T10:24:34Z" level=error msg="/go/src/github.com/oursky/skygear/router/errors.go:61 (0x5602e2)
	errorFromRecoveringPanic: log.Errorf(\"%s\", debug.Stack())
/go/src/github.com/oursky/skygear/router/router.go:101 (0x56621b)
	(*Router).ServeHTTP.func1: resp.Err = errorFromRecoveringPanic(r)
/usr/local/go/src/runtime/asm_amd64.s:437 (0x466ace)
	call32: CALLFN(·call32, 32)
/usr/local/go/src/runtime/panic.go:423 (0x437069)
	gopanic: reflectcall(nil, unsafe.Pointer(d.fn), deferArgs(d), uint32(d.siz), uint32(d.siz))
/usr/local/go/src/runtime/panic.go:42 (0x435729)
	panicmem: panic(memoryError)
/usr/local/go/src/runtime/sigpanic_unix.go:24 (0x44be2a)
	sigpanic: panicmem()
/go/src/github.com/oursky/skygear/handler/auth.go:143 (0x50631a)
	(*SignupHandler).Handle: principalID, authData, err := authProvider.Login(p.AuthData)
/go/src/github.com/oursky/skygear/router/router.go:142 (0x5651d2)
	(*Router).ServeHTTP: handler.Handle(payload, &resp)
/usr/local/go/src/net/http/server.go:1699 (0x49b0ad)
	(*ServeMux).ServeHTTP: h.ServeHTTP(w, r)
/go/src/github.com/oursky/skygear/router/logging.go:103 (0x563670)
	(*LoggingMiddleware).ServeHTTP: l.Next.ServeHTTP(rlogger, r)
/usr/local/go/src/net/http/server.go:1862 (0x49babe)
	serverHandler.ServeHTTP: handler.ServeHTTP(rw, req)
/usr/local/go/src/net/http/server.go:1361 (0x49930e)
	(*conn).serve: serverHandler{c.server}.ServeHTTP(w, w.req)
/usr/local/go/src/runtime/asm_amd64.s:1721 (0x468e21)
	goexit: BYTE	$0x90	// NOP
"
[...]

PR #4, opened 2016-03-23
Check whether auth provider exists

cheungpat:
The name of the auth provider is checked for existence rather panicking.

connects skygeario/skygear-server#3


ben181231:
Picked as 9756437

PR #5, opened 2016-03-29
pubsub: fix last subscriber stealing all published message

limouren:
:dancer:


rickmak:
landed at 7d38c44 for v0.6.1 and 80cff7a for master.

PR #9, opened 2016-04-11
Support public read access ACL

ben181231:
connect oursky/skygear-server#647

### DIFF
--- a/skydb/pq/builder.go
+++ b/skydb/pq/builder.go
@@ -255,6 +255,11 @@ func (p accessPredicateSqlizer) ToSql() (sql string, args []interface{}, err err
 	if p.user.ID == "" {
 		panic("cannot build access predicate without user")
 	}
+	escapedID, err := json.Marshal(p.user.ID)
+	if err != nil {
+		panic("unexpected serialize error on user_id")
+	}
+
 	var b bytes.Buffer
 	b.WriteString(`(`)
 	for _, role := range p.user.Roles {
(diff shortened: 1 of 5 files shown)


cheungpat:
I looked through it and it lgtm but I wonder if @rickmak can have a second look because he knows  better about the ACL

ben181231:
@cheungpat Okay. It should be tested together with the updates on SDKs.

skydb/pq/builder.go
```diff
@@ -262,19 +262,17 @@ func (p accessPredicateSqlizer) ToSql() (sql string, args []interface{}, err err
...
+		b.WriteString(fmt.Sprintf(`_access @> '[{"role": %s}]' OR `, escapedRole))
 	}
-	b.WriteString(`_access @> '[{"user_id":`)
-	escapedID, _ := json.Marshal(p.user.ID)
-	if err != nil {
-		panic("unexpected serialize error on user_id")
-	}
-	b.Write(escapedID)
-	b.WriteString(`}]' OR _access IS NULL OR _owner_id = ?)`)
+	b.WriteString(fmt.Sprintf(`_access @> '[{"user_id": "%s"}]' OR `, p.user.ID))
```
rickmak: `json.Marshal` is used to escaped the `user_id`

If you use fmt.Sprintf and user id is string like `oursky"rickmak"`, the SQL will not able to execute.

rickmak:
Other looks fine to me.

skydb/pq/builder.go
ben181231: Oh. I see.

Issue #23, opened 2016-04-28
Feature Request: logging should use mime.ParseMediaType


louischan-oursky:
logging should use mime.ParseMediaType instead plain string comparison to determine media type.

- Skygear Server Date/Version: all
- Application Platform:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Expected Results
Should be able to log if `Content-Type` is `application/json charset=utf-8`

#### Actual Results
`Content-Type` must be `application/json` otherwise body will not be logged.

PR #24, opened 2016-04-28
router/logging: fix request body not logged if Content-Type contains parameters

limouren:
Connect #23

### DIFF
--- a/router/logging.go
+++ b/router/logging.go
@@ -18,6 +18,7 @@ import (
 	"bufio"
 	"bytes"
 	"io/ioutil"
+	"mime"
 	"net"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


router/logging.go
```diff
@@ -103,9 +104,18 @@ func (l *LoggingMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 }
+
+func isBodyLoggable(h http.Header) bool {
+	contentType := h.Get("Content-Type")
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		return false
+	}
+	return mediaType == "" || mediaType == "application/json"
+}
```
rickmak: For logic that only usable to `LoggingMiddleware` I would prefer you put it on struct.

And it will be great this logger can use use as following:

```
finalMux = &router.LoggingMiddleware{
	Skips: []string{
		"/files/",
		"/_/pubsub/",
		"/pubsub/",
	},
	MimeConcern: []string{
		"application/json"
	},
	Next: finalMux,
}
```

rickmak:
I am sending another PR regarding this: https://github.com/SkygearIO/skygear-server/pull/25

router/logging.go
```diff
@@ -103,9 +104,18 @@ func (l *LoggingMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	l.Next.ServeHTTP(rlogger, r)
 
 	log.Debugln("------ Response: ------")
-	if !skipBody && (w.Header().Get("Content-Type") == "" || w.Header().Get("Content-Type") == "application/json") {
+	if !skipBody && isBodyLoggable(r.Header) {
```
rickmak: btw, here should be `isBodyLoggable(w.Header())`

PR #25, opened 2016-05-03
Use mime package for mime processing and allow config of mime type concern

rickmak:
connects #23

### DIFF
--- a/router/logging.go
+++ b/router/logging.go
@@ -18,6 +18,7 @@ import (
 	"bufio"
 	"bytes"
 	"io/ioutil"
+	"mime"
 	"net"
 	"net/http"
 	"strings"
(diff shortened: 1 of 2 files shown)


router/logging.go
```diff
@@ -103,9 +105,24 @@ func (l *LoggingMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		return false
+	}
+	concern := false
+	for _, c := range l.MimeConcern {
+		if mediaType == c {
+			concern = true
+			break
+		}
+	}
+	return concern
```
cheungpat: Why not simply `return false` here? (and remove the `concern` variable)

rickmak:
Updated to remove the variable.

Issue #15, opened 2016-04-15
HTTP headers from plugin handlers are not written to client


royuen:
My plugin handler returns this:

```
        resp = Response(json.dumps(output))
        resp.headers.add('X-Total-Count', 100)
        return resp
```

This appears in the log between skygear-server and plugin

```
DEBU[0406] Called  ./run_plugin.sh [./run_plugin.sh --subprocess handler spaces] returning: {"result": {"header": {"Content-Length": ["14029"], "X-Total-Count": ["100"]...
```

But the header is not written to client:

```
< HTTP/1.1 200 OK
< Access-Control-Allow-Credentials: true
< Access-Control-Allow-Origin: *
< Date: Fri, 08 Apr 2016 04:16:29 GMT
< Content-Type: text/plain; charset=utf-8
< Transfer-Encoding: chunked
```

I think this is because this is how handler is implemented in `plugin/handler.go`:

```
	response.Meta = responsePayload.Header
	response.WriteHeader(responsePayload.Status)  # headers written out
	response.Write(responsePayload.Body)  # headers from the plugin added to `http.ResponseWriter` here.
```

In other words, the headers from plugin are processed after the response writer has already written out the headers.

PR #27, opened 2016-05-05
Fix missing headers returned from plugins

ben181231:
connect #15


cheungpat:
please rebase this

cheungpat:
Need test case

ben181231:
Test case added

Issue #6, opened 2016-03-30
Support integer column


limouren:
e.g. we have a record "food" which has an integer column "tip":

```
               Table "app.food"
     Column     |            Type             | Modifiers 
----------------+-----------------------------+-----------
 _id            | text                        | not null
 tip            | integer                     | 
```

I want CRUD to works correctly on "food" records. 

The state of affairs is fetch, save and delete works correctly except that for save - when I am trying to supply a value on that integer column, I got an error.

## cmd

```bash
> curl -X POST -H "Content-Type: application/json" \
  -d @- http://localhost:8080/api <<EOF
{
  "api_key": "API_KEY",
  "database_id": "_public",
  "action": "record:save",
  "access_token": "4075602a-52a0-49d7-95a6-675fa514e319",
  "records": [
    {
      "_id": "food/4c5f6213-a022-4cb4-ad8d-bd85d6282eb5",
      "tip": 999
    }
  ]
}
EOF
{"result":[{"_id":"food/4c5f6213-a022-4cb4-ad8d-bd85d6282eb5","_type":"error","code":10000,"message":"pq: relation \"APP.food_tip_seq\" does not exist","name":"UnexpectedError"}]}
```

## Server Log

```
DEBU[92908] POST /                                       
DEBU[92908] ------ Header: ------                        
DEBU[92908] Accept: [*/*]                                
DEBU[92908] Content-Type: [application/json]             
DEBU[92908] X-Real-Ip: [127.0.0.1]                       
DEBU[92908] X-Forwarded-For: [127.0.0.1]                 
DEBU[92908] Connection: [close]                          
DEBU[92908] Content-Length: [260]                        
DEBU[92908] User-Agent: [curl/7.43.0]                    
DEBU[92908] ------ Request: ------                       
DEBU[92908] {  "api_key": "API_KEY",  "database_id": "_public",  "action": "record:save",  "access_token": "4075602a-52a0-49d7-95a6-675fa514e319",  "records": [    {      "_id": "food/4c5f6213-a022-4cb4-ad8d-bd85d6282eb5",      "tip": 999    }  ]} 
DEBU[92908] Opening DBConn: {pq com_ufo_api postgres://127.0.0.1/ufo?sslmode=disable} 
DEBU[92908] Get DB OK                                    
WARN[92908] 55d4a56d-87a3-4e13-b9b8-7153ed9325cf         
[...]

PR #28, opened 2016-05-05
Check sequence exist before update

ben181231:
connect #6


cheungpat:
It looks okay, but it seems checking whether an integer column has a sequence each time a record is saved seems expensive, especially when saving multiple records of the same type.

I think it is better to check whether it is a sequence at the time when the record schema is determined.

PR #29, opened 2016-05-06
Fix test case fails on Travis

ben181231:


ben181231:
The test case failure is caused by [this commit](https://github.com/google/go-gcm/commit/3fd314a898b95cb8d1f846ef3fc56c81f8d3fc9e).

Issue #22, opened 2016-04-27
Feature request: Ignoring ACL when master key is used


cheungpat:
It is possible to use master key to impersonate another user using the `_user_id`, but ACL would still apply. The use case is that a maintenance program will want to update records on behalf of a user, but does not want to be restricted by ACL.

PR #30, opened 2016-05-06
Allow master key to override ACL restriction

ben181231:
connect #22


rickmak:
The checking should be done at handler.

PR #31, opened 2016-05-09
Allow master key to override ACL restriction

ben181231:
connect #22

### DIFF
--- a/handler/record.go
+++ b/handler/record.go
@@ -298,7 +298,7 @@ func (h *RecordSaveHandler) Handle(payload *router.Payload, response *router.Res
 		UserInfo:      payload.UserInfo,
 		RecordsToSave: p.Records,
 		Atomic:        p.Atomic,
-		WithMasterKey: payload.HasMasterAccess(),
+		WithMasterKey: payload.HasMasterKey(),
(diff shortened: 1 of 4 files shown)


handler/record.go
```diff
@@ -298,6 +298,7 @@ func (h *RecordSaveHandler) Handle(payload *router.Payload, response *router.Res
 		UserInfo:      payload.UserInfo,
 		RecordsToSave: p.Records,
 		Atomic:        p.Atomic,
+		WithMasterKey: payload.HasMasterAccess(),
```
cheungpat: On one hand you have a the `router` package hide the fact that the user has master key and expose whether the user has master access, but later in the handler, the fact that the user has master access is converted to say the user has a master key.

If having master access does not mean the user have master key, this line is wrong.
If having master access means the user have master key, there is no need for such distinction, and `HasMasterAccess` should be called `HasMasterKey` instead

ben181231: Got it.

ben181231: I will rename it to `HasMasterKey`

ben181231:
@cheungpat Updated

PR #32, opened 2016-05-10
Consider deleting non-existing device as success

louischan-oursky:
If we think deletion is an idempotent operation, then deleting non-existing device should be considered as success. It is because both result in a desired state, that is the device will no longer exist after the statement is executed.


rickmak:
By analysis the git praise, I suggest @limouren to be the reviewer.

limouren:
LGTM. Any objections from others?

rickmak:
Merge via 0bda796

PR #57, opened 2016-06-28
Allow master key to bypass record ACL in query

cheungpat:
The was work earlier (#22) that bypass ACL when used with master key,
but that work only apply when saving, deleting and fetching
records. The query action still enforce ACL though.

This commit leave skydb.Query.ReadableBy as empty struct
when the request is authenticated with a master key.

connects #51


ben181231:
Picked as e2727c3

PR #62, opened 2016-06-28
Implement union database

cheungpat:
UnionDatabase is a database containing all records in the PublicDatabase
and all PrivateDatabase. This database is only intended for admin user and
ACL settings do not apply.

The preprocessor is also modified to allow client to select a union database. A private database can be selected by ID if the client is authenticated with master key.


cheungpat:
@rickmak This pull request is updated to support the union database and inject db preprocessor update

Issue #47, opened 2016-06-03
Looping on sending push notification may result in some push notification missing


royuen:
Case 1:
Put a for loop to send push notification in cloud code. Some push notification may be missed.

Case 2:
If a user has a lot of device tokens, some of them may be missed when we send push notification to that user

cheungpat:
I still think switching to the new HTTP/2 protocol/library is the way to go. At least you get response from APNS.

I think this might be relevant to the problem you are seeing
https://github.com/timehop/apns/issues/70

PR #68, opened 2016-07-06
Use HTTP/2 APNS protocol

cheungpat:
This commit switch the underlying APNS library to
RobotsAndPencils/buford, which sends push notification to APNS
through HTTP/2 protocol. The new protocol provides immediate feedback
for each notification sent.

connects #47

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,7 +1,13 @@
 {
-	"ImportPath": "github.com/SkygearIO/skygear-server",
-	"GoVersion": "go1.5",
+	"ImportPath": "github.com/skygeario/skygear-server",
+	"GoVersion": "go1.6",
+	"GodepVersion": "v74",
 	"Deps": [
+		{
+			"ImportPath": "github.com/RobotsAndPencils/buford/push",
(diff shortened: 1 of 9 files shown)


cheungpat:
To try this out, pull from `quay.io/skygeario/skygear-server:new-push` which uses the new HTTP/2 APNS protocol. Note that this version is based on 0.13.0.

cc @rockychan @royuen

cheungpat:
@rockychan Please pull again for the `new-push` image, which supports APNS certificate containing multiple topics. There is no configuration options for topic name as skygear-server will use the UID in the APNS certificate Subject as the default topic name.

cheungpat:
@rickmak updated, see if you have any comment

cheungpat:
@rickmak: @rockychan reported that he encountered a skygear-server panic in `DeleteDeviceByToken` which I think could occurs when:

* a device token is registered to multiple users (which is permitted by database schema unique constraint, because that constraint is a tuple of `user_id, type, token`)
* APNS reported the device token is bad or is unregistered

`DeleteDeviceByToken` check whether the number of deleted device is 1. I removed this check to avoid the panic.

Issue #70, opened 2016-07-07
User Query does not include the roles


ben181231:
When querying user using `user:query`, currently the response does not include the user's roles:
```
{
  "result": [
    {
      "data": {
        "_id": "cd612328-3cf3-421f-af0e-474e06f6a9ac",
        "email": "user4@skygear.dev",
        "username": ""
      },
      "id": "cd612328-3cf3-421f-af0e-474e06f6a9ac",
      "type": "user"
    }
  ]
}
```

royuen:
- [ ] Change user query to be only called by admin
- [ ] Return user role as well

PR #72, opened 2016-07-11
Include user role in user query

ben181231:
connect #70


cheungpat:
User is not really required here. Therefore the test should be:

* if client has master key, allow request
* if client is authenticated and the `UserInfo` has admin roles, allow request
* otherwise, deny request

ben181231:
Okay. I will add some test cases.

PR #77, opened 2016-07-15
Add Stateless token store

cheungpat:
connects #74

### DIFF
--- a/authtoken/authtoken.go
+++ b/authtoken/authtoken.go
@@ -150,6 +150,7 @@ type Configuration struct {
 	Path           string
 	Prefix         string
 	Expiry         int64
+	Secret         string
 }
 
 // InitTokenStore accept a implementation and path string. Return a Store.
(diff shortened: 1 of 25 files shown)


cheungpat:
Do not merge yet, I am trying to add a test case for user info with the token valid since field

cheungpat:
Updated the pull request with the added test case

authtoken/authtoken.go
```diff
@@ -162,6 +178,8 @@ func InitTokenStore(config Configuration) Store {
 		store = NewFileStore(config.Path, config.Expiry)
 	case "redis":
 		store = NewRedisStore(config.Path, config.Prefix, config.Expiry)
+	case "stateless":
```
rickmak: `jwt`? Seem expose the implementation to developer will save use from explanation on how we do the stateless store.

authtoken/stateless.go
```diff
@@ -0,0 +1,115 @@
...
+		return Token{}, err
+	}
+
+	token := Token{}
+	r.setTokenFromClaims(claims, &token)
+	token.AccessToken = signedString
+	return token, nil
+}
+
+// Get tries to read the specified access token from redis store and
```
rickmak: `redis` store?

cheungpat:
@rickmak updated the PR

PR #81, opened 2016-07-20
Refactor to make sqlizer creation consistent

cheungpat:


cheungpat:
I am going to post a different PR

Issue #56, opened 2016-06-28
record:query does not check if record field exists


cheungpat:
When `record:query` is called with a query having a non-existent field, the record query will return an `UnexpectedError`:

``` javascript
const Note = skygear.Record.extend("note");
var query = new skygear.Query(Note);
query.equalTo("id", "note/d5a02b77-cc7a-4f93-8cb9-4806b3762767");
skygear.publicDB.query(query).then(...);
```

This code would produce the following error message.

``` javascript
{
    status: 500,
    error: {
        code: 10000,
        message: "pq: column note.id does not exist",
        name: "UnexpectedError"
    }
}
```

There are two choices here:

* Since we have JIT record schema, one way to do it is to return empty array of records whenever a query includes non-existent fields.
* Treat the query as bad data, therefore return an error. We can pick `BadRequest` (not preferred), `InvalidArgument` (preferred), or we can create a new kind of error like `InvalidQuery`.

Thanks for @TakLee96 for reporting the issue in skygeario/skygear-SDK-JS#22.

PR #82, opened 2016-07-21
Check field exists before performing query

cheungpat:
connects skygeario/skygear-server#56

### DIFF
--- a/handler/asset.go
+++ b/handler/asset.go
@@ -178,7 +178,7 @@ func (h *AssetUploadURLHandler) Handle(payload *router.Payload, response *router
 
 	written, tempFile, err := copyToTempFile(payload.Req.Body)
 	if err != nil {
-		response.Err = skyerr.NewUnknownErr(err)
+		response.Err = skyerr.MakeError(err)
 		return
 	}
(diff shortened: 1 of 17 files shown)


rickmak:
This PR is not working with `in` predicate

Input:
```
{
    "action": "record:query",
    "access_token": "token",
    "database_id": "_private",
    "record_type": "note",
    "predicate": [
        "in",
        "bye",
        {
          "$type": "keypath",
          "$val": "404"
        }
    ]
}
```

Output: 
```
HTTP/1.1 500 Internal Server Error
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: *
Connection: keep-alive
Content-Length: 97
Content-Type: application/json
Date: Thu, 21 Jul 2016 11:17:08 GMT
Server: nginx/1.8.1

{
    "error": {
        "code": 10000, 
        "message": "pq: column note.404 does not exist", 
        "name": "UnexpectedError"
    }
}
```

cheungpat:
Right, the contains comparison sqlizer doesn’t create the expression sqlizer using the new function, which it should.

Issue #33, opened 2016-05-17
NULL token saved to DB causing type error on send


rickmak:
For unknown reason, the a NULL token is saved to `_device` table. And causing type error on sending push notification. 

- Skygear Server Date/Version: v0.10
- Application Platform: portal user

#### Expected Results
The NULL will got rejected during save time.

#### Actual Results
Got following error on calling `push:users`

```
{
  "error": {
    "name": "UnexpectedError",
    "code": 10000,
    "message": "panic occurred while handling request: sql: Scan error on column index 2: unsupported driver -> Scan pair: <nil> -> *string"
  }
}
```

cheungpat:
It appears that skydb save `skydb.Device` with NULL `token` if the request’s `device_token` field is empty. This is [explicitly done this way](https://github.com/SkygearIO/skygear-server/blob/master/skydb/pq/device.go#L97). There is also test case that explicitly test that the handler [accepts device with empty token](https://github.com/SkygearIO/skygear-server/blob/master/handler/device_test.go#L146).

Therefore, my opinion is to keep existing behavior and not reject device with empty token.

PR #83, opened 2016-07-21
Fix scan NULL token in QueryDevicesByUser

cheungpat:
If `device` table contains device with NULL token, Scan will complain
 that it is not able to scan a NULL value to a type of string.

connects #33


rickmak:
Merge via 31a1ce9

Issue #53, opened 2016-06-22
Multiple schema in same postgresql will result migration fails on citext


rickmak:
Try to use two schema under same DB, but it fails on create citext extension.

- Skygear Server Date/Version: v0.12
- Application Platform:
- [X] Is this a regression?

#### Expected Results
One database will allow multiple skygear schema working.

#### Actual Results

```
ERRO[0001] Error executing schema migration "" -> "c0397f15027": pq: type "citext" does not exist 
ERRO[0001] Failed to start skygear: failed to open connection: skydb/pq: unable to migrate database because of an error = skydb/pq: failed to init database: pq: type "citext" does not exist 
```
#### Steps to reproduce

1. init the skygear-server with `APP_NAME=first`, skygear-server will do the database migration correctly.
2. Terminate skygear-server
3. Change the `APP_NAME` config to `second`
4. Try to kick start the skygear-server again, skygear-server will raise migration error

cheungpat:
It appears that extension is created specific to schema. Therefore we might need to specify the schema when creating the extension.

Many examples on the web do this, where `public` is the name of the schema.

```
CREATE EXTENSION IF NOT EXISTS citext WITH SCHEMA public;
```

Quoting [documentation on `CREATE EXTENSION`](https://www.postgresql.org/docs/9.1/static/sql-createextension.html) (emphasis mine):

> The name of the schema in which to install the extension's objects, given that the extension allows its contents to be relocated. The named schema must already exist. **If not specified, and the extension's control file does not specify a schema either, the current default object creation schema is used.**

PR #88, opened 2016-07-22
Create the citext to public schema

rickmak:
I modify the migration directly, since current deployment is one schema per
db. I think it will be okay for just make it works better in new deployment.
connect #53


rickmak:
Moved the `CREATE EXTENSION postgis;` to migration too.

PR #89, opened 2016-07-26
Handle invalid data format for pubsub actions

ben181231:
connect SkygearIO/skygear-SDK-JS#27

### DIFF
--- a/pubsub/ws.go
+++ b/pubsub/ws.go
@@ -128,6 +128,15 @@ func (w *WsPubSub) reader(c *connection) {
 			c.ws.WriteMessage(websocket.CloseMessage, nil)
 			return
 		}
+		if payload.Channel == "" {
+			log.Debugf("Got empty channel.")
+			c.ws.WriteMessage(
+				websocket.TextMessage,
+				[]byte("Error: channel should not be empty. Closing Connection"),
+			)
+			c.ws.WriteMessage(websocket.CloseMessage, nil)
+			return
+		}
 		switch payload.Action {
 		case "sub":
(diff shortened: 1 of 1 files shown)


pubsub/ws.go
```diff
@@ -142,9 +151,18 @@ func (w *WsPubSub) reader(c *connection) {
...
-				Channel: payload.Channel,
-				Data:    []byte(*payload.Data),
+			if payload.Data == nil {
+				log.Debugf("Got nil pub data.")
+				c.ws.WriteMessage(
+					websocket.TextMessage,
+					[]byte("Error: missing data to publish. Closing Connection"),
+				)
+				c.ws.WriteMessage(websocket.CloseMessage, nil)
+			} else {
```
cheungpat: I think the convention in golang is for error scenario to return early so the normal path has a lower indentation level.

ben181231: Oh. I see.

pubsub/ws.go
```diff
@@ -142,6 +151,15 @@ func (w *WsPubSub) reader(c *connection) {
...
 			c.channels = append(c.channels, payload.Channel)
 		case "pub":
+			if payload.Data == nil {
+				log.Debugf("Got nil pub data.")
+				c.ws.WriteMessage(
+					websocket.TextMessage,
+					[]byte("Error: missing data to publish. Closing Connection"),
+				)
+				c.ws.WriteMessage(websocket.CloseMessage, nil)
+				break
```
cheungpat: this need to be `return`

Issue #93, opened 2016-07-28
Locking down schema in production mode


cheungpat:
In `record:save`, the schema is automatically extended when there is a new field. This issue should lock down the schema when in production mode (i.e. `dev-mode = NO`)

* should not allow client to change schema when not in dev mode
* the above rule is bypassed when the request is using the master key

PR #97, opened 2016-07-28
Move record-related utilities into separate file

cheungpat:


cheungpat:
This is related to #93

cheungpat:
I am assuming that @rickmak is okay with this.

PR #98, opened 2016-07-28
Disable schema migration for non-dev-mode

cheungpat:
Requests authenticated with master key can temporarily override
this restriction.

connects #93


rickmak:
ping @cheungpat travis fails

cheungpat:
@rickmak fixed

Issue #75, opened 2016-07-13
Provide a way to initial the first admin role


rickmak:
When the developer try to use the role base ACL, he will need to setup the first admin role and admin users in the database in order to make it works.

- Skygear Server Date/Version: v0.13 13 July

rickmak:
Solution direction:
- Create admin user and role on first migration
- Or only do it at Cloud deployment
- First user as admin

PR #101, opened 2016-08-01
Add migration for initial admin user

ben181231:
connect #75


cheungpat:
@ben181231 looks like needing doc update to warn user that there is a default admin account

i would say it is okay to put the migration step at the bottom... I am not sure of the reasons of putting it in the middle of the migration steps.

ben181231:
@cheungpat 
Oh. I will update the documentation.
The reason why I put the migration in the middle is that I don't want to affect the existing users.

limouren:
Being one of the affected users, I would prefer new migrations to be made to rectify past migration issues (even the script the script is incorrect). Now I have to review all revisions w.r.t the DB and make corresponding changes by hand.

Issue #107, opened 2016-08-09
Support Cloud Asset upload


ben181231:
### Tasks for Skygear-Server
- [x] Load config from env.
- [x] Request for signer token
- [x] Refresh signer token periodically
- [x] New endpoint `asset:put` for new upload protocol

### Tasks for SDKs
- [x] Support on iOS SDK
- [x] Support on JS SDK
- ~~Support on Android SDK~~

### Tasks for Others
- [ ] Update documentation

rickmak:
Pending: ASSET_STORE_HOST should default to HOST if not specific explicitly.

PR #119, opened 2016-08-30
Support Cloud Asset

ben181231:
connect #107

### DIFF
--- a/pkg/server/asset/cloud.go
+++ b/pkg/server/asset/cloud.go
@@ -20,6 +20,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"io"
+	"net/http"
 	"net/url"
 	"strconv"
 	"strings"
@@ -170,6 +171,47 @@ func (s CloudStore) PutFileReader(
 	return errors.New(
 		"Directly uploading files is not available for cloud-based asset store",
 	)
+}
+
+// GeneratePostFileRequest return a PostFileRequest for uploading asset
+func (s CloudStore) GeneratePostFileRequest(name string) (*PostFileRequest, error) {
+	log.
+		WithField("name", name).
+		Info("Start generate post file request for Cloud Asset")
+
+	urlString := strings.Join(
+		[]string{s.host, "asset", s.appName, name},
+		"/",
+	)
+
+	req := goreq.Request{
+		Method:  http.MethodPut,
+		Uri:     urlString,
+		Timeout: 10 * time.Second,
+	}.WithHeader("Authorization", "Bearer "+s.authToken)
+
+	res, err := req.Do()
+	if err != nil {
+		log.
+			WithField("url", urlString).
+			WithField("error", err).
+			Error("Fail to request for pre-signed POST request")
+
+		return nil, errors.New("Fail to request for pre-signed POST request")
+	}
+
+	postRequest := &PostFileRequest{}
+	err = res.Body.FromJsonTo(postRequest)
+	if err != nil {
+		log.
+			WithField("error", err).
+			WithField("response", res.Body).
+			Error("Fail to parse the response of pre-signed POST request")
+
+		return nil, errors.New("Fail to parse the response of pre-signed POST request")
+	}
+
+	return postRequest, nil
 }
 
 // SignedURL return a signed URL with expiry date
--- a/pkg/server/asset/cloud.go
+++ b/pkg/server/asset/cloud.go
@@ -15,16 +15,21 @@
 package asset
 
 import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/base64"
 	"errors"
 	"io"
+	"net/url"
+	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/franela/goreq"
 )
 
 const (
+	cloudAssetURLExpiryInterval          = 15 * time.Minute
 	cloudAssetSignerTokenRefreshInterval = 30 * time.Minute
 	cloudAssetSignerTokenExpiryInterval  = 2 * time.Hour
 )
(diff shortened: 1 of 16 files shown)


pkg/server/asset/cloud.go
```diff
@@ -0,0 +1,311 @@
...
+		Timeout: 10 * time.Second,
+		QueryString: struct {
+			ExpiredAt int64 `url:"expired_at"`
+		}{expiredAt},
+	}.WithHeader("Authorization", "Bearer "+s.authToken)
+
+	res, err := req.Do()
+	if err != nil {
+		log.
+			WithField("url", urlString).
```
cheungpat:
```
log.(logrus.WithFields{
}).Error(/* ... */)
```

pkg/server/asset/cloud.go
```diff
@@ -0,0 +1,218 @@
...
+
+// PutFileReader return a writer for uploading files
+func (s CloudStore) PutFileReader(
+	name string,
+	src io.Reader,
+	length int64,
+	contentType string,
+) error {
+	return errors.New(
+		"Directly upload files is not available for cloud-based asset store",
```
cheungpat: s/upload/uploading/

pkg/server/asset/cloud.go
```diff
@@ -0,0 +1,218 @@
...
+
+// GetFileReader returns a reader for reading files
+func (s CloudStore) GetFileReader(name string) (io.ReadCloser, error) {
+	return nil, errors.New(
+		"Directly getting files is not available for cloud-based asset store",
+	)
+}
+
+// PutFileReader return a writer for uploading files
+func (s CloudStore) PutFileReader(
```
cheungpat: I am not fond of the style to have every parameter on separate line.

Are you aiming at a line-width limit?

pkg/server/asset/cloud.go
```diff
@@ -0,0 +1,218 @@
...
+// NewCloudStore creates a new cloud asset store
+func NewCloudStore(
+	appName string,
+	host string,
+	authToken string,
+	publicURLPrefix string,
+	privateURLPrefix string,
+	public bool,
+) (*CloudStore, error) {
+	if appName == "" {
```
cheungpat: I think you could just panic... the reason being that passing wrong parameters are programming error and there is no way to handle such error.

pkg/server/asset/cloud.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/franela/goreq"
+)
+
+const (
+	cloudAssetSignerTokenRefreshInterval = 30 * time.Minute
+	cloudAssetSignerTokenExpiryInterval  = 2 * time.Hour
+)
+
+// CloudStore models the skygear cloud asset store
+type CloudStore struct {
```
cheungpat: Since this struct only implements other interface, you might just set this struct private.

pkg/server/asset/cloud.go
```diff
@@ -0,0 +1,218 @@
...
+			Error("Fail to parse the response for refresh Cloud Asset Signer Token")
+
+		return
+	}
+
+	log.
+		WithField("response", resBody).
+		Info("Successfully got new Cloud Asset Signer Token")
+
+	s.signerTokenMutex.Lock()
```
cheungpat: It is idiomatic to defer s.signerTokenMutex.Unlock here.

pkg/server/asset/cloud.go
```diff
@@ -0,0 +1,218 @@
...
+// CloudStore models the skygear cloud asset store
+type CloudStore struct {
+	appName   string
+	host      string
+	authToken string
+	urlPrefix string
+	public    bool
+
+	// signer related
+	signerToken              string
```
cheungpat: You could have created another struct just for signing... there are a few variables that seem signer related and not relevant to rest of the struct?

pkg/server/asset/cloud.go
```diff
@@ -189,15 +195,60 @@ func (s CloudStore) PutFileReader(
...
+			Error("Fail to parse the unsigned URL")
+
+		return "", errors.New("Fail to parse the unsigned URL")
+	}
+
+	if !s.IsSignatureRequired() {
+		return targetURL.String(), nil
+	}
+
+	s.signerTokenMutex.RLock()
```
cheungpat: put this in its own function so that you can call `RUnlock` deferred

pkg/server/handler/asset.go
```diff
@@ -133,49 +139,72 @@ func (h *AssetGetURLHandler) Handle(payload *router.Payload, response *router.Re
...
+//    http://localhost:3000/files/filename
+//
 type AssetUploadURLHandler struct {
 	AssetStore    skyAsset.Store   `inject:"AssetStore"`
 	AccessKey     router.Processor `preprocessor:"accesskey"`
 	DBConn        router.Processor `preprocessor:"dbconn"`
 	preprocessors []router.Processor
 }
 
+type assetUploadRequest struct {
```
cheungpat: this struct does not belong here, move it elsewhere (above struct definition or below all functions defined for a struct

pkg/server/handler/asset.go
```diff
@@ -218,6 +247,61 @@ func (h *AssetUploadURLHandler) Handle(payload *router.Payload, response *router
...
+		if err != nil {
+			return nil, err
+		}
+	} else if method == http.MethodPut {
+		filename = clean(payload.Params[0])
+		contentType = httpRequest.Header.Get("Content-Type")
+		fileReader = httpRequest.Body
+	} else {
+		return nil, errors.New(
+			"Method " + method + " is not supported",
```
cheungpat: `fmt.Errorf(/*...*/)`

ben181231:
@cheungpat @rickmak The PR updated.

rickmak:
ASSET_STORE_HOST should default to HOST if not specific explicitly. other :100:

ben181231:
@rickmak `HOST` is set to `localhost:3000` on `.env.sample` which will not work if it is directly set to `ASSET_STORE_HOST`.

Issue #105, opened 2016-08-03
Handle pubsub messages in multi instance scenario


cheungpat:
To fix #91, we have to handle how pubsub messages are routed to the client. To solve this, the leader instance will be asked to keep pubsub connections for all clients. Pubsub generated from other instances should forward pubsub messages to the leader instance.

Tasks:

* add new configuration envvar for specifying the leader pubsub URL
* forward pubsub messages generated from the instance to the leader (which includes subscriptions and cloudcode-generated messages)
* leader should be aware that the messages coming from the non-leaders are generated server side and not generated client side

rickmak:
Revisiting this issue, in https://github.com/SkygearIO/skygear-server/pull/127 we added the configuration on slave and disabled the whole pubsub and cron feature at slave mode.

So the do @cheungpat think following two feature are still meaningful to implement
> 
- forward pubsub messages generated from the instance to the leader (which includes subscriptions and cloudcode-generated messages)
- leader should be aware that the messages coming from the non-leaders are generated server side and not generated client side

Or better tol provide a deployment example on how to route the traffic correctly using the program like nginx?

cheungpat:
@rickmak Thanks for bringing it to my attention.

I think the feature is complete in skygear-server and py-skygear to support multiple instances. Agree that we should have deployment example. Where to put it?

> forward pubsub messages generated from the instance to the leader (which includes subscriptions and cloudcode-generated messages)

Cloudcode generated messages are forwarded to leader using PUBSUB_URL. In slave mode, skygear-server does not generate notifications for subscriptions

> leader should be aware that the messages coming from the non-leaders are generated server side and not generated client side

This is just a reminder not to cause infinite loop forwarding notifications among multiple instances. Since only the leader will act as pubsub hub. This is a non-issue.

rickmak:
Let put the example here: https://github.com/skygear-demo/multiple-server

Issue #104, opened 2016-08-03
Disable cron job for non-leader instance


cheungpat:
To fix #91, a configuration will be added (#103) which designate an instance as leader or non-leader. Non-leader should disable cron job, so that only one instances (the leader) will be triggering cron job.

Issue #103, opened 2016-08-03
Add configuration for specifying leader instance


cheungpat:
The fastest way to fix #91, is to designate an instance as a leader instance among multiple running skygear-servers. Features may be enabled/disabled depending on whether the instance is the leader.

Environment variable suggestion: An instance is a leader if an environment variable is set as `IS_LEADER=1` (or equal to other boolean true value such as `yes`, `true` etc)

PR #127, opened 2016-09-01
Implement slave mode

cheungpat:
In slave mode, features that are not multiple-instance-safe are diabled.
These features include pubsub, subscription and cron job. Running them
in multiple-instance environment will cause data loss (pubsub messages sent
to an instance not delivered to all clients) or unexpected behaviour (cron jobs
run more than once).

Slave mode can be activated with environment variable `SLAVE=1/YES/true`
(default is false). 

connects skygeario/skygear-server#103
connects skygeario/skygear-server#104
connects skygeario/skygear-server#105

### DIFF
--- a/main.go
+++ b/main.go
@@ -70,6 +70,10 @@ func main() {
 	log.Infof("Starting Skygear Server(%s)...", skyversion.Version())
 	connOpener := ensureDB(config) // Fatal on DB failed
 
+	if config.App.Slave {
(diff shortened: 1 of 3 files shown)


cheungpat:
Just realised that `PUBSUB_URL` is deprecated in py-skygear

rickmak:
Provide Slave default at `NewConfiguration`?

cheungpat:
rebased @rickmak

PR #129, opened 2016-09-02
Update README to match Makefile

rickmak:


cheungpat:
cherry-picked as 485f6c2

PR #130, opened 2016-09-02
Add `make test`

rickmak:


Makefile
```diff
@@ -25,6 +25,10 @@ vendor:
 build:
 	$(DOCKER_RUN) go build -o $(DIST) $(GO_BUILD_ARGS)
 
+.PHONY: test
+test:
+	$(DOCKER_RUN) go test ./pkg/...
```
cheungpat: replace `./pkg/...` with `$(glide nv)`

cheungpat: but I actually think yours is better because glide may not be installed with go only lives inside docker

cheungpat:
cherry-picked as ea4cee9

PR #131, opened 2016-09-02
Update amz.v3 for signing s3 asset with UTC

rickmak:


cheungpat:
cherry-picked as b3f133c

PR #141, opened 2016-09-19
Allow db connection to start new transaction

cheungpat:
Previous implementation does not allow connection to start
a new transaction after a previously committed or rolled back transaction.
This scenario is hit in auth:login using an auth provider because UserInfo
is updated after creating a new user.


oursky-travis:

cheungpat:
@lampercy thank you for notifying me of this bug

rickmak:
Merge via 111c27f

Issue #145, opened 2016-09-27
Docker-compose up fails on first time


rickmak:
(user) got a problem when I run `docker-compose -f docker-compose.yml up` of `skygear-server` first time

```
p_1    | time="2016-09-26T04:17:23Z" level=error msg="Unable to begin transaction for schema migration: dial tcp 172.18.0.3:5432: getsockopt: connection refused" logger=skydb
app_1    | time="2016-09-26T04:17:23Z" level=error msg="Failed to start skygear: failed to open connection: skydb/pq: unable to connect to database because of a network error = dial tcp 172.18.0.3:5432: getsockopt: connection refused"
```

- Skygear Server Date/Version: 0.17.1
- Application Platform:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Expected Results
First run will work.

#### Actual Results
Second run will work.

#### Steps to reproduce

1. Checkout skygear-server
2. `docker-compose -f docker-compose.yml up`

cheungpat:
This is not a regression.

cheungpat:
There are a couple of solutions to this problem:

* change the docker image so that skygear-server will restart on fails
* introduce a delay before skygear-server is run
* skygear-server wait for longer before giving up

pick one

PR #146, opened 2016-09-27
Mark the skygear-server restart to always and remove redis

rickmak:
It will solve the first time init fails. The skygear-server will restart until
it works and user will see it up and running after the DB is finished init.

connect #145

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,12 +12,7 @@ services:
     volumes:
     - db_data:/var/lib/postgresql/data
     ports:
-    - "5432:5432"
-
-  redis:
-    image: redis:3.0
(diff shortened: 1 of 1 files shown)


cheungpat:
I don’t like it, you are not solving the real problem. But I will be merging it anyway.

cheungpat:
@rickmak removing redis from the docker-compose.yml affected my ability to run go test (which run test on redis authtoken)

rickmak:
I found the `make test` is not working when `WITH_DOCKER` it set too.

cheungpat:
if you are saying “is not working” means not able to run test suites related to redis, we are on the same page.

PR #147, opened 2016-09-27
Remove deprecated requirements.txt

rickmak:
It should no longer useful since this repo have no python deps now.


cheungpat:
cherry-picked as 8f0e4dd

Issue #154, opened 2016-10-06
Update the image to include ca-certificate


rickmak:
apline linux don't include root ca-certificate by default.

refs: https://github.com/docker-library/golang/issues/97
refs: https://bugs.alpinelinux.org/issues/5821

- Skygear Server Date/Version: 0.16.0
- Application Platform:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Expected Results
HTTPS works.

#### Actual Results
`x509: failed to load system roots and no roots provided`

PR #155, opened 2016-10-06
Add ca-certificate to deps building pipeline

rickmak:
connects #154


cheungpat:
the deps docker image is updated on Docker Hub and Quay.io

PR #161, opened 2016-10-13
Fix fast restart of skygear-server result on worker stale

rickmak:
zmq broker don't accept non register plugin heartbeat. It will force the plugin
resend a proper Ready and go through the init process again.

refs #150

### DIFF
--- a/pkg/server/plugin/zmq/broker.go
+++ b/pkg/server/plugin/zmq/broker.go
@@ -65,6 +65,7 @@ type Broker struct {
 	backend       *goczmq.Sock
 	bothPoller    *goczmq.Poller
 	backendPoller *goczmq.Poller
+	workers       workerQueue
 	freshWorkers  chan []byte
 	logger        *logrus.Entry
 }
@@ -98,6 +99,7 @@ func NewBroker(name, frontendAddr, backendAddr string) (*Broker, error) {
 		backend:       backend,
 		bothPoller:    bothPoller,
 		backendPoller: backendPoller,
+		workers:       newWorkerQueue(),
 		freshWorkers:  make(chan []byte, 1),
 		logger:        namedLogger,
 	}, nil
@@ -106,11 +108,10 @@ func NewBroker(name, frontendAddr, backendAddr string) (*Broker, error) {
 // Run kicks start the Broker and listens for requests. It blocks function
 // execution.
 func (lb *Broker) Run() {
-	workers := newWorkerQueue()
 	heartbeatAt := time.Now().Add(HeartbeatInterval)
 	for {
 		var sock *goczmq.Sock
-		if workers.Len() == 0 {
+		if lb.workers.Len() == 0 {
 			sock = lb.backendPoller.Wait(heartbeatIntervalMS)
 		} else {
 			sock = lb.bothPoller.Wait(heartbeatIntervalMS)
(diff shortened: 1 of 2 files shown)


pkg/server/plugin/zmq/broker.go
```diff
@@ -217,10 +225,11 @@ func (q *workerQueue) Ready(worker pworker) {
 	for i, w = range workers {
 		if bytes.Equal(w.address, worker.address) {
 			*q = append(append(workers[:i], workers[i+1:]...), worker)
```
cheungpat: an array that doesn’t work like any other arrays in golang
since we are changing the variable address here... are we sure all references to the array are updated?

pkg/server/plugin/zmq/broker.go
```diff
@@ -138,37 +165,11 @@ func (lb *Broker) Run() {
...
-
-	return backendPoller, bothPoller
-}
-
-func handleWorkerStatus(workers *workerQueue, address []byte, status string) {
+func (lb *Broker) handleWorkerStatus(workers *workerQueue, address []byte, status string) {
 	switch status {
 	case Ready:
+		workers.Ready(newWorker(address))
+		lb.freshWorkers <- address
```
cheungpat: looks like a good move to here

pkg/server/plugin/zmq/broker.go
```diff
@@ -168,11 +171,11 @@ func (lb *Broker) Run() {
...
 	switch status {
 	case Ready:
-		workers.Ready(newWorker(address))
-		lb.freshWorkers <- address
 		log.Infof("zmq/broker: ready worker = %s", address)
+		workers.Add(newWorker(address))
+		lb.freshWorkers <- address
 	case Heartbeat:
-		// do nothing
+		// no-ops
```
cheungpat: s/no-ops/no-op/

pkg/server/plugin/zmq/broker.go
cheungpat: oh you changed it in a later commit! good job!

cheungpat:
> zmq broker don't accept non register plugin heartbeat. It will force the plugin
> resend a proper Ready and go through the init process again.

I am not sure if this is the old/undesired or new/desired behaviour

rickmak:
> I am not sure if this is the old/undesired or new/desired behaviour

New, desired behaviour.

pkg/server/plugin/zmq/broker.go
```diff
@@ -194,54 +197,80 @@ func newWorker(address []byte) pworker {
...
+func (q *workerQueue) Add(worker pworker) {
+	q.addresses[string(worker.address)] = true
+	err := q.Tick(worker)
+	if err == nil {
+		return
+	}
+}
+
+func (q *workerQueue) Tick(worker pworker) error {
+	if _, ok := q.addresses[string(worker.address)]; !ok {
```
rickmak: `workerQueue.pworkers` Only holy the idle worker.  That is the queue for consuming order. It son't store all previously registered worker. So the state is not duplicated.

Hence, if we make `Add` only register worker, but not push to the queue. It will take next round heartbeat to push itself to worker.

limouren: Okay I forgot that a worker is removed from the queue before it is handed over a job. It LGTM then.

No I meant `Add` adds a worker to the queue and `Tick` will never add a new worker. Anyway it's not relevant now.

cheungpat: Although the pull request is merged (good job rickmak and limouren!), I’d just like to say that some comments to describe `Tick` and `Add` would be a great idea. Maybe next time.

Issue #160, opened 2016-10-11
Reuse the zmq socket or use go chan


rickmak:
Current zmqTransport is very generous on opening new sock https://github.com/SkygearIO/skygear-server/blob/master/pkg/server/plugin/zmq/rpc.go#L260

Since the zmq socket deallocation is asynchronous. The default 1024 limit will be easily reach.

- Skygear Server Date/Version: v0.17.0
- Application Platform:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Actual Results
https://github.com/SkygearIO/skygear-server/issues/150

PR #163, opened 2016-10-14
WIP: Use channel in messging between goroutine in stead of zmq.REQ

rickmak:
This will resolve socket exhaust problem on high concurrency.

connects #160


rickmak:
Close since Travis Keep building on the wrong commit.

PR #164, opened 2016-10-17
Add go1.7 to test matrix

rickmak:


cheungpat:
why?

cheungpat:
specifically why removing `-cover`?

rickmak:
Remove covert is my wrong.

My idea is to keep track the current code base work with go1.7 and make it easy for us to dev on 1.7 if we wanted to on future.

cheungpat:
superseded by #203

PR #165, opened 2016-10-17
Properly stop pubsub hub at test case

rickmak:


pkg/server/pubsub/hub.go
```diff
@@ -61,7 +62,7 @@ func (h *Hub) run() {
 			log.Warnf("Broadcast %v:%s", p.Channel, p.Data)
 			h.publish(p.Channel, p.Data)
 		case <-h.stop:
-			break
```
cheungpat: I think it is better to return here. The log function near the ends of the function should be wrapped inside a `defer` instead.

rickmak:
Updated.

PR #181, opened 2016-10-25
Use channel in messaging between goroutine instead of zmq.REQ

cheungpat:
This will resolve socket exhaust problem on high concurrency.

connects #160


oursky-travis:

PR #203, opened 2016-10-27
Fix travis test

cheungpat:
The fix comes in multiple parts:

* Drop table when cleaning up test db connection, this will also drop
  the db table change trigger.
* Wrap createTable/dropTable operation in a transaction.
* The pq part of the test suite has problem running with other parts,
  so we run the pq part separately.

connects #182

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: go
 
-go: 1.6.3
+go:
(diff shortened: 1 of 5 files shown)


oursky-travis:

cheungpat:
@rickmak This is ready for review.

Issue #140, opened 2016-09-15
Multi-Save incompatiable schema error


akiroz:
Incorrect error object in SDK-JS multiple record save operation.
- Skygear Server Date/Version: Skygear Cloud (Sep 15, 2016)
- Application Platform: SDK-JS / Chromium
#### Expected Results

``` js
{
  status: 409
  error: {
    code: 114
    message: "failed to migrate record schema"
    name: "IncompatibleSchema"
  }
}
```
#### Actual Results

``` js
{
  _id: "person/d0bcbb3a-c379-460a-9dba-1f46eb980df1"
  _type: "error"
  code: 10000
  name: "UnexpectedError"
  message: "pq: invalid input syntax for type double precision: \"2016-09-15T07:21:15.851Z\""
}
```
#### Steps to reproduce
-  create schema by saving record:

``` js
{
_id: "605a5375-f617-4593-b418-1aa60ecaffe8"
_recordType: "person"
name: "frederica"
age: 19
height: 164
}
```
-  save record with incompatible type schema (changed `age` to `Date` object)

``` js
{
_id: "605a5375-f617-4593-b418-1aa60ecaffe8"
_recordType: "person"
ownerID: "7888d764-a20a-4c2c-89c2-8eb02de0a079"
createdAt: Thu Sep 15 2016 15:20:34 GMT+0800 (HKT)
createdBy: "7888d764-a20a-4c2c-89c2-8eb02de0a079"
updatedAt: Thu Sep 15 2016 15:20:34 GMT+0800 (HKT)
updatedBy: "7888d764-a20a-4c2c-89c2-8eb02de0a079"
name: "frederica"
age: Thu Sep 15 2016 15:21:15 GMT+0800 (HKT)
height: 164
}
```
-  see error object in the `errors` array

cheungpat:
@akiroz need your schema information and your request payload

akiroz:
@cheungpat updated.

cheungpat:
only happening if there are multiple records to save or it also happens when saving single record?

akiroz:
only multi-record save.

PR #208, opened 2016-10-27
Fix unable to detect schema conflict

cheungpat:
This only happens when saving multiple records with a
schema-incompatible record being latter than schema-compatible ones.

connects #140


rickmak:
Consider adding test case?

PR #212, opened 2016-10-28
Make the broker stoppable and use new worker in each test case

rickmak:
connect #206

### DIFF
--- a/pkg/server/plugin/zmq/broker.go
+++ b/pkg/server/plugin/zmq/broker.go
@@ -93,6 +93,9 @@ type Broker struct {
 	logger  *logrus.Entry
 	// for stoping the channeler
 	stop chan int
+	// internal state for stoping the zmq Run when true.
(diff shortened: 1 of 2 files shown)


pkg/server/plugin/zmq/broker.go
```diff
@@ -93,6 +93,9 @@ type Broker struct {
 	logger  *logrus.Entry
 	// for stoping the channeler
 	stop chan int
+	// internal state for stoping the zmq Run when true.
```
cheungpat: s/stoping/stopping/

pkg/server/plugin/zmq/broker.go
```diff
@@ -93,6 +93,9 @@ type Broker struct {
...
+	// Should use the stop chan to stop. The stop chan will set this varibale.
```
cheungpat: s/varibale/variable/

Issue #219, opened 2016-11-04
Lifecycle state in which skygear server only accept plugin request


cheungpat:
Following #199, we would like to allow plugin to make request to the server before server accepts client requests. For example, we want to allow plugins to create Records for plugin’s use.

When plugin calls `send_action` for the above case, it should include a special attribute (e.g. `_from_plugin`) that the server only honor when the request is also authenticated with master key.

When the attribute is set to `true`, the server should accept the request in the lifecycle state between all plugins are ready and before accepting client request.

This need support from py-skygear (separate issue not created).

PR #221, opened 2016-11-08
Support plugin only request during plugin initialization

ben181231:
connect #219

### DIFF
--- a/pkg/server/plugin/handler.go
+++ b/pkg/server/plugin/handler.go
@@ -56,12 +56,22 @@ func NewPluginHandler(info pluginHandlerInfo, ppreg router.PreprocessorRegistry,
 func (h *Handler) Setup() {
 	if h.UserRequired {
 		h.preprocessors = h.PreprocessorList.GetByNames(
-			"plugin", "authenticator", "dbconn", "inject_user", "require_user")
+			"authenticator",
+			"dbconn",
+			"inject_user",
+			"require_user",
+			"plugin_ready",
+		)
 	} else if h.AccessKeyRequired {
 		h.preprocessors = h.PreprocessorList.GetByNames(
-			"plugin", "authenticator")
+			"authenticator",
+			"plugin_ready",
+		)
 	} else {
-		h.preprocessors = h.PreprocessorList.GetByNames("plugin")
+		h.preprocessors = h.PreprocessorList.GetByNames(
+			"accesskey",
+			"plugin_ready",
+		)
 	}
 }
 
--- a/pkg/server/preprocessor/plugin.go
+++ b/pkg/server/preprocessor/plugin.go
@@ -26,11 +26,34 @@ type EnsurePluginReadyPreprocessor struct {
 	PluginContext *plugin.Context
 }
 
-func (p *EnsurePluginReadyPreprocessor) Preprocess(payload *router.Payload, response *router.Response) int {
-	if !p.PluginContext.IsReady() {
-		log.Errorf("Request cannot be handled because plugins are unavailable at the moment.")
-		response.Err = skyerr.NewError(skyerr.PluginUnavailable, "plugins are unavailable at the moment")
+func (p *EnsurePluginReadyPreprocessor) Preprocess(
+	payload *router.Payload,
+	response *router.Response,
+) int {
+
+	// allow any request when plugins are ready
+	if p.PluginContext.IsReady() {
+		return http.StatusOK
+	}
+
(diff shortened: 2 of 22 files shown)


oursky-travis:
@ben181231, thanks for your PR! By analyzing the annotation information on this pull request, we identified @rickmak to be a potential reviewer

pkg/server/preprocessor/plugin.go
```diff
@@ -26,11 +26,29 @@ type EnsurePluginReadyPreprocessor struct {
...
+		}
+
+		return http.StatusOK
+	}
+
+	response.Err = skyerr.NewError(
+		skyerr.PluginUnavailable,
+		"plugins are unavailable at the moment",
+	)
+	return http.StatusServiceUnavailable
```
rickmak: I think a test case for this preprocessor is needed as it before more complicated.

pkg/server/preprocessor/plugin.go
```diff
@@ -26,11 +26,29 @@ type EnsurePluginReadyPreprocessor struct {
...
+	payload *router.Payload,
+	response *router.Response,
+) int {
+	if p.PluginContext.IsReady() {
+		return http.StatusOK
 	}
-	return http.StatusOK
+
+	if p.PluginContext.IsInitialized() {
+		if !payload.HasMasterKey() || !payload.IsFromPlugin() {
```
rickmak: The isFormPlugin() should only have effect if there is present of `master_key`.

ben181231: Okay. Maybe I should rename it to `IsPluginRequest()`

pkg/server/preprocessor/plugin.go
ben181231: Agree.

cheungpat:
I will leave more comments

pkg/server/skyerr/skyerr.go
```diff
@@ -119,6 +119,9 @@ const (
 	// PluginTimeout occurs when an operation carried by a plugin is timed out
 	PluginTimeout
 
+	// PluginInitializing occurs when any of the plugins are initializing
```
cheungpat: “Error codes for expected error condition should be placed above this line.”

i.e. do not insert errors in the middle of a list

pkg/server/router/model.go
```diff
@@ -150,6 +150,13 @@ func (p *Payload) HasMasterKey() bool {
 	return p.AccessKey == MasterAccessKey
 }
 
+// IsValidPluginRequest returns true when the payload indicate that the
+// request is from plugin AND the master key is used
+func (p Payload) IsValidPluginRequest() bool {
+	fromPlugin, _ := p.Data["_from_plugin"].(bool)
```
cheungpat: It is the job of a preprocessor to decide whether the request is a valid plugin request, not the `Request` struct. Move this function out of here.

pkg/server/plugin/handler.go
```diff
@@ -56,12 +56,22 @@ func NewPluginHandler(info pluginHandlerInfo, ppreg router.PreprocessorRegistry,
...
+			"authenticator",
+			"dbconn",
+			"inject_user",
+			"require_user",
+			"plugin_ready",
+		)
 	} else if h.AccessKeyRequired {
 		h.preprocessors = h.PreprocessorList.GetByNames(
-			"plugin", "authenticator")
+			"authenticator",
```
cheungpat: do you need `access_key` here? (and above)

pkg/server/router/model.go
cheungpat: Especially that the “HasMasterKey” function only returns valid value after a auth preprocessor update it.

pkg/server/skyerr/skyerr.go
ben181231: Oh. I see.

pkg/server/router/model.go
ben181231: Does it conflict with [this comment](https://github.com/SkygearIO/skygear-server/pull/221#discussion_r87138247) ?

pkg/server/plugin/handler.go
ben181231: The `access_key` is needed since the plugin handler is setting `AccessKeyRequired` to be true.

cheungpat: Ah I see now, either `accesskey` or `authenticator` is required. So `authenticator` fits the bill here.

Issue #213, opened 2016-10-28
Use ST_DWithin in where clause


louischan-oursky:
Use `ST_DWithin(p1, p2, d)` (http://postgis.net/docs/ST_DWithin.html) which automatically makes use of index available on the column to replace the expression `ST_Distance_Sphere(p1, p2) < d`

PR #222, opened 2016-11-08
Make use of `ST_DWithin` to compare distance

cheungpat:
Previously, predicate comparing distance were always compared using
`ST_Distance_Sphere`. Using `ST_DWithin` is better for comparing
where a location is within a certain distance because the latter
make use of available indexes.

While skygear-server does not create indexes on location field
automatically. Developer can add index to columns on their own.

connects #213


oursky-travis:

rickmak:
Merge via c06cd21

Issue #218, opened 2016-11-03
"unexpected Parse response " when signup a new user


bensonby:
**Problem**

Signing up a new user with give 500 error with message "Driver: bad connection"

<img width="491" alt="screen shot 2016-11-03 at 16 48 51" src="https://cloud.githubusercontent.com/assets/246195/19960056/75ccf4c2-a1e5-11e6-9e25-c5fda3b899d9.png">

**Steps to reproduce**

1. go to http://sdkjsexample.skygeario.com/static/example/
2. change the skygear endPoint and apiKey to do a signup for a portal staging app

  ```javascript
  apiKey = '<api key>';
  endPoint = 'https://bensonby3.staging.skygeario.com/';
  skygear.config({'apiKey': apiKey, 'endPoint': endPoint}).then(function() {
    skygear.signupWithEmail('bensonby+4@oursky.com', 'bensonby+4')
  });
  ```

**Note**

- It does not work for `bensonby3` (which has some cloud code), but it works for my other staging app `bensonby4`.

**Log**

```
time="2016-11-03T08:04:54Z" level=debug msg="------ Request: ------" logger=router 
time="2016-11-03T08:04:54Z" level=debug msg="{\"action\":\"auth:signup\",\"api_key\":\"someapikey\",\"access_token\":null,\"username\":\"bensonby+2@oursky.com\",\"email\":null,\"password\":\"********\"}" logger=router 
time="2016-11-03T08:04:54Z" level=debug msg="Opening DBConn: {pq bensonby3 postgresql://bensonby3-4041:dbpassword@skygear-cloud-staging.ctlziuhp6ccd.ap-southeast-1.rds.amazonaws.com:5432/bensonby3-4041}" logger=preprocessor 
time="2016-11-03T08:04:54Z" level=debug msg="Get DB OK" logger=preprocessor 
time="2016-11-03T08:04:55Z" level=debug msg="Executed SQL successfully with sql.Queryx" args=[] error=<nil> executionCount=1 logger=skydb sql="SELECT id FROM \"app_bensonby3\".\"_role\" WHERE by_default = true" 
time="2016-11-03T08:04:55Z" level=debug msg="0xc8204af6d0: Beginning transaction" logger=skydb 
time="2016-11-03T08:04:55Z" level=debug msg="0xc8204af6d0: Done beginning transaction 0xc820429e00" logger=skydb 
[...]

PR #224, opened 2016-11-09
Fix various issues with creating user on sign up

cheungpat:
* Add support for bigint db type
* Fix not enumerating all rows from database during record schema query
* Fix record schema cached even when there was an error querying record schema
* Update sqlx/libpq to latest version

connects #218


oursky-travis:

PR #229, opened 2016-11-11
Fix plugin lambda / handler always require API key

ben181231:


oursky-travis:
@ben181231, thanks for your PR! By analyzing the annotation information on this pull request, we identified @rickmak to be a potential reviewer

cheungpat:
I will merge this because this fix is needed ASAP but the true fix should be #230

Issue #231, opened 2016-11-14
Check if an unsupported data type is queried, make sure the error message is understandable


royuen:
- JS SDK
- iOS SDK
- Android SDK

cheungpat:
The following error is returned from skygear when the database contains a column with the type `money`:

```json
{
  "error": {
    "name": "UnexpectedError",
    "code": 10000,
    "message": "received unknown data type = money for column = money"
  }
}
```

This is what the user will see:

![Uploading Screen Shot 2016-11-17 at 2.23.17 PM.png…]()

royuen:
- [ ] If the column is not supported, the record should still be returned, and that column will have a specific json notation to specify this is unsupported and thus there is no value.

PR #236, opened 2016-11-22
Better handling for unrecognized data type

cheungpat:
This is achieved by introducing a new skydb type called Unknown, which
will be passed to client when the server sees a column with an
unrecognized data type.

Previous behavior will result in skygear server throwing an error when
it sees a column with unrecognized data type.

connects skygeario/skygear-server#231


oursky-travis:

rickmak:
Merge via a6ddba0

PR #248, opened 2016-12-07
Fix wrongly set S3 Asset ACL

ben181231:


cheungpat:
According to doc site:

> If your S3 is readable by the public, you can set `ASSET_STORE_PUBLIC` to `YES`
so that the access signature can be omitted.

Rejecting as the config only applies for object signing.

PR #255, opened 2016-12-15
Limit the number of bytes in request log

cheungpat:
The existing implementation prints the response body for certain type
this pose a problem to log collection when the response body is too big.


oursky-travis:

rickmak:
Merge via f1370ea

Issue #249, opened 2016-12-09
Add API to unset device token


royuen:
### Tasks

- [x] add `device:unregister` endpoint
- [ ] Update iOS SDK
- [x] Update Android SDK
- [x] Update JS SDK

PR #256, opened 2016-12-16
Support unregister device

ben181231:
connect #249

### DIFF
--- a/pkg/server/handler/device.go
+++ b/pkg/server/handler/device.go
@@ -49,6 +49,25 @@ func (payload *deviceRegisterPayload) Validate() skyerr.Error {
 	return nil
 }
 
+type deviceUnregisterPayload struct {
+	ID string
+}
+
+func (payload *deviceUnregisterPayload) Decode(data map[string]interface{}) skyerr.Error {
+	if err := mapstructure.Decode(data, payload); err != nil {
+		return skyerr.NewError(skyerr.BadRequest, "fails to decode the request payload")
+	}
+	return payload.Validate()
+}
+
+func (payload *deviceUnregisterPayload) Validate() skyerr.Error {
+	if payload.ID == "" {
+		return skyerr.NewInvalidArgument("Missing device id", []string{"id"})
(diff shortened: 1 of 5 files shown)


oursky-travis:
@ben181231, thanks for your PR! By analyzing the history of the files in this pull request, we identified @rickmak and @cheungpat to be potential reviewers.

pkg/server/handler/device.go
```diff
@@ -167,3 +182,90 @@ func (h *DeviceRegisterHandler) Handle(rpayload *router.Payload, response *route
...
+			"deviceID": payload.ID,
+			"err":      err,
+		}).Errorln("Fail to get device")
+
+		response.Err = skyerr.NewResourceFetchFailureErr("device", payload.ID)
+		return
+	}
+
+	// delete all all devices with the same token
+	if err := conn.DeleteDevicesByToken(device.Token, skydb.ZeroTime); err != nil {
```
cheungpat: This piece of code seems doesn’t belongs here. Is this to make sure there are no devices with the same token? But doing this will also delete the device that we will be saving.

Wondering why and wondering if there is a better place where this piece of code belongs.

ben181231: 😅

pkg/server/handler/device.go
```diff
@@ -167,3 +182,90 @@ func (h *DeviceRegisterHandler) Handle(rpayload *router.Payload, response *route
...
+		log.WithFields(logrus.Fields{
+			"deviceID": payload.ID,
+			"err":      err,
+		}).Errorln("Fail to get device")
+
+		response.Err = skyerr.NewResourceFetchFailureErr("device", payload.ID)
+		return
+	}
+
+	// delete all all devices with the same token
```
ben181231: 😅

pkg/server/handler/device.go
ben181231: > Is this to make sure there are no devices with the same token?

Yes.

Issue #239, opened 2016-11-28
Support token-based APNS


ben181231:
### Tasks
- [x] Support APNS token
- [x] Support Accepting Bundle Info. / Topic when Device Registration
- [x] Support Send Push Notification to Specific Bundle / Topic
- [x] Collect Bundle Info. on iOS SDK
- [x] Collect Bundle Info. on Android SDK
- [x] Collect Bundle Info. on Javascript SDK
- [x] Update `push_devices` and `push_users` on py-skygear

royuen:
1. Use APNS token
1. Use APNS topic to decide which token to use.

royuen:
- [ ]  Verify if APNs auth key works like this:
  -  Each team can have multiple auth key
  - Different bundle IDs could share the same auth key, also both sandbox and production
- [ ] Support 1 APNs auth key / 1 APNs cert / 1 FCM
- [ ] SDKs - when register device, supply bundle ID (auto for iOS & Android, manual for JS on React Native case)
- [ ] SDKs - send push notification (param: user) => will send to all devices of the user with bundle ID stored in _device table
- [ ] SDKs - send push notification (param; user, bundle_id) => will send to all devices of the user with the bundle ID specified in param
- [ ] User need to do migration if they upgrade to us the new APNs auth key, to manually enter existing bundle id to _device table

PR #259, opened 2017-01-03
Support token based APNS

ben181231:
connect #239

### DIFF
--- a/main.go
+++ b/main.go
@@ -415,9 +415,9 @@ func initAPNSPusher(config skyconfig.Configuration, connOpener func() (skydb.Con
 
 	switch config.APNS.Type {
 	case "cert":
-		pushSender = initCertBaseAPNSPusher(config, connOpener)
+		pushSender = initCertBasedAPNSPusher(config, connOpener)
 	case "token":
-		pushSender = initTokenBaseAPNSPusher(config, connOpener)
+		pushSender = initTokenBasedAPNSPusher(config, connOpener)
 	default:
 		log.Fatalf("Unknown APNS Type: %s", config.APNS.Type)
 	}
@@ -426,7 +426,7 @@ func initAPNSPusher(config skyconfig.Configuration, connOpener func() (skydb.Con
 	return pushSender
 }
 
-func initCertBaseAPNSPusher(
+func initCertBasedAPNSPusher(
 	config skyconfig.Configuration,
 	connOpener func() (skydb.Conn, error),
 ) push.APNSPusher {
@@ -448,7 +448,7 @@ func initCertBaseAPNSPusher(
 		key = string(keyPEMBlock)
 	}
 
-	pushSender, err := push.NewCertBaseAPNSPusher(
+	pushSender, err := push.NewCertBasedAPNSPusher(
 		connOpener,
 		push.GatewayType(config.APNS.Env),
 		cert,
@@ -461,7 +461,7 @@ func initCertBaseAPNSPusher(
 	return pushSender
 }
 
-func initTokenBaseAPNSPusher(
+func initTokenBasedAPNSPusher(
 	config skyconfig.Configuration,
 	connOpener func() (skydb.Conn, error),
 ) push.APNSPusher {
@@ -476,7 +476,7 @@ func initTokenBaseAPNSPusher(
 		key = string(keyBytes)
 	}
 
-	pushSender, err := push.NewTokenBaseAPNSPusher(
+	pushSender, err := push.NewTokenBasedAPNSPusher(
 		connOpener,
 		push.GatewayType(config.APNS.Env),
 		config.APNS.TokenConfig.TeamID,
--- a/main.go
+++ b/main.go
@@ -410,32 +410,84 @@ func initPushSender(config skyconfig.Configuration, connOpener func() (skydb.Con
 	return routeSender
 }
 
-func initAPNSPusher(config skyconfig.Configuration, connOpener func() (skydb.Conn, error)) *push.APNSPusher {
-	cert := config.APNS.Cert
-	key := config.APNS.Key
-	if config.APNS.Cert == "" && config.APNS.CertPath != "" {
-		certPEMBlock, err := ioutil.ReadFile(config.APNS.CertPath)
+func initAPNSPusher(config skyconfig.Configuration, connOpener func() (skydb.Conn, error)) push.APNSPusher {
+	var pushSender push.APNSPusher
+
+	switch config.APNS.Type {
+	case "cert":
+		pushSender = initCertBaseAPNSPusher(config, connOpener)
+	case "token":
+		pushSender = initTokenBaseAPNSPusher(config, connOpener)
+	default:
+		log.Fatalf("Unknown APNS Type: %s", config.APNS.Type)
+	}
+
+	go pushSender.Start()
+	return pushSender
+}
+
+func initCertBaseAPNSPusher(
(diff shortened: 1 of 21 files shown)


pkg/server/skydb/pq/device.go
```diff
@@ -24,16 +24,46 @@ import (
 )
 
 func (c *conn) GetDevice(id string, device *skydb.Device) error {
-	builder := psql.Select("type", "token", "user_id", "last_registered_at").
+	builder := psql.Select("type", "token", "user_id", "topic", "last_registered_at").
 		From(c.tableName("_device")).
-		Where("id = ?", id)
+		Where("id = ? AND topic IS NOT NULL", id)
```
cheungpat: in the following cases:
* pusher that does not require a topic to send a notification (do they exists?)
* users who migrated from older version of skygear-server to new version but they keep using cert-based APNS pusher

`topic IS NOT NULL` will never return a device because there is no topic associated with the device.

pkg/server/skydb/pq/device.go
```diff
@@ -24,16 +24,46 @@ import (
...
+
+	device.Token = nullableToken.String
+	device.UserInfoID = nullableUserID.String
+	device.LastRegisteredAt = device.LastRegisteredAt.In(time.UTC)
+	device.ID = id
+
+	return nil
+}
+
+func (c *conn) GetDeviceByIDAndTopic(id, topic string, device *skydb.Device) error {
```
cheungpat: There is zero or one device that will match an ID, therefore the topic is redundant here.

This also pose problems with the caller having to decide whether to call `GetDevice` or `GetDeviceByIDAndTopic`. In golang, you used function pointer, which seems to much work.

I am more inclined if you merge these two functions and adopt one of the following approach:

* if `topic` is empty string, ignore the `topic` filter
* remove the `topic` argument, and the caller is supposed to check whether the return devices has the correct topic (which is the job of the pusher IMO)

pkg/server/skydb/pq/device.go
ben181231: After changing the interface of `APNSPusher`, it requires a topic to send push notification and the topic is referring to the device's one for keeping consistency of cert-based APNS and token-based APNS.

As the previous discussion, the old users need a migration to insert the device topic even though they keep using cert-based APNS.

That is why I set `topic IS NOT NULL` here.

cheungpat: > As the previous discussion, the old users need a migration to insert the device topic even though they keep using cert-based APNS.

I am not aware we have a conclusion like that.

pkg/server/skydb/pq/device.go
ben181231: Okay. It seems that the `GetDeviceByIDAndTopic` does not make sense. Let me remove it. And I will make the checking on handler instead of pusher, since I think it is handler's responsibility to query the devices and ask pusher to send push notification.

pkg/server/skydb/pq/device.go
cheungpat: The conclusion was something like: “need a migration to insert the device topic *if* they use token-based APNS”

cheungpat: btw if the user is expected to do data migration when they upgrade, the device topic is always non-empty, why make the column nullable?

ben181231: The reason to make it nullable is that skygear server would not know the topic when migrating from existing users (i.e. what is done in `pkg/server/skydb/pq/migration/1981535c8aeb_add_device_topic.go`).

main.go
```diff
@@ -410,32 +410,84 @@ func initPushSender(config skyconfig.Configuration, connOpener func() (skydb.Con
...
+		pushSender = initTokenBaseAPNSPusher(config, connOpener)
+	default:
+		log.Fatalf("Unknown APNS Type: %s", config.APNS.Type)
+	}
+
+	go pushSender.Start()
+	return pushSender
+}
+
+func initCertBaseAPNSPusher(
```
cheungpat: I think either `CertBased` or simply `Cert`

cheungpat:
In my opinion, I think pusher do not need a reference of `conn`.

You end up with this dance of function calling to remove a device:

1. handleFailedNotification
2. unregisterDevice
3. pusher.deleteDeviceToken (this is probably one step more than necessary)
4. pusher.conn.DeleteDevicesByToken

pkg/server/skydb/pq/device.go
ben181231: > The conclusion was something like: “need a migration to insert the device topic if they use token-based APNS”

If that is the case, I would make the following changes:

- the push handler will query all devices to be sent (by device ID or by user ID), no matter whether the `topic` is `NULL` or not.
- the push handler will filter the queried devices depending on whether the topic is specified on the request.
- the pusher will send push notification to each of the queried devices and the pusher will decide whether the notification can sent to that device.
    - the cert-based APNS pusher will respect the topic specified in the cert, instead of the one on device, so that it can send to the devices with `NULL` topic.
    - the token-based APNS pusher will use the topic on device and ignore those devices with `NULL` topic.

do you think it is okay ?

cheungpat: > The reason to make it nullable is that skygear server would not know the topic when migrating from existing users

You may need [this](http://stackoverflow.com/questions/512451/how-can-i-add-a-column-to-a-postgresql-database-that-doesnt-allow-nulls).

cheungpat: > do you think it is okay ?

yes, thanks

ben181231: > > The reason to make it nullable is that skygear server would not know the topic when migrating from existing users
> 
> You may need [this](http://stackoverflow.com/questions/512451/how-can-i-add-a-column-to-a-postgresql-database-that-doesnt-allow-nulls).

I cannot come up with a reasonable default value. 😅

ben181231:
> In my opinion, I think pusher do not need a reference of `conn`.

Should I make something like `APNSFeedbackHandler` to handle the feedback from APNS and remove the bad APNS tokens on that handler? Then the pusher will not have the reference to `conn`.

pkg/server/handler/device_test.go
```diff
@@ -39,7 +40,21 @@ func (conn *naiveConn) GetDevice(id string, device *skydb.Device) error {
...
-	if !ok {
+	if !ok || targetDevice.Topic == "" {
+		return skydb.ErrDeviceNotFound
+	}
+
+	*device = targetDevice
+	return nil
+}
+
+func (conn *naiveConn) GetDeviceByIDAndTopic(id, topic string, device *skydb.Device) error {
```
cheungpat: this is not needed

pkg/server/push/cert_apns.go
```diff
@@ -0,0 +1,213 @@
...
+		"deviceID":    device.ID,
+		"deviceTopic": device.Topic,
+	})
+
+	if m == nil {
+		logger.Warn("Cannot send push notification with nil data.")
+		return errors.New("push/apns: push notification has no data")
+	}
+
+	if device.Topic == "" {
```
cheungpat: Since cert-based pusher does not require topic (the topic is in the certificate), what will happen if the device has null topic here?

ben181231:
Updated.

cheungpat:
I am going to merge this.

cheungpat:
thanks ben

PR #280, opened 2017-02-02
Add Nix derivation for building

b123400:
Now you can build with a single command `nix-build`~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,17 @@ $ # export WITH_ZMQ=1 # If you need ZeroMQ support
 $ make build
 ```
 
+#### Building with Nix
+
+Assuming you have [Nix](https://nixos.org/nix/) installed,
+Skygear can be built with the following command:
+
+```shell
+nix-build default.nix
+```
+
+You will have a symbolic link `result-bin` linking to the binary.
+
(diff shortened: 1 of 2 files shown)


cheungpat:
Question (I have no background in nix):

Do we need to keep deps.nix and glide.lock in sync? if yes, perhaps we need to add a build target in `Makefile` so that we will update both files.

README.md
```diff
@@ -69,6 +69,8 @@ If using Mac OS X, you can get the ZeroMQ dependencies using Homebrew:
 $ brew install libsodium zeromq czmq
 ```
 
+Or you can use Nix mentioned below, which also manages system dependencies.
```
cheungpat: I’d prefer to keep this section simple and don’t burden contributors with choices here.

Like I said, people who use nix should know what they are doing and they will just run the command below.

b123400:
Updated, now it downloads deps using `glide` before build.

cheungpat:
I tried building skygear-server with nix and it works flawlessly. That’s really simple and great.

I also added `result-bin` to `.gitignore`.

b123400:
Thx! Forgot about gitignore...

Issue #284, opened 2017-02-09
Signup fails with


rickmak:
If the plugin have any user before_save hook in plugin. The skygear-server will fails user signup with error.

This is the example of plugin code that will trigger the bugs.
```python
@skygear.before_save("user", async=False)
def handle_user_before_save(record, original_record, db):
    return
```

- Skygear Server Date/Version: server 0.21.1, python 0.21.0
- Application Platform: Python
- [x] Is this a regression?
- [ ] Attached logs, screenshots

#### Expected Results
The user sign up successfully.

#### Actual Results

User will get follow error

```json
{
    "error": {
        "code": 10000, 
        "message": "panic occurred while handling request: runtime error: invalid memory address or nil pointer dereference", 
        "name": "UnexpectedError"
    }
}
```

Stacktrace
```golang
ERRO[0009] goroutine 70 [running]:
runtime/debug.Stack(0x45724e0, 0x456bbc0, 0xc42000c0e0)
	/usr/local/Cellar/go/1.7.1/libexec/src/runtime/debug/stack.go:24 +0x79
github.com/skygeario/skygear-server/pkg/server/router.errorFromRecoveringPanic(0x456bbc0, 0xc42000c0e0, 0xc42000c0e0, 0x40148e6)
	/Users/rickmak/go/src/github.com/skygeario/skygear-server/pkg/server/router/errors.go:61 +0x1c5
github.com/skygeario/skygear-server/pkg/server/router.(*Router).ServeHTTP.func1(0xc4202c5730, 0x46606c1, 0x6, 0xc4202f3940)
	/Users/rickmak/go/src/github.com/skygeario/skygear-server/pkg/server/router/router.go:107 +0x310
panic(0x456bbc0, 0xc42000c0e0)
	/usr/local/Cellar/go/1.7.1/libexec/src/runtime/panic.go:458 +0x243
encoding/json.(*encodeState).marshal.func1(0xc4202f29a0)
	/usr/local/Cellar/go/1.7.1/libexec/src/encoding/json/encode.go:272 +0x1b9
panic(0x456bbc0, 0xc42000c0e0)
	/usr/local/Cellar/go/1.7.1/libexec/src/runtime/panic.go:458 +0x243
context.(*valueCtx).Value(0xc4202a5470, 0x453a5e0, 0xc4202f78e0, 0x453a5e0, 0xc4202f78e0)
[...]

PR #286, opened 2017-02-09
Fix request context not initialized

cheungpat:
This fix is backported to apply on v0.21 branch.

refs #284


rickmak:
Merge via c33d896

PR #287, opened 2017-02-11
Actually fetch assets data in query

b123400:
connect https://github.com/SkygearIO/skygear-SDK-JS/issues/164

The original code does not query assets at all, feels like intentionally leaving some fields blank to save a query, so this PR is not really fixing bug, but adding functionality.

### DIFF
--- a/pkg/server/skydb/pq/asset.go
+++ b/pkg/server/skydb/pq/asset.go
@@ -15,7 +15,6 @@
 package pq
 
 import (
-	"database/sql"
 	"errors"
 
 	"github.com/skygeario/skygear-server/pkg/server/skydb"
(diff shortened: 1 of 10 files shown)


pkg/server/skydb/pq/asset.go
```diff
@@ -45,6 +46,42 @@ func (c *conn) GetAsset(name string, asset *skydb.Asset) error {
 	return err
 }
 
+func (c *conn) GetAssets(names []string) ([]skydb.Asset, error) {
```
ben181231: Better to change `GetAsset` to use `GetAssets` to reduce code duplication.

rickmak:
ping @b123400 ?

b123400:
Updated

PR #290, opened 2017-02-14
Remove CloseNotifier

cheungpat:
The CloseNotifier is in conflict with the Hijacker interface (they
cannot be used at the same time). The context will not be cancelled when
the close channel from CloseNotifier is called. This is because the behavior
is not well understood. We can introduce it later if it poses a real problem.

connects #291


rickmak:
Merge via b1a2c5e

PR #293, opened 2017-02-15
Fix schema:fetch when record type has no fields

cheungpat:
The old behavior is that the fields key contains
null value if the record type has no fields. This
is because the slice is not initialized to empty slice.


ben181231:
Picked as 4b81eab.

Issue #295, opened 2017-02-16
Bidirectional plugin transport and multiplexing


cheungpat:
- [x] [Golang] Try to send a dummy message after a new request, before receiving the response 
- [x] [Python] Try to catch that message
- [x] [Python] Look at the normal http transport
- [x] [Python] Send request via zmq
- [x] [Golang] Handle that request
  - [x] Create fake http request from zmq request
  - [x] Inject into http server
- [x] [Golang] Send back the worker id in response
- [x] [Python] Ask the worker to handle the response
- [x] [Golang] Handle worker-reuse when a worker is calling itself
- [x] [Golang] Support multiple plugins
- [x] [Golang] Add a stack count
  - [x] Set a maximum and throw error when exceeded 

---

The current plugin transport only support communication initiated from the server side of the transport. 

There is an existing mechanism for the plugin to make request to the server just like a normal client by making http request through request router. This mechanism is often referred to as `send_action` named after the python function with the same name.

It is proposed that we should implement bidirectional plugin transport so that plugin-initiated request are made through the plugin transport instead. Multiplexing refers to how transport route communication among workers.

## Benefits for implementing bidirectional plugin transport and multiplexing

* easier configuration - while transport may require extra config for bidirectional communication, no config is necessary for api key, master key and server endpoint.
* reduced overhead - imitating a client request requires the plugin to make extra http connection, which is extra overhead when using zmq transport.
* tracking request lifecycle - keep track of control flow across process boundary (to support better logging).
* reuse blocked workers - control flow can be handed off to the blocked worker if the blocked worker is handling the same request.

## Implementation approach

* (existing behavior) plugin runtime register a new worker for each thread
* transport keep track of context when sending message across process boundary
* transport to increment call stack when crossing process boundary
* when call-stack=0, select a free worker from pool, worker is no longer in the pool for other requests
* call-stack++ and send message to worker
* worker may call server code (new behavior), two options here (see below)
* call-stack++ when receiving message from worker that will call other server code
* when call-stack=2, re-use the same worker previously selected, use request unique ID to find the worker
* when plugin code returns data (i.e. not calling other server code) call-stack is decremented
* when call-stack=0, the worker is marked free and put back to pool

### ZMQ Transport protocol options

Option 1: Overload the current ZMQ transport protocol design in that plugin code calls server code by using the response message, the server code returns data by using the request message.

This option is less intrusive to the overall ZMQ transport protocol design. The semantic maybe convoluted.

[...]

PR #300, opened 2017-02-28
[WIP] Bidirectional plugin transport and multiplexing 

b123400:
connect https://github.com/SkygearIO/skygear-server/issues/295
depends on https://github.com/SkygearIO/py-skygear/pull/123

Work in progress

### DIFF
--- a/pkg/server/plugin/zmq/broker.go
+++ b/pkg/server/plugin/zmq/broker.go
@@ -103,6 +103,8 @@ type Broker struct {
 	addressChan map[string]chan []byte
 	// ReqChan is a channel for sending incoming request to handler
 	ReqChan chan *parcel
+	// respChan is a channel for sending response from handler to zmq
+	respChan chan [][]byte
 	// for RPC timeout, used by Channeler
 	timeout chan string
 	// determine how long will timeout happen, relative to the
@@ -131,6 +133,7 @@ func NewBroker(name, backendAddr string, timeoutInterval int) (*Broker, error) {
 		recvChan:        make(chan *parcel, 10),
 		addressChan:     map[string]chan []byte{},
 		ReqChan:         make(chan *parcel, 10),
+		respChan:        make(chan [][]byte, 10),
 		timeout:         make(chan string),
 		timeoutInterval: time.Duration(timeoutInterval),
 		workers:         newWorkerQueue(),
@@ -261,6 +264,7 @@ func (lb *Broker) Channeler() {
 						[]byte{},
 						response,
 					}
+					lb.respChan <- frames
 				}(parcel)
 			} else if messageType == Response {
 				respChan, ok := lb.addressChan[requestID]
@@ -308,6 +312,8 @@ func (lb *Broker) Channeler() {
 			go lb.setTimeout(address)
 		case address := <-lb.timeout:
 			respChan, ok := lb.addressChan[address]
+		case frames := <-lb.respChan:
+			push.SendMessage(frames)
 			if !ok {
 				break
 			}
--- a/pkg/server/plugin/zmq/broker.go
+++ b/pkg/server/plugin/zmq/broker.go
@@ -92,15 +92,88 @@ func requestToChannelKey(requestID string, bounceCount int) string {
 	return fmt.Sprintf("%s-%d", requestID, bounceCount)
 }
 
-// Broker implements the Paranoid Pirate queue described as follows:
+// Broker implements a protocol based on the Paranoid Pirate queue described as follows:
 // Related RFC: https://rfc.zeromq.org/spec:6/PPP
 // refs: http://zguide.zeromq.org/py:all#Robust-Reliable-Queuing-Paranoid-Pirate-Pattern
 // with the addition of:
 //
 // 1. Shutdown signal, which signifies a normal termination of worker to provide
 //    a fast path of worker removal
+// 2. Extra frames for bidirectional and multiplexing, see below
+//    Related issue: https://github.com/SkygearIO/skygear-server/issues/295
+//
 // TODO: channeler can be separated into a separate struct, communiate with
 // broker using frontend chan, workers and pull/push sock address.
+//
+// In PPP, the message has 3 frames, in the extended format, it has 7.
+// 1. Worker address
+// 2. Empty frame
+// 3. Message type
+// 4. Bounce Count
+// 5. Request ID
+// 6. Empty frame
+// 7. Message body
+//
+// Message type is either "REQ" or "RES"
+// Bounce count is an integer that increase when the request is nested, starts from 0
+// Request ID is a string that identify the request (including nested requests)
+//
+// The flow goes like this.
+// # Connecting
+// 1. A plugin connects to the server via ZMQ
+// 2. Plugin sends a Ready message, with a worker ID
+// 3. Server remembers that as a ready worker, and starts sending heartbeat.
+//
+// # Server sends requests to plugin
+// 1. Server finds a free worker
+// 2. Server sends the request message, e.g.
+//    [WORKER-ADDR|0|REQ|0|REQ-ID|0|body....]
+// 3. The plugin replies with the following message
+//    [WORKER-ADDR|0|RES|0|REQ-ID|0|reply body...]
+// Plugin can send requests to server in the exact same format
+//
+// # Nested requests (send a request when handling a request)
+// 1. Server finds a free worker
+// 2. Server sends the request message, e.g.
+//    [WORKER-ADDR|0|REQ|0|REQ-ID|0|Do you have beer?]
+// 3. The plugin sends a NESTED request to the server, bounce_count increases by 1
+//    [WORKER-ADDR|0|REQ|1|REQ-ID|0|Do you have money?]
+// 4. Server responds:
+//    [WORKER-ADDR|0|RES|1|REQ-ID|0|Yes I have money]
+// 5. Plugin responds:
+//    [WORKER-ADDR|0|RES|0|REQ-ID|0|Yes I have beer]
+//
+// # Maximum bounce count
+// broker has a maximum bounce count (maxBounce),
+// when it is exceeded, the request fails immediately, this prevents infinity loop
+//
+// # About multiple plugins
+// When multiple plugins are involved, both request ID and bounce-count is accumulated
+// accross plugins. For example, this is what happens when Plugin A calls Plugin B
+//
+// PluginA code:
+// ```
+// @op('foo:hello')
+// def english():
+//     send_action('foo:ciao')
+//     return {'key':'thanks'}
+// ```
+
+// PluginB code:
+// ```
+// @op('foo:ciao')
+// def italian():
+//     return {'key':'grazie'}
+// ```
+// Would results in:
+// 1. Browser calls /foo/hello
+// 2. Server calls pluginA with `REQ|0|request-id`
+// 3. PluginA calls server with `REQ|1|request-id`
+// 4. Server calls pluginB with `REQ|2|request-id`
+// 5. PluginB responses server with `RES|2|request-id`
+// 6. Server responses pluginA with `RES|1|request-id`
+// 7. PluginA responses server with `RES|0|request-id`
+
 type Broker struct {
 	// name is assume to be unique and used to construct the zmq address
 	name string
--- a/pkg/server/plugin/zmq/broker.go
+++ b/pkg/server/plugin/zmq/broker.go
@@ -130,10 +130,12 @@ type Broker struct {
 	// internal state for stopping the zmq Run when true.
 	// Should use the stop chan to stop. The stop chan will set this variable.
 	stopping bool
+	// The maximum bounce count, request larger than this number will be aborted
+	maxBounce int
 }
 
 // NewBroker returns a new *Broker.
-func NewBroker(name, backendAddr string, timeoutInterval int) (*Broker, error) {
+func NewBroker(name, backendAddr string, timeoutInterval int, maxBounce int) (*Broker, error) {
 	namedLogger := log.WithFields(logrus.Fields{
 		"plugin": name,
 		"eaddr":  backendAddr,
@@ -153,6 +155,7 @@ func NewBroker(name, backendAddr string, timeoutInterval int) (*Broker, error) {
 		logger:          namedLogger,
 		stop:            make(chan int),
 		stopping:        false,
+		maxBounce:       maxBounce,
 	}
 
 	go broker.Run()
@@ -307,6 +310,11 @@ func (lb *Broker) Channeler() {
 				requestID = generateRequestID()
 			}
 			bounceCount := p.bounceCount
+			if bounceCount > lb.maxBounce {
+				lb.logger.Infof("zmq/broker: bounce count of %d exceeded the maximum %d\n", bounceCount, lb.maxBounce)
+				p.respChan <- []byte{1}
+				break
+			}
 			if address == "" {
 				if p.retry < HeartbeatLiveness {
 					p.retry += 1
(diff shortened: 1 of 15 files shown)


rickmak:
Pick ccbaca1 and be85c21 to a separate PR?

pkg/server/router/common.go
```diff
@@ -31,25 +31,40 @@ import (
 // to Router and Gateway.
 type commonRouter struct {
 	payloadFunc      func(req *http.Request) (p *Payload, err error)
-	matchHandlerFunc func(req *http.Request, p *Payload) (h Handler, pp []Processor)
+	matchHandlerFunc func(path string, method string, p *Payload) (h Handler, pp []Processor)
```
rickmak: After reading it with the `router.Payload`. I think the matchHandlerFunc should only accept `Payload` alone. The action string should be a part of data. The HTTP method should be part of Meta?

I think `Meta` is designed for that purpose, but implementation just ignore it.

pkg/server/plugin/zmq/handler.go
```diff
@@ -0,0 +1,73 @@
...
+	w.response = append(w.response, body...)
+	return len(body), nil
+}
+func (w *zmqResponseWriter) WriteHeader(status int) {
+	w.header = status
+}
+
+func (p *zmqTransport) SetRouter(router *router.Router) {
+	p.router = router
+}
```
rickmak: Put all `zmqTransport` member into the same files?

pkg/server/plugin/zmq/handler.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	responseWriter := &zmqResponseWriter{}
+	resp := router.NewResponse(responseWriter)
+
+	newCtx := context.WithValue(payload.Context, ZMQWorkerIDContextKey, workerID)
+	payload.Context = newCtx
+
+	// TODO(b123400), put real path and method here
+	p.router.HandlePayload("schema/fetch", "POST", payload, resp)
```
rickmak: Should be `p.router.HandlePayload(payload, resp)` ideally?

pkg/server/plugin/zmq/broker.go
```diff
@@ -48,23 +50,37 @@ const (
...
+func newParcel(frame []byte) *Parcel {
+	return &Parcel{
+		Worker:   "",
+		RespChan: make(chan []byte),
+		Frame:    frame,
 		retry:    0,
 	}
 }
 
+var requestIDPrefix = ""
```
rickmak: I think this `requestIDPrefix` should be at `router.Payload` Meta?

pkg/server/plugin/zmq/broker.go
```diff
@@ -221,50 +243,93 @@ func (lb *Broker) Channeler() {
...
 				[]byte{},
-				p.frame,
+				[]byte(Request),
+				[]byte("0"),
+				[]byte(requestID),
+				[]byte{},
+				p.Frame,
 			}
-			lb.addressChan[address] = p.respChan
+			lb.addressChan[requestID] = p.RespChan
```
rickmak: The key should be `prefix` only?

pkg/server/plugin/zmq/broker.go
```diff
@@ -221,50 +243,93 @@ func (lb *Broker) Channeler() {
...
+				respChan <- message
+			}
 		case p := <-lb.recvChan:
 			// Save the chan and dispatch the message to zmq
 			// If current no worker ready, will retry after HeartbeatInterval.
 			// Retry for HeartbeatLiveness times
-			address := lb.workers.Next()
+			// TODO(b123400): instead of asking for a new worker
+			// get the current worker (if possible, e.g. when nested, we already have the worker id)
+			// Or, we can put the worker back to pool when received request from plugin? not very good?
```
rickmak: I think we only put the worker back to pool once the bounce count is back to zero? For worker not in pool, we retrieve it by `requestPrefix` for request that bounce count is not zero?

pkg/server/router/common.go
b123400: I am not 100% sure, but I remember Boris asked why do we have both action name in url path and payload, and the answer was like because sometimes we don't have the payload (e.g. a GET request). Create a fake payload from url when payload is nil?

pkg/server/plugin/zmq/handler.go
b123400: ok~

pkg/server/plugin/zmq/handler.go
b123400: The current implementation depends on both path and method, so I have to pass it in.
We either:
- Make it path / method independent, a payload alone is enough for routing
- Hide the path / method somewhere in the payload, maybe `Meta` as you said

pkg/server/plugin/zmq/broker.go
b123400: > we only put the worker back to pool once the bounce count is back to zero
Agree

pkg/server/plugin/zmq/broker.go
b123400: I would like to save the response channels by something like `"${request_id}-${bounce_count}"` which is more clear and we don't have to split id into parts.
(Bounce count is not implemented yet so you don't see it in code)

Splitting id into prefix-suffix means we will introduce another schema in the protocol, which is not necessary. The current design allows any form of request-id and worker-id, it can be a uuidv4 and everything will be fine, prefix-suffix just makes things easier to debug, but we should not enforce it in protocol level.

pkg/server/plugin/zmq/broker.go
b123400: I seems to have misunderstanding on how should a request-id looks like. I followed [python's code](https://github.com/SkygearIO/py-skygear/blob/5e769490051038837c7050019fe7fabb49010c2e/skygear/transmitter/zmq.py#L49) where the first part is a global constant.

Wait... in python it is per worker constant... I will make it per worker in Golang too, so it is more consistent

b123400: I seems to have misunderstanding on how should a request-id looks like. I followed [python's code](https://github.com/SkygearIO/py-skygear/blob/5e769490051038837c7050019fe7fabb49010c2e/skygear/transmitter/zmq.py#L49) where the first part is a global constant.

Wait... in python it is per worker constant... I will make it per worker in Golang too, so it is more consistent

pkg/server/router/common.go
rickmak: The payload design rationale is: payload contain all information on the request. Transport may duplicate the information to transport specific place. For example, HTTP path in HTTP transport, access_token at HTTP header. But the payload will always contains all information skygear server need to process the request. The additional copy will assumed to be for fast routing.

In the case of ZMQ, we may copy the action into routing frame. It will enable the program to route the request without actually parsing the whole JSON.

pkg/server/plugin/zmq/handler.go
rickmak: Payload alone should be enough for routing.

pkg/server/plugin/zmq/broker.go
rickmak: The python worker ID are designed to identify the worker. Here the request ID are design to identify request.

In python, the prefix is global to the process. The suffix is global to the thread. So in skygear-server, we can just look at the worker related log and know which process x thread that log is belonging to.

So in here, the request ID should do the same. By looking at the log, we will know which nested request cycle a particular request is belong to. The prefix generate on skygear-server accepting request from SDK.

pkg/server/plugin/zmq/broker.go
rickmak: This is not on protocol design level? It just about how golang relate the address and chan. So using a more ease to debug format look the right choice to me.

pkg/server/plugin/zmq/broker.go
```diff
@@ -48,23 +50,45 @@ const (
...
+	Response = "RES"
 )
 
-// parcel is used to multiplex the chan with zmq worker
-type parcel struct {
-	respChan chan []byte
-	frame    []byte
-	retry    int
+// Parcel is used to multiplex the chan with zmq worker
+type Parcel struct {
```
limouren: Why make it public? I didn't see it used outside of the `zmq` package, and if there are such case, we probably have other problems.

pkg/server/plugin/zmq/handler.go
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package zmq
+
+import (
+	"net/http"
+)
+
+type zmqResponseWriter struct {
```
limouren: Extremely sad seeing zmq structs implementing http's interface. We should aim to remove this coupling.

pkg/server/plugin/zmq/rpc.go
```diff
@@ -176,6 +194,54 @@ func (p *zmqTransport) ipc(req *pluginrequest.Request) (out []byte, err error) {
...
+
+	// TODO(b123400), put real path and method here
+	p.router.HandlePayload(path, method, payload, resp)
+
+	responseChan <- responseWriter.response
+}
+
+func (p *zmqTransport) listenRequests() {
+	for {
+		select {
```
limouren: It's a bit odd to see a `select` with only a single `case`. Typically we will simply write `parcel := <- p.broker.ReqChan`.

pkg/server/plugin/zmq/rpc.go
```diff
@@ -176,6 +194,54 @@ func (p *zmqTransport) ipc(req *pluginrequest.Request) (out []byte, err error) {
...
+		case parcel := <- p.broker.ReqChan:
+			go p.handleRequest(parcel)
```
limouren: IMO, unless you have other plans at mind, it's a bit odd to spawn a goroutine just to take messages from a channel and then spawn another goroutine to handle that message. Shouldn't we just spawn that goroutine at the calling site instead?

b123400: What do you mean by calling site? I am doing this because I want everything to be async, one slow request should not block other requests

limouren: I meant instead of having people write things like this:

```go
parcel = makeParcel()
broker.ReqChan <- parcel
// do other things
```

Let them write this:

```go
go func() {
	parcel = makeParcel()
	broker.ReqChan <- parcel
}()

// do other things immediately
```

If you don't want to block, demand it explicitly. The problem of this `listenRequests` is that it does far too little things to justify itself a thread of execution. Consider it in Erlang or Elixir, would you spawn a process just to pass massage hand to hand without doing anything? I bet no.

In this request-response case, it might even be better to make a function to hide the complexity of asynchrony:

```go
// purely imaginary
func doRequest(parcel Parcel) ([]byte, error) {
	respChan = make(chan []byte)
	errChan = make(chan error)

	parcel.respChan = respChan
	parcel.errChan = errChan

	broker.ReqChan <- parcel
	select {
		resp := <- respChan:
			return resp, nil
		err := <- errChan:
			return nil, err
		// other things here, like timeout
	}
}

// some busy code here...
go func() {
	parcel = make_parcel()
	resp, err = doRequest(parcel)
	// now process
}()

// do other things immediately
```

b123400: Um... let me confirm if I understand your comment correctly. The current implementation is
1. Broker puts parcel to `ReqChan`
2. zmqTransport listens `ReqChan`, and spawn a thread (handleRequest) per request
3. `handleRequest` does things synchronously 

So it is `broker -> listenRequests -> handleRequest`, 3 levels of async.
Your suggestion is, `broker -> handleRequest`, 2 level of async.

When I write the code, I am not really concerning about how many levels, or is it too small or big, but their responsibility, their scope.
Broker has been focusing on 1 thing only, sending and receiving zmq messages. Both input and output of broker has always been just buffer, it receives buffers via `.RPC` and output via `reqChan`. It does not have the knowledge of transport (the Transport interface). It doesn't even care if the response is handled, if you don't care about the response, you can just throw away the channel and everything will be fine.
On the other hand, `zmqTransport` is more about application logic, it transform concept of Request and Response into buffer, bridging the abstract concept of request response to the underlying bytes.
`zmqTransport` has the knowledge of the broker, but not the other way round. That is what I am try to maintain while making the broker interface, broker should output a stream of requests, it will be callback if it is Javascript, or delegate if it is Objective-C. 

If there is a single piece of code, somewhere, doing both `make_parcel()` and `doRequest(parcel)`, we will be crossing the line, now broker will know about the transport, it will enforce every request has to be handled. While it is always the case that requests are handled, the 2 pieces of code will be tied together and we are not separating the responsibilities anymore.

Speaking of Erlang, I think it's kind of different because their threading is more lightweight, and they do encourage many small processes.

pkg/server/plugin/zmq/broker.go
b123400: OK, lets put the conversation about request-id in this thread.
Let me coin 2 terms first,

1. __The big request__ - Outer most request, requests that our current implementation can make
2. __The small request__ - Nested request, requests that are made when handling another request.

What you are saying is, request-id should have 2 parts. `PREFIX-SUFFIX`, the prefix refers to the __big__ request, and suffix refers to __small__ requests, right?

For example:

```
@op('foo:hello')
def english():
    send_action('foo:ciao')
    return {'key':'thanks'}

@op('foo:ciao')
def italian():
    return {'key':'grazie'}
```

Would results in:
1. Browser calls /foo/bar
2. Server calls plugin with request-id "ZZZZ-AAAA" (foo:hello)
3. Plugin calls server with request-id "ZZZZ-BBBB" (foo:ciao)
4. Server calls plugin with request-id "ZZZZ-CCCC" (foo:ciao)
5. Plugin responses server with request-id "ZZZZ-CCCC" (foo:ciao)
6. Server responses plugin with request-id "ZZZZ-BBBB" (foo:ciao)
7. Plugin responses server with request-id "ZZZZ-AAAA" (foo:hello)

If the this is wrong, can you provide the correct data for the above code?

Assuming I am correct, saving prefix is not good enough.
1. In step (2), server will save channel for "ZZZZ-AAAA" with the key "ZZZZ".
2. In step (4), server will save channel for "ZZZZ-CCCC" with the key "ZZZZ".
Saving only the suffix is acceptable, but I cannot find any reason not to save the entire request id, after all, `addressChan` is broker level, not request level.

pkg/server/plugin/zmq/handler.go
rickmak: Agree!

pkg/server/plugin/zmq/broker.go
b123400: OK~ just make it private again

pkg/server/plugin/zmq/handler.go
b123400: Agree, but I am afraid I won't refactor in this PR.

Although I am told the current routing / handler mechanism is "http independent", it does not seems to be true. For example `router.Response` depends on `http.ResponseWriter`, and `router.NewGateway` depends on the concept of "path", which zmq does not have.

Since the current implementation is http orientated, I am going to follow it.

rickmak:
@limouren let me invite you to do the peer review on channel with me 💯

pkg/server/plugin/zmq/rpc.go
limouren: Okay I probably have a different picture from yours. In my mind Broker is a mediator between skygear-server (i.e. zmq transport currently) and zmq workers. So the primary role of broker is to exchange messages between transport and workers. Previously only transport knows about broker because requests are only initiated from skygear-server side. Now we allow workers to initiate requests. And it appear naturally to me for broker to know about both ends and exchange messages between them. I cannot speak for whether others in the team have this same picture as mine though.

limouren: As a side track I think goroutine is at least as lightweight as Erlang's process - correct me if I am wrong.

rickmak:
I merged this PR into zmq-multiplex branch, and open following issue in order to make it good enough to merge to master.

Please refer to #303, #304, #305, #306, #307, #308

Issue #261, opened 2017-01-07
Simplify records needs a user to query/update via cloud functions


david90:

royuen:
- [ ] Improve generate access token (portal: can choose will or will not expire, JWT)
- [ ] Write guide for generate access token for cloud usage
- [ ] If login with master key, it will login as "master-key-user" user. All recording saving will be belonging to this "master-key-user" user.

chpapa:
This issue is tightly related with Skygear Portal / [skycli](https://github.com/skygeario/skycli for actual usage.

So when @b123400 work on this issue, what are the client side deliverables are we expecting? Or no deliverables on client side? (In that case, can we create those issues in skygear portal / skycli and ref here?)

If the deliverable include portal, we will need some design?

chpapa:
Discussed again with @b123400 @rickmak and @ben181231 

1. Why we need another access token? We already have asUserId in CloudContainer which allow Cloud Functions call with User ID to "switch" to another user immediately. Ref: https://docs.skygear.io/guides/cloud-function/calling-skygear-api/python/#using-the-master-key
2. We can still have master_key default user. @b123400 think it is too much constraint for developers to lose control to one username.

Agreement so far:

1. We will check with @david90 again to see if asUserID is good enough.
2. We will still implement the Cloud Functions call, when called with MasterKey, will access the Database as a Master Key God User.

Any comments? @david90 @cheungpat

chpapa:
@david90 responded said asUserID is actually good enough...

I'm actually thinking if we really need to implement the master key thing.....? Would that really help?

ben181231:
I think adding Master Key God User increases the difficulties for developers to understand the whole auth mechanism since it is adding another exceptional case for the auth and ACL mechanism.

b123400:
After discussing on Slack, the conclusion is creating a god user when the request is using the master key and no user auth is present

rickmak:
May @david90 provide a use-case?

david90:
For example in a cron job:
1) the job is to fetch the latest 50 notes and push to slack.
2) the job is to fetch the latest 50 notes of the user and push to slack.

chpapa:
Am I correct that in the use cases @david90 suggested, you just wish to save the notes as some users, but don't care whose?

In those cases, which approach is the best? (Master API = god user, asUserID is enough, or others...?)

cheungpat:
[...]

PR #321, opened 2017-03-16
Create master user when master key is provided

b123400:
connect #261 

Ready for review~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ DIST_DIR = ./dist/
 DIST := skygear-server
 VERSION := $(shell git describe --always --tags)
 GO_BUILD_LDFLAGS := -ldflags "-X github.com/skygeario/skygear-server/pkg/server/skyversion.version=$(VERSION)"
-GO_TEST_TIMEOUT := 1m
+GO_TEST_TIMEOUT := 1m30s
 OSARCHS := linux/amd64 linux/386 linux/arm windows/amd64 windows/386 darwin/amd64
 
 ifeq (1,${WITH_ZMQ})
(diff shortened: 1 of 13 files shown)


rickmak:
I remember someone want this:
> every usernames are equal just like everyone is equal

So a better solution can be: `asUserID` is `_god` at plugin runtime by default. i.e. If the `skygear-node` or `py-skygear` runtime supply `_god` as `asUserID` attribute if it is not supplied by the developer. Of course in cloud deployment it will always set to `_god`.

So in skygear-server, the logic will become `create a user if not exists when accessing with masterkey`

Though?

b123400:
Oh I made it partially similar to your suggestion.

- If master key is used and no user is supplied, a `_god` user is create on the server side.
- The created user has `_id = _god` and empty username

rickmak:
Yes. I know.

My question is are you willing to remove the hard code here and put it in `plugin`. Do you thinks it is more beautiful? And API wise more versatile.

b123400:
I think implicit user creation should be avoided. We are doing this because we need a user to fill in the `created_by` `updated_by` columns, it is an internal process, that I hope the plugin wouldn't need to know.

If we allow plugins to create new users via `asUserID`, we will have implicit user creation which I think is confusing, and not useful. If developer wants an account, they should register, not specifying a unused id in `asUserID`

rickmak:
> If developer wants an account, they should register

This is exactly what issue #261 telling us the user don't want to register an account. They just want their record being saved and associated with the device/user.

As far as I know, #261 is coming from IoT request. It will be really glad if @david90 can provide a use-case for discussion.

ben181231:
> If we allow plugins to create new users via `asUserID`, we will have implicit user creation which I think is confusing, and not useful.

I don't like implicit user creation either. That seems quite confusing to users, unless we have enough documentation to emphasize it. But I cannot think of a better solution than that and it fits IoT use cases a lot.

b123400:
Ok, I will create user implicitly

PR #324, opened 2017-03-21
Create user when master key is provided

b123400:
connect #261

### DIFF
--- a/pkg/server/preprocessor/preprocessor.go
+++ b/pkg/server/preprocessor/preprocessor.go
@@ -15,6 +15,7 @@
 package preprocessor
 
 import (
+	"context"
 	"net/http"
 	"strings"
 	"time"
@@ -50,16 +51,29 @@ func isTokenStillValid(token router.AccessToken, userInfo skydb.UserInfo) bool {
 
 func (p InjectUserIfPresent) Preprocess(payload *router.Payload, response *router.Response) int {
 	if payload.UserInfoID == "" {
-		log.Debugln("injectUser: empty UserInfoID, skipping")
-		return http.StatusOK
+		if !payload.HasMasterKey() {
+			log.Debugln("injectUser: empty UserInfoID, skipping")
+			return http.StatusOK
+		}
+		payload.UserInfoID = "_god"
+		payload.Context = context.WithValue(payload.Context, router.UserIDContextKey, "_god")
 	}
 
 	conn := payload.DBConn
 	userinfo := skydb.UserInfo{}
 	if err := conn.GetUser(payload.UserInfoID, &userinfo); err != nil {
-		log.Errorf("Cannot find UserInfo.ID = %#v\n", payload.UserInfoID)
-		response.Err = skyerr.NewError(skyerr.UnexpectedUserInfoNotFound, err.Error())
-		return http.StatusInternalServerError
+		if err == skydb.ErrUserNotFound && payload.HasMasterKey() {
+			userinfo = skydb.UserInfo{
+				ID: payload.UserInfoID,
+			}
+			if err := payload.DBConn.CreateUser(&userinfo); err != nil && err != skydb.ErrUserDuplicated {
+				return http.StatusInternalServerError
+			}
+		} else {
+			log.Errorf("Cannot find UserInfo.ID = %#v\n", payload.UserInfoID)
+			response.Err = skyerr.NewError(skyerr.UnexpectedUserInfoNotFound, err.Error())
+			return http.StatusInternalServerError
+		}
 	}
 
 	// If an access token exists checks if the access token has an IssuedAt
--- a/pkg/server/preprocessor/preprocessor_test.go
+++ b/pkg/server/preprocessor/preprocessor_test.go
@@ -303,5 +303,47 @@ func TestInjectUserProcessor(t *testing.T) {
 			So(pp.Preprocess(&payload, &resp), ShouldEqual, http.StatusUnauthorized)
 			So(resp.Err.Code(), ShouldEqual, skyerr.AccessTokenNotAccepted)
 		})
+
+		Convey("should create and inject user when master key is used", func() {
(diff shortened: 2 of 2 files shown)


oursky-travis:
@b123400, thanks for your PR! By analyzing the history of the files in this pull request, we identified @rickmak, @cheungpat and @limouren to be potential reviewers.

cheungpat:
I am going to review this pull request when it is ready for review.

rickmak:
@b123400 Please try to make only one assignees after you create the PR.

cheungpat:
@rickmak The extra reviewer assignments was made automatically by a bot. I think he just didn't come around to update it again after the bot did.

b123400:
Am I expected to reset what the bot did?

cheungpat:
> Am I expected to reset what the bot did?

You are expected to override the bot’s suggestions when you see fit. (I mean... if you see that it is wrong of course you should change it, but in this particular instance I suppose you created the PR and then you didn’t wait for bot to come around and add assignees? Me don’t usually wait either. So I assumed you just didn’t notice the bot added more assignees.)

I am not fond of strict processes as long as PRs are merged and bugs are squashed.

limouren:
Poor bot...

rickmak:
poor chima...

pkg/server/preprocessor/auth.go
```diff
@@ -113,9 +114,18 @@ func (p *UserAuthenticator) Preprocess(payload *router.Payload, response *router
 	// For master access key, it is possible to impersonate any user of
 	// the caller's choosing.
 	if payload.HasMasterKey() {
-		if userID, ok := payload.Data["_user_id"].(string); ok {
-			payload.UserInfoID = userID
-			payload.Context = context.WithValue(payload.Context, router.UserIDContextKey, userID)
+		userID, ok := payload.Data["_user_id"].(string)
+		if !ok {
```
cheungpat: This wouldn’t work as I would expect in these situations:

* If payload contains `_user_id` key and it is an empty string (`userID` is empty string and `ok` is `true`)
* If payload contains `_user_id` key and it is something that is not a string (integer or map), should the server ignores the input and assume `_god` in this situation?

pkg/server/preprocessor/auth.go
```diff
@@ -113,9 +114,18 @@ func (p *UserAuthenticator) Preprocess(payload *router.Payload, response *router
...
+			userID = "_god"
+		}
+		payload.UserInfoID = userID
+		payload.Context = context.WithValue(payload.Context, router.UserIDContextKey, userID)
+		// Try to create the implicitly user if it does not exist
+		info := skydb.UserInfo{
+			ID: userID,
+		}
+		if err := payload.DBConn.CreateUser(&info); err != nil && err != skydb.ErrUserDuplicated {
```
cheungpat: I’d say it is the “inject user” preprocessor’s job to create a user, this is because:

* so the auth preprocessor does not have to depends on existence of a db connection
* the inject user preprocessor checks if there is a user already, and the user can be created if the user does not exists
* try creating user in each master key authenticated requests would generate many constraint violation at the db side and might result in unnecessary log entries
* constraint violation might have non-trivial performance penalty? (I am not sure about this one)

pkg/server/preprocessor/auth_test.go
```diff
@@ -97,12 +100,33 @@ func TestUserAuthenticator(t *testing.T) {
...
+			So(resp.Err, ShouldBeNil)
+
+			conn := payload.DBConn.(*skydbtest.MapConn)
+			_, ok := conn.UserMap["_god"]
+			So(ok, ShouldBeTrue)
+		})
+
+		Convey("test master key and new user_id will create user", func() {
 			payload.Data["api_key"] = "master-key"
+			payload.Data["_user_id"] = "whatever"
```
cheungpat: Might need to add a test case in case the `_user_id` key has non-string value.

Makefile
```diff
@@ -2,7 +2,7 @@ DIST_DIR = ./dist/
 DIST := skygear-server
 VERSION := $(shell git describe --always --tags)
 GO_BUILD_LDFLAGS := -ldflags "-X github.com/skygeario/skygear-server/pkg/server/skyversion.version=$(VERSION)"
-GO_TEST_TIMEOUT := 1m
+GO_TEST_TIMEOUT := 1m30s
```
cheungpat: The commit message says to see if the test would pass if the timeout is extended. Is this still needed after you finished your changes?

pkg/server/preprocessor/auth.go
b123400: Oh sorry, I just looked at the preprocessor folder and decided `auth.go` seems to be the most relevant place to add function, without realising there is inject user. After moving things to inject user, issues in your other comments are gone!

pkg/server/preprocessor/preprocessor.go
```diff
@@ -50,16 +51,29 @@ func isTokenStillValid(token router.AccessToken, userInfo skydb.UserInfo) bool {
...
+		payload.Context = context.WithValue(payload.Context, router.UserIDContextKey, "_god")
 	}
 
 	conn := payload.DBConn
 	userinfo := skydb.UserInfo{}
 	if err := conn.GetUser(payload.UserInfoID, &userinfo); err != nil {
-		log.Errorf("Cannot find UserInfo.ID = %#v\n", payload.UserInfoID)
-		response.Err = skyerr.NewError(skyerr.UnexpectedUserInfoNotFound, err.Error())
-		return http.StatusInternalServerError
+		if payload.HasMasterKey() {
```
cheungpat: It appears that you are assuming that `GetUser` returns an error iff the user is not found. In reality an error could also occurs for other kinds of error such as other database problems. I think it is better to check the error type to make sure the user does not exist before creating a new user.

b123400:
Updated~

cheungpat:
Thanks.

Issue #327, opened 2017-03-22
Missing asset content type in response of record save


ben181231:
When save a record with asset typed field, `$content_type` of the field is expected. Actual, it returns empty string.

```shell
$ curl \
-X POST \
-H "Content-Type: application/json" \
-d '{
    "action": "record:save",
    "api_key": "changeme",
    "access_token": "***",
    "database_id": "_public",
    "records": [
    	{
    		"_id": "Demo\/001",
    		"_access": [
    			{
    				"public": true,
    				"level": "read"
    			}
    		],
    		"img": {
    			"$type": "asset", 
    			"$name":"dabfeed9-6993-4130-a825-9f010b037ce7-Record-Image",
    			"$content_type":"image\/jpeg",
    			"$url":"http:\/\/localhost:3001\/files\/dabfeed9-6993-4130-a825-9f010b037ce7-Record-Image"
    		}
    	}
	]
}
' "http://localhost:3001/" | python -m json.tool

{
    "result": [
        {
            "_access": [
                {
                    "level": "read",
                    "public": true
                }
            ],
            "_created_at": "2017-03-22T09:37:54.622852Z",
            "_created_by": "7546ea5a-f993-4592-b274-45a2bc6c54bc",
            "_id": "Demo/001",
            "_ownerID": "7546ea5a-f993-4592-b274-45a2bc6c54bc",
            "_type": "record",
            "_updated_at": "2017-03-22T09:47:19.185657Z",
            "_updated_by": "7546ea5a-f993-4592-b274-45a2bc6c54bc",
            "img": {
                "$content_type": "",
                "$name": "dabfeed9-6993-4130-a825-9f010b037ce7-Record-Image",
                "$type": "asset",
                "$url": "http://localhost:3001/files/dabfeed9-6993-4130-a825-9f010b037ce7-Record-Image"
            }
        }
    ]
}
```

rickmak:
@b123400 we will need this issue to make a minor release. Do you think it can be done by tomorrow? Than I will merge the zmq-multiplex to master.

b123400:
Fixed, though I think it is not completely what you are expecting.

There are 2 issues here:

1. Despite the content of the request, the server wasn't returning `$content_type`. e.g. You don't need to specify "img" in the above request body, the bug still happens.
2. Record's assets are not updated accordingly. (e.g. The "$content_type" you specified in the payload is not going to update the value in database)

I have fixed (1), but not (2).
[...]

PR #328, opened 2017-03-22
Return asset content type in response of record save

b123400:
connect #327 

This PR makes sure asset's information is complete, but [it is not perfect](https://github.com/SkygearIO/skygear-server/issues/327#issuecomment-288522101)

### DIFF
--- a/pkg/server/handler/record_test.go
+++ b/pkg/server/handler/record_test.go
@@ -376,6 +376,13 @@ func TestRecordSaveDataType(t *testing.T) {
 	Convey("RecordSaveHandler", t, func() {
 		db := skydbtest.NewMapDB()
 		conn := skydbtest.NewMapConn()
+		conn.AssetMap = map[string]skydb.Asset{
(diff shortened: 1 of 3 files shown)


oursky-travis:
@b123400, thanks for your PR! By analyzing the history of the files in this pull request, we identified @cheungpat, @limouren and @rickmak to be potential reviewers.

b123400:
Updated to inject signer after querying asset.

Not sure if it's golang issue or our problem, but the current structure make things super easy to be invalid. If I want to query a record, I have to remember to inject the assets, and then inject the signer, forget anyone them will give no warning at all.

rickmak:
There should be an runtime error `log.Warnf("Unable to generate signed url of asset because no singer is injected.")`, refs: https://github.com/SkygearIO/skygear-server/blob/master/pkg/server/skydb/record.go#L196

Or you want an exception? Reason is?

rickmak:
Or you are expecting compiling error?

PR #331, opened 2017-03-28
Make content type optional at incoming JSON

rickmak:
follow up of #287.
Making it compulsory will break old SDK and not necessary, since the server will
guess mimeType in Handler


oursky-travis:
@rickmak, thanks for your PR! By analyzing the history of the files in this pull request, we identified @limouren, @cheungpat and @ben181231 to be potential reviewers.

Issue #333, opened 2017-03-29
skygear.every result in nil pointer dereference


rickmak:
- Skygear Server Date/Version: v0.22.1
- Application Platform: py-skygear v0.22.1
- [x] Is this a regression?
- [x] Attached logs, screenshots

last working skygear version is v0.21.0.

With following cloud code.

```
@skygear.every("@every 5s")
def five_second():
    log.warn('time wasted 5 seconds')
```

Result in follow exception
```
ERRO[0034] panic occurred while calling plugin ipc       logger=plugin recovered=runtime error: invalid memory address or nil pointer dereference
ERRO[0034] goroutine 26 [running]:
runtime/debug.Stack(0xc42015f880, 0xc4202e5a60, 0x1)
	/usr/local/Cellar/go/1.7.1/libexec/src/runtime/debug/stack.go:24 +0x79
github.com/skygeario/skygear-server/pkg/server/plugin/zmq.(*zmqTransport).ipc.func1()
	/go/src/github.com/skygeario/skygear-server/pkg/server/plugin/zmq/rpc.go:152 +0x119
panic(0x456e920, 0xc42000c0e0)
	/usr/local/Cellar/go/1.7.1/libexec/src/runtime/panic.go:458 +0x243
github.com/skygeario/skygear-server/pkg/server/plugin/zmq.(*zmqTransport).ipc(0xc42023e000, 0xc42015f800, 0x0, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/skygeario/skygear-server/pkg/server/plugin/zmq/rpc.go:172 +0x1df
github.com/skygeario/skygear-server/pkg/server/plugin/zmq.(*zmqTransport).rpc(0xc42023e000, 0xc42015f800, 0x0, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/skygeario/skygear-server/pkg/server/plugin/zmq/rpc.go:112 +0x50
github.com/skygeario/skygear-server/pkg/server/plugin/zmq.(*zmqTransport).RunTimer(0xc42023e000, 0xc420275ac0, 0x12, 0x48ed950, 0x0, 0x0, 0x4172f91, 0x180001, 0x0, 0x0, ...)
	/go/src/github.com/skygeario/skygear-server/pkg/server/plugin/zmq/rpc.go:94 +0xd4
github.com/skygeario/skygear-server/pkg/server/plugin.(*Plugin).initTimer.func1()
[...]

PR #334, opened 2017-03-29
Fix plugin timer request not properly initialized

cheungpat:
refs #333


oursky-travis:
@cheungpat, thanks for your PR! By analyzing the history of the files in this pull request, we identified @rickmak, @limouren and @ben181231 to be potential reviewers.

cheungpat:
Probably not a good idea to create a pull request using github web interface.

PR #337, opened 2017-03-30
Fix plugin timer request not properly initialized

cheungpat:
connects #333


cheungpat:
Don’t know how to write unit test to guard against this particular kind of bug without refactoring the transport interface. @rickmak  New issue #338 opened for that.

rickmak:
Merge via cb71a30

Issue #262, opened 2017-01-10
Support PostGIS GeoJSON and ST_Contains


lampercy:
### To save polygon
```
const polygon = new skygear.GeoJSON({
  "type": "Polygon",
  "coordinates": [[
    [100.0, 0.0], [101.0, 0.0], [101.0, 1.0], [100.0, 1.0], [100.0, 0.0]
  ]]
});
```

### To search geometry by location
```
const query = new skygear.Query(District);
const geoLocation = new skygear.Geolocation(22.283, 114.15);
query.containsValue('area', geoLocation);
```

### To search location by geometry
```
const place = new Place({
  name: 'IFC HK',
  location: geoLocation,
});
skygear.publicDB.save(place);

const query = new skygear.Query(Place)
query.contains('location', polygon); 
```

### To search geometry by geometry
```
const query = new skygear.Query(District);
query.containsValue('area', polygon);
```

### Point
```
{
  "type": "Point", 
  "coordinates": [102.0, 0.5]
}
```

### LineString
```
{
  "type": "LineString",
  "coordinates": [
    [102.0, 0.0], [103.0, 1.0], [104.0, 0.0], [105.0, 1.0]
  ]
}
```

### Polygon with interior rings
```
{ "type": "Polygon",
  "coordinates": [
    [[100.0, 0.0], [101.0, 0.0], [101.0, 1.0], [100.0, 1.0], [100.0, 0.0]],
    [[100.2, 0.2], [100.8, 0.2], [100.8, 0.8], [100.2, 0.8], [100.2, 0.2]]
  ]
}
```

### Multipolygon
```
{ "type": "MultiPolygon",
  "coordinates": [
    [[[102.0, 2.0], [103.0, 2.0], [103.0, 3.0], [102.0, 3.0], [102.0, 2.0]]],
    [[[100.0, 0.0], [101.0, 0.0], [101.0, 1.0], [100.0, 1.0], [100.0, 0.0]],
     [[100.2, 0.2], [100.8, 0.2], [100.8, 0.8], [100.2, 0.8], [100.2, 0.2]]]
    ]
}
```

cheungpat:
You also need to wrap the GeoJSON object into a skygear value type object.

A few examples: https://github.com/SkygearIO/skygear-server/blob/master/pkg/server/skydb/skyconv/map.go

Perhaps something like `{"$type": "geojson", "$value": {"type": "MultiPolygon" /* ... */}}`.

[...]

PR #340, opened 2017-04-03
Implement GeoJSON query using ST_Contains

cheungpat:
I am creating this WIP PR to request for comment. The pull request has some test case and I expect them to pass, but I don’t have much confidence in the quality of the code at the moment. Rather than me vetting the code for longer, I would like to have comments for specific parts to improve since I have already spent too much time then expected.

@rickmak @lampercy please comment. thanks.

connects #262


cheungpat:
please review @rickmak

rickmak:
We need to modify the the `handler/query.go` for parsing the query?

I try this predicate and it fails to parse at parser:

```
[
     "in",
     {
       "$type": "geojson",
       "$val": {
         "type": "Point", 
         "coordinates": [102.0, 0.5]
       }
     },
     {
       "$type": "keypath",
       "$val": "geojson"
     }
]
```

I will continue to review tomorrow.... need sometime to load the model back to my mind.

cheungpat:
> We need to modify the the handler/query.go for parsing the query?

Yes. Missed that part.

cheungpat:
The new commits should fix the problem.

Issue #319, opened 2017-03-15
DEV_MODE behaviour is inconsistent with integer type column


louischan-oursky:
`DEV_MODE` behaviour is inconsistent.

- Skygear Server Date/Version: v0.22.1
- Application Platform:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Steps to reproduce

1. Create a skygear-like table with one column set to postgresql data type `integer`
2. Use sdk api `save` to save a record.

#### Actual Results
When `DEV_MODE=YES`, saving to a table containing `integer` column is fine and the column does **NOT** auto migrate to `double precision`.
When `DEV_MODE=NO`, skygear server complains it cannot perform migration.

#### Expected Results
I think skygear-server should complain that `integer` is never a supported data type regardless of `DEV_MODE` is on or off.

cheungpat:
I think the desired result should be that when the column has `integer` type, the following should happens, regardless of `DEV_MODE` settings:

* when saving a record with field value that can be interpreted as integer without losing precision, saving should be successful
* otherwise, the saving should result in an error (probably incompatible schema/data type (forgot the actual name))

cheungpat:
On the other hand, I am guessing what @louischan-oursky meant when he said the behaviour is inconsistent. I haven’t looked at the code yet, but I suppose that when `DEV_MODE=NO`, `integer` is considered incompatible with `double precision` while when `DEV_MODE=YES`, it is considered compatible. We could probably improve the compatibility check.

PR #341, opened 2017-04-05
Fix inconsistent behavior saving to integer column

cheungpat:
When saving to integer column, the request my fail depending on different
dev mode settings. This is due to separate checking involved when
determining if two record schemas are compatible.

connects #319


oursky-travis:

rickmak:
Merge via 98c7f3b

PR #346, opened 2017-04-08
Revamp query sqlizer into its own package

cheungpat:


oursky-travis:

Issue #347, opened 2017-04-11
Plugin HTTP handler or Lambda not getting called, after triggering after_save with async=True


Steven-Chan:
Skygear Server Date/Version: v0.22.1, python plugin with zmq

#### Steps to reproduce

1. client call save record
2. trigger after_save in plugin
3. client call http handler or lambda
4. the handler function is not called
5. plugin time out after a while

Methods to make the handler work:
either:
- use http instead of zmq
- set async=False

Steven-Chan:
plugin:
```python
import logging
import skygear

log = logging.getLogger(__name__)


@skygear.handler('handler', method=['POST'])
def handler(req):
    return {
        'result': 'handler is OK'
    }


@skygear.after_save('recordA', async=True)
def recordA_async_after_save(record, original_record, conn):
    log.info('recordA_async_after_save')
```

client.js
```js
const skygear = require('skygear');
const agent = require('superagent');

const RecordA = skygear.Record.extend('recordA');

skygear.config({
  'apiKey': 'apikey',
  'endPoint': 'http://localhost:3003/',
}).then(() => {
  return skygear.signupAnonymously();
}).then(() => {
  const r = new RecordA({ 'content': 'Hello World' });
  return skygear.publicDB.save(r, {atomic: true});
}).then(() => {
  console.log('save success');
  agent.post('http://localhost:3003/handler').end((err, res) => {
    if (err) {
      console.log('http handler err');
    } else {
      console.log('http handler success');
    }
  });
}).catch(() => {
  console.log('save failed');
});
```

b123400:
Cause identified.

[ZMQ broker expect the response channel to be consumed.](https://github.com/SkygearIO/skygear-server/blob/6f651fba7b39c6359b6055b85d087142ee0491d4/pkg/server/plugin/zmq/broker.go#L220-L230)
[zmqTransport does not always consume response channel.](https://github.com/SkygearIO/skygear-server/blob/6f651fba7b39c6359b6055b85d087142ee0491d4/pkg/server/plugin/zmq/rpc.go#L162-L172)

When a request is ended before async handler responds, transport gives up consuming from broker. Broker keeps waiting for the channel to be consumed.

Working on a fix.

chpapa:
@Steven-Chan Can you help verify the pull request fixed your issue? If yes please close it.

rickmak:
Steven verified it is fixed.

PR #348, opened 2017-04-13
Remove response channel when it is not consumed anymore

b123400:
connect https://github.com/SkygearIO/skygear-server/issues/347

### DIFF
--- a/pkg/server/plugin/zmq/rpc.go
+++ b/pkg/server/plugin/zmq/rpc.go
@@ -159,8 +159,9 @@ func (p *zmqTransport) ipc(req *pluginrequest.Request) (out []byte, err error) {
 
 	reqChan := make(chan chan []byte)
 	p.broker.RPC(reqChan, in)
+	respChan := <-reqChan
 	select {
-	case msg := <-<-reqChan:
+	case msg := <- respChan:
 		// Broker will sent back a null byte if time out
 		if bytes.Equal(msg, []byte{0}) {
 			err = fmt.Errorf("Plugin time out")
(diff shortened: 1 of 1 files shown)


oursky-travis:
@b123400, thanks for your PR! By analyzing the history of the files in this pull request, we identified @cheungpat to be a potential reviewer.

pkg/server/plugin/zmq/broker.go
```diff
@@ -286,6 +286,15 @@ func (lb *Broker) setTimeout(address string) {
 	lb.timeout <- address
 }
 
+func (lb *Broker) cancelRPC(requestChan chan []byte) {
+	for k, v := range lb.addressChan {
```
cheungpat: Only one coroutine should be modifying `lb.addressChan`.

pkg/server/plugin/zmq/rpc.go
```diff
@@ -159,15 +159,17 @@ func (p *zmqTransport) ipc(req *pluginrequest.Request) (out []byte, err error) {
...
-	case msg := <-<-reqChan:
+	case msg := <- respChan:
 		// Broker will sent back a null byte if time out
 		if bytes.Equal(msg, []byte{0}) {
 			err = fmt.Errorf("Plugin time out")
 		} else {
 			out = msg
 		}
 	case <-req.Context.Done():
+		p.broker.cancelRPC(respChan)
```
cheungpat: This doesn’t seem to be robust because the context can be done and the request can finish at the same time.

b123400:
Right, I think I will choose the prior

> Hold the guarantee that the caller will always consume from the channel. This requires creating a goroutine that always consume from the channel regardless of whether context is done.

b123400:
Updated

Issue #342, opened 2017-04-05
Chat plugin does not work with zmq-multiplex


b123400:
OK, finally get the reason.
Chat plugin uses few after_save hooks with `async=True`, if I set all of them to `async=False` it works fine.
The current protocol assumes requested to be nested, if a child request does not end, the parent request cannot end. This is not true when `async=True`.

I can think of few directions to solve the problem.

1. Force `async=False` when using a zmq-multiplex
2. When `async=True`, execute hooks with another worker (not reusing worker)
    - This means we might go back to the [original issue](https://github.com/SkygearIO/skygear-server/issues/295), workers might be used up when there are a lot of async hooks
3. When `async=True`, execute hooks with newly created worker
    - This requires quite some changes because workers are currently spawned at the plugin side
4. When the plugin is executing a hook with `async=True`, it should internally queue up all messages, reorder them to make them sync again.
    - Better than (1). This way golang will still perform hooks in async, so no slow down at skygear-server side
    - The hooks are sync, arguably a not exactly what `async=True` means, it is a complicated behaviour for user to understand
    - Difficult to maintain

b123400:
I just updated the issue with what I have found in the zmq + chat runtime. Any advise...?
@cheungpat @rickmak @ben181231

cheungpat:
I would opt for choice 4 and I agree with you that it is the most difficult one.

I think we should not allow a request to spawn coroutines that 1) run concurrently, and 2) call cloud code. This will potentially consume workers  in a short amount of time and worker is a limited resource for non-async python.

I think to have hook run `async=True` is for having the hook to run after the database transaction commits and the request does not have to wait for the hook to finish. If I am wrong, others could state what their expectations maybe.

We still need to discuss if you go for option 4 but I am pretty sure we do not break people’s assumption. But option 1 definitely does.

rickmak:
> async=True is for having the hook to run after the database transaction commits and the request does not have to wait for the hook to finish.

Yes. Here we are kind of mix-up concept of blocking and async. Async implies non-blocking. Sync implies blocking.

So for 4, it will purely become non-blocking sync operation. Which may be fine, it may actually don't break how people currently understand and options. And how we describe it to user in [doc](https://docs.skygear.io/guides/cloud-function/database-hooks/python/)

b123400:
> So for 4, it will purely become non-blocking sync operation. Which may be fine, it may actually don't break how people currently understand and options. And how we describe it to user in doc

[...]

PR #349, opened 2017-04-14
Fix zmq issue with chat plugin

b123400:
connect https://github.com/SkygearIO/skygear-server/issues/342

This PR makes sure when server is calling an async hook, it does not reuse worker.

After fixing the bug, chat still does not work because of another bug: Heartbeat will put workers back to the pool, which is not preferred because it may be waiting for a response of a nested request. This is fixed too.

### DIFF
--- a/pkg/server/plugin/zmq/rpc.go
+++ b/pkg/server/plugin/zmq/rpc.go
@@ -79,8 +79,8 @@ func (p *zmqTransport) RunHandler(ctx context.Context, name string, in []byte) (
 	return
 }
 
-func (p *zmqTransport) RunHook(ctx context.Context, hookName string, record *skydb.Record, originalRecord *skydb.Record) (*skydb.Record, error) {
-	out, err := p.rpc(pluginrequest.NewHookRequest(ctx, hookName, record, originalRecord))
+func (p *zmqTransport) RunHook(ctx context.Context, hookName string, record *skydb.Record, originalRecord *skydb.Record, async bool) (*skydb.Record, error) {
+	out, err := p.rpc(pluginrequest.NewHookRequest(ctx, hookName, record, originalRecord, async))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


oursky-travis:
@b123400, thanks for your PR! By analyzing the history of the files in this pull request, we identified @cheungpat to be a potential reviewer.

rickmak:
Why is the py-skygear send back heartbeat if the process is not finished? Sending heartbeat represent the worker is ready to accept another jobs.

It should be a bugs in py-skygear but not skygear-server.

pkg/server/plugin/zmq/rpc.go
```diff
@@ -169,9 +169,16 @@ func (p *zmqTransport) ipc(req *pluginrequest.Request) (out []byte, err error) {
 		return
 	}
 
+	workerIDs, needReuseWorker := req.Context.Value(ZMQWorkerIDsContextKey).(map[string]string)
```
rickmak: Reusing variable is generally not a good idea.

b123400:
> Why is the py-skygear send back heartbeat if the process is not finished? Sending heartbeat represent the worker is ready to accept another jobs.

I thought heartbeat means "I am not dead", instead of "I am ready to accept a job". py-skygear sends heartbeats because it is waiting for the response of a nested request, so it is actually "not dead, can receive a nested request or response".

If you think heartbeat should not happen in this case, I can modify them

rickmak:
Not I think a heartbeat should or should not happen in this case, but will it help on simplifying the actual code?

The heartbeat is actually aligning with "I am ready to accept job" in master. While I agree the word "heartbeat" mean "I am not dead" conceptually. What is the use to know a worker is not dead?

Separating this two concept will increase the code complexity while it gives little benefit. I believe the parcel already contains the address information. Together with the request time out mechanic, assuming the worker is alive within the nested request cycle will greatly simplify the code and make it more maintainable.


This PR, with d189218 will increasingly complexity of workerQueue interface. It makes the object become difficult to understand and less maintainable.

Let look back at the original interface.
- `Len`, `Next`, `Ready` and `Purge`
This is a very standard queue interface. Easy to understand.

At some point, I rename the `Ready` to `Tick` and introduce `Remove`. In retrospect, the rename is not a great choice.

If we add `Borrow` and `Renew`, it become non-standard queue and not much gain of features.

b123400:
OK, let me fix it at python's side

b123400:
Updated PR, python's PR is [here](https://github.com/SkygearIO/py-skygear/pull/136)

rickmak:
Merge via e9cca72

PR #360, opened 2017-04-26
Fix schema:default_access inline doc

cheungpat:


oursky-travis:
@cheungpat, thanks for your PR! By analyzing the history of the files in this pull request, we identified @ben181231 to be a potential reviewer.

ben181231:
Picked as bd68b0e

Issue #277, opened 2017-01-27
Abort the DB connection on API timeout


rickmak:
This is a follow up of #271.

The API will time out now. But the respective DB connection will not abort, but waiting the DB connection to return.  In the case of deadlock, the DB connection will not return and the resource will hold up. It will be better for skygear-server to actively abort the long running DB connection.

cheungpat:
Golang support cancelable queries in go1.8

rickmak:
go1.8 is scheduled to be released on Feb 14th/15th/16th

refs: https://groups.google.com/forum/#!topic/golang-dev/G09HaWfMl6g

cheungpat:
first step: #282

cheungpat:
https://github.com/golang/go/releases/tag/go1.8

cheungpat:
fixed in #366

PR #366, opened 2017-05-05
Support for cancellable query

cheungpat:
connects #277 

Skygear Server is now built with golang 1.8.1.


pkg/server/skydb/pq/pq.go
```diff
@@ -71,14 +72,14 @@ func isNetworkError(err error) bool {
 	return ok
 }
 
-type queryxRunner interface {
-	Exec(query string, args ...interface{}) (sql.Result, error)
-	Queryx(query string, args ...interface{}) (*sqlx.Rows, error)
-	QueryRowx(query string, args ...interface{}) *sqlx.Row
+type queryxContextRunner interface {
```
rickmak: What is the use of this interface...?

cheungpat: Interesting question. This interface ceased to have any use in skydb.

cheungpat:
> Will we provide default timeout context?

When router handle a request, it creates a context with a timeout of 60s and this context is passed the the database connection struct in the conn_db preprocessor. So database query will be cancelled when the request is timed out.

pkg/server/skydb/pq/pq.go
rickmak: Let's remove it?

cheungpat: Interesting. I thought I pushed something.

Issue #371, opened 2017-06-02
Record type should be alphanumeric


cheungpat:
The record type should only consist of alphanumeric string that begins with a alpha character.

cheungpat:
It does look like to be intentional for record type to accept non-alphanumeric characters. https://github.com/SkygearIO/skygear-server/blob/6f651fba7b39c6359b6055b85d087142ee0491d4/pkg/server/skydb/pq/schema_test.go#L106

PR #372, opened 2017-06-02
Force record type to be alphanumeric + underscore

cheungpat:
connects #371


oursky-travis:

cheungpat:
It does look like to be intentional for record type to accept non-alphanumeric characters. https://github.com/SkygearIO/skygear-server/blob/6f651fba7b39c6359b6055b85d087142ee0491d4/pkg/server/skydb/pq/schema_test.go#L106

Issue #375, opened 2017-06-05
Create field-based ACL model


ben181231:
Refer to the [field-based ACL feature](https://github.com/SkygearIO/features/blob/master/features/48-user-profile-discovery/field-based-acl.md), this issue includes:

- Models related to field-based ACL (no need to take effect in this issue):
  - Accessibility levels (`ReadWrite`, `ReadOnly` and `NoAccess`)
  - Discovery levels (`Queryable`, `Discoverable` and `NotQueryable`)
  - Field-based Access Control Entity
  - User Roles
- Database migration
- Endpoints for field-based ACL:
  - Getting field-based ACL
  - Updating field-based ACL

cheungpat:
I will take care of the model and database migration first and handle the endpoints later.

PR #381, opened 2017-06-09
Implement db and struct for FieldACL

cheungpat:
Handler will come later.

refs #375


oursky-travis:

pkg/server/skydb/access.go
```diff
@@ -126,3 +130,192 @@ func (acl RecordACL) Accessible(userinfo *UserInfo, level RecordACLLevel) bool {
...
+		(entry.RecordField != field && entry.RecordField != WildcardRecordField) {
+		return false
+	}
+
+	// TODO: Check access here
+
+	return (mode == ReadFieldAccessMode && entry.Readable) ||
+		(mode == WriteFieldAccessMode && entry.Writable) ||
+		(mode == CompareFieldAccessMode && entry.Comparable) ||
+		(mode == DiscoverFieldAccessMode && entry.Discoverable)
```
ben181231: Should we handle the following situation:

1. `mode` is `ReadFieldAccessMode`, but `entry.Readable` is `false`, `entry.Writable` is `true` ?
1. `mode` is `DiscoverFieldAccessMode`, but `entry.Discoverable` is `false`, `entry.Comparable` is `true` ?

ben181231: Or this inconsistency should be blocked from SDK ?

pkg/server/skydb/access.go
```diff
@@ -126,3 +130,192 @@ func (acl RecordACL) Accessible(userinfo *UserInfo, level RecordACLLevel) bool {
...
+
+	// CompareFieldAccessMode means the access mode is for query
+	CompareFieldAccessMode
+)
+
+// FieldACL contains all field ACL rules for all record types
+type FieldACL struct {
+	wildcardRecordType FieldACLEntryList
+	recordTypes        map[string]FieldACLEntryList
+}
```
ben181231: It is necessary to keep 1 dictionary + 1 array ? Or we can improve the sorting methods of `FieldACLEntryList` to keep it just 1 array ?

cheungpat: We could have used a single dictionary. Using a single array works but that is not efficient IMO. Using a map to find the first matching record type is O(1). Using an array to find the first matching record type is O(n).

pkg/server/skydb/access.go
```diff
@@ -126,3 +130,192 @@ func (acl RecordACL) Accessible(userinfo *UserInfo, level RecordACLLevel) bool {
...
+	if result != 0 {
+		return result < 0
+	}
+
+	result = compare(list[i].RecordField, list[j].RecordField, WildcardRecordField)
+	if result != 0 {
+		return result < 0
+	}
+
+	return strings.Compare(list[i].UserRole, list[j].UserRole) < 0
```
cheungpat: I am working on a branch that uses struct in `UserRole`. So I will mark this as TODO.

pkg/server/skydb/access.go
cheungpat: > Should we handle the following situation:

Both of these situation assumes Writable rule is a superset of the Readable rule. Are we taking this assumption? If we are taking this assumption perhaps we should use a single column for read/write and another column for compare/discover?

pkg/server/skydb/access.go
ben181231: Single dictionary is also good to me.

pkg/server/skydb/access.go
ben181231: Okay

pkg/server/skydb/access.go
cheungpat: > Or this inconsistency should be blocked from SDK ?

If you think it is an inconsistency perhaps we should not design the database table this way?

ben181231: In the [feature proposal](https://github.com/SkygearIO/features/blob/master/features/48-user-profile-discovery/field-based-acl.md#access-control-entity-ace), the ACL entity only has 2 fields representing accessibility level and discovery level, which implies that `ReadWrite` is superset of `Readable` and `Queryable` is superset of `Discoverable`. 

But in the [API changes section of the proposal](https://github.com/SkygearIO/features/blob/master/features/48-user-profile-discovery/field-based-acl.md#changes-on-api-at-skygear-server), `ReadWrite` is separated into `readable` and `writeable`. So that means that assumption is ensure on SDK and Skygear server will not handle that situation. 

Do you agree?

cheungpat: > Do you agree?

Yes.

cheungpat:
@ben181231 I hope the update has addressed your concerns.

PR #388, opened 2017-06-16
Fix potential problem when saving new record

cheungpat:
recordSaveHandler (recordutil.go) did not check if the fetched
record is a new record or an existing record, causing the
originalRecordMap to contain incorrect data. This PR also removes
the need for a hot fix to populate new record fields.


oursky-travis:
@cheungpat, thanks for your PR! By analyzing the history of the files in this pull request, we identified @limouren to be a potential reviewer.

Issue #376, opened 2017-06-05
Support field-based accessibility control for trivial user roles


ben181231:
Refer to the [field-based ACL feature](https://github.com/SkygearIO/features/blob/master/features/field-based-acl.md), this issue includes the support of field-based accessibility (`ReadWrite`, `ReadOnly` and `NoAccess`) for the following user roles:

- Specific User: the user specified by user ID
- Defined Roles: the user-defined roles
- Any Users: any logged-in users
- Public: any users with correct API key

skygeario/features#48

PR #390, opened 2017-06-19
 Implement read/write Field ACL for fixed user roles

cheungpat:
connects #376
refs skygeario/features#48

### DIFF
--- a/pkg/server/handler/record.go
+++ b/pkg/server/handler/record.go
@@ -279,6 +279,18 @@ func (h *RecordSaveHandler) Handle(payload *router.Payload, response *router.Res
 		return
 	}
 
+	resultFilter := recordResultFilter{
+		AssetStore: h.AssetStore,
+		UserInfo:   payload.UserInfo,
+		FieldACL: func() skydb.FieldACL {
+			acl, err := payload.DBConn.GetRecordFieldAccess()
+			if err != nil {
+				panic(err)
+			}
+			return acl
+		}(),
+	}
+
 	log.Debugf("Working with accessModel %v", h.AccessModel)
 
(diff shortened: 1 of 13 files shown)


oursky-travis:

pkg/server/handler/record.go
```diff
@@ -437,6 +449,17 @@ func (h *RecordFetchHandler) Handle(payload *router.Payload, response *router.Re
 	}
 
 	db := payload.Database
+	resultFilter := recordResultFilter{
```
rickmak: I think confront with the pattern of `newRecordResultFilter` is better?

pkg/server/handler/recordutil.go
```diff
@@ -26,6 +27,45 @@ func injectSigner(record *skydb.Record, store asset.Store) {
...
+		}
+
+		record.Remove(key)
+	}
+
+	if len(nonWritableFields) > 0 {
+		return skyerr.NewDeniedArgument("Unable to save to some record fields because of Field ACL denied update.", nonWritableFields)
+	}
+	return nil
+}
```
rickmak: Since these two methods mutate the `record` internal state, put this logic into record struct make more sense to me.

cheungpat: I hope I am not mistaken, you are talking about `scrubRecordFieldsForRead ` and `scrubRecordFieldsForWrite`, right?

I am not sure how to put the logic under `skydb.Record` because that will make the `skydb.Record` dependent on `skydb.FieldACL` and `skydb.UserInfo`. `skydb.Record` has been simple and it is not dependent on other struct.

PR #399, opened 2017-06-23
[WIP] Add role:assign/role:revoke handler

rickmak:
This PR is WIP on allowing batch operation on users' roles

Finished:
- skydb AssignRole/RevokeRole
- role:assign handler

Pending:
- The return format of the handler
- role:revoke handler
- test case on handler

Note:
For return format of handler, I am thinking should we return the affected entries. Or just success/fails instead of echo. But no time to really think through it. Feel free to comment.


cheungpat:
Superseded by #404

Issue #378, opened 2017-06-05
Support field-based ACL on owner user role


ben181231:
Refer to the [field-based ACL feature](https://github.com/SkygearIO/features/blob/master/features/field-based-acl.md), this issue includes the support of field-based ACL on owner user role.

PR #402, opened 2017-06-27
Support Field ACL for dynamic field and owner

cheungpat:
connects #379
connects #378
refs SkygearIO/features#48

### DIFF
--- a/pkg/server/skydb/access.go
+++ b/pkg/server/skydb/access.go
@@ -196,7 +196,8 @@ func (acl FieldACL) Accessible(
 			return true
 		}
 
-		if !entry.UserRole.Match(userInfo, record) {
+		if !entry.UserRole.Applicable(mode) ||
+			!entry.UserRole.Match(userInfo, record) {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/server/skydb/access.go
```diff
@@ -440,6 +446,35 @@ func (r FieldUserRole) Match(userinfo *UserInfo, record *Record) bool {
...
+	return false
+}
+
+// Applicable returns true if the field user role is able to evaluate
+// the specified access mode.
+func (r FieldUserRole) Applicable(accessMode FieldAccessMode) bool {
+	if r.Type.RecordDependent() {
+		return accessMode == ReadFieldAccessMode ||
+			accessMode == WriteFieldAccessMode
+	}
```
ben181231: Need some comments here about the relation between `RecordDependent`, `accessMode` and `Applicable`.

cheungpat: On second thought, `Match` functions already checked for cases that is covered by the `Applicable` function. This is because if we are doing a query, the `record` var is `nil`, and the `Match` function exists early returning `false`, which is exactly what `Applicable` will return.

cheungpat:
@ben181231 check again

PR #407, opened 2017-06-29
Fix unable to reset FieldACL with empty Field ACL

cheungpat:
Without this change, SetRecordFieldAccess will result in panic because
the function is trying to INSERT without any data.


oursky-travis:
@cheungpat, thanks for your PR! By analyzing the history of the files in this pull request, we identified @ben181231 to be a potential reviewer.

ben181231:
Picked as dcff59a

Issue #393, opened 2017-06-21
New User Profile Design: use _auth as authentication object, update auth api


Steven-Chan:
https://github.com/SkygearIO/features/pull/68

- [ ] New database schema
  - [x] `_auth`
  - [x] `_user`, a view for backward compatibility
- [ ] Migrate database schema for skygear-server version upgrade
  - [x] Rename `_user` table to `_auth`
  - [x] Create a view named `_user` for backward compatibility
  - [x] Rename field `auth` to `provider_info` in `_auth`
  - [x] Move `username` and `email` from `_auth` to `user`
- [ ] Update signup and login api for the updated database schema and env. var.
  - [x] Migrate `user` table according to env. var. `AUTH_RECORD_KEYS` when signup
  - [x] new signup api signature: `signup(authData, password, profile?)`, where `authData` is defined in `AUTH_RECORD_FIELDS` and profile is a dictionary and is optional
  - [x] new login api signature: `login(authData, password)`

Note:
This issue is not taking care of field-base ACL.

## handling `AUTH_RECORD_TYPE` and `AUTH_RECORD_FIELDS` in `user` table migration

Default values:
- AUTH_RECORD_TYPE: `user`
- AUTH_RECORD_FIELDS: `username,email`

Note:
`Migration steps` below means the operations performed by skygear-sever when server version upgraded, and when `AUTH_RECORD_TYPE` and `AUTH_RECORD_FIELDS` may or may not be presented.

### scenario 1:

- AUTH_RECORD_TYPE NOT specified
- AUTH_RECORD_FIELDS NOT specified

Migration steps:
1. Move `username` and `email` from `_auth` to `user`
1. Add unique constraint to `username` to `user`
1. Add unique constraint to `email` to `user`

### scenario 2:

- AUTH_RECORD_TYPE set to `human`
- AUTH_RECORD_FIELDS NOT specified

Migration steps:
1. Move `username` and `email` from `_auth` to `user`
1. Add unique constraint to `username` to `user`
1. Add unique constraint to `email` to `user`
1. Rename table `user` to `human`

### scenario 3:

- AUTH_RECORD_TYPE NOT specified
- AUTH_RECORD_FIELDS set to `username,(username,email)`, meaning `username` is required and `email` is optional

Migration steps:
1. Move `username` and `email` from `_auth` to `user`
1. Add unique constraint to `username` to `user`
1. Add unique constraint to `username` and `email` to `user`

### scenario 4:

[...]

Issue #41, opened 2016-05-27
Case insensitive search on username and email


cheungpat:
Skygear-server stores username and email with case preserved. It appears that skygear-server perform case sensitive search on username and email for auth and user discovery.

There is an offline discussion about this and we want to support case preserved username and email while having case insensitive search on these two fields.

cheungpat:
@rickmak might want to take a look at [citext module](https://www.postgresql.org/docs/current/static/citext.html) which stores the field with case preserved but compare case insensitively. This also works nicely with index and unique constraint.

If there is a collation for case insensitive comparison, we can use that instead, but [such collation doesn’t seem to exist](http://stackoverflow.com/questions/18807276/how-to-make-my-postgresql-database-use-a-case-insensitive-collation).

PR #411, opened 2017-06-30
Migrate database and update handlers for new user profile design

Steven-Chan:
connect #393

- Update database schema
  - rename `user` to `auth` in all tables
  - create record table `user` with username and email
- Rename `AuthInfo` to `ProviderInfo`
- Rename `UserInfo` to `AuthInfo`
- Remove `Username` and `Email` in `UserInfo`
- Remove `user:query`, `user:update` and `user:link`
- Remove user discovery funcs
- Refactor `TxDatabase` to `Transactional` and `TxDatabase`

All handler request and response signature are unchanged.

============

Update:

- Treat pq `citext` as normal string


Steven-Chan:
I found I couldn't signup with email or username with the new database schema because the column type is incompatible, `citext` vs `text`.

And then I found https://github.com/SkygearIO/skygear-server/issues/41.

Since username and email are now stored in a normal record table, which means these columns are no longer special column to skygear. Then they need to handle schema migration, but `citext` does not exist in skygear record, which leads to the conflict.

Steven-Chan:
I think the concept `case insensitive comparison` is not compatible to the new user profile design, so needs some discussion first.

cheungpat:
Many thanks @Steven-Chan!

pkg/server/preprocessor/preprocessor.go
```diff
@@ -50,6 +50,7 @@ func isTokenStillValid(token router.AccessToken, authInfo skydb.AuthInfo) bool {
 }
 
 func (p InjectUserIfPresent) Preprocess(payload *router.Payload, response *router.Response) int {
+	// TODO: Inject both AuthInfo and user Record
```
cheungpat: Why not do it now?

pkg/server/skydb/authinfo.go
```diff
@@ -43,45 +43,43 @@ type ProviderInfo map[string]map[string]interface{}
...
 	LastSeenAt      *time.Time   `json:"last_seen_at,omitempty"`
 }
 
-// NewAuthInfo returns a new AuthInfo with specified username, email and
-// password. An UUID4 ID will be generated by the system as unique identifier
-func NewAuthInfo(username string, email string, password string) AuthInfo {
-	id := uuid.New()
+// AuthData contains the unique authentication data of a user
+// e.g.: {"username": "userA", "email": "userA@abc.com"}
+type AuthData map[string]interface{}
```
cheungpat: Consider calling `AuthData` as `AuthIdentifiers`.

cheungpat:
Seems lacking test case

pkg/server/handler/common.go
```diff
@@ -38,11 +39,19 @@ type AuthResponse struct {
 	LastSeenAt  *time.Time `json:"last_seen_at,omitempty"`
 }
 
-func NewAuthResponse(info skydb.AuthInfo, accessToken string) AuthResponse {
+func NewAuthResponse(info skydb.AuthInfo, authdata *skydb.AuthData, accessToken string) AuthResponse {
+	username := ""
+	email := ""
+
+	if authdata != nil {
+		username, _ = (*authdata)["username"].(string)
```
Steven-Chan: The structure of `AuthData` would finally be dynamic, the `GetXXX()` methods would not exist.

pkg/server/skydb/authinfo.go
Steven-Chan: I did consider `AuthIdentifiers`, but `AuthIdentifiers` seems to just the long form of `AuthIDs` and there we have `AuthInfoID` in somewhere which is a struct variable storing user id as string.

pkg/server/preprocessor/preprocessor.go
Steven-Chan: The implementation of injecting user record looks pretty complicated. Since the preprocessor of injecting AuthInfo would create the auth record if user has master key, I assume that of injecting user would also do the same.

Either the preprocessor call recordSaveHandler, or it call db save directly, and I think both would not be finished in a few lines. Currently only the MeHandler need this and I think the current PR is big enough, so I want to leave it later.

Steven-Chan:
@cheungpat updated

pkg/server/handler/authutil.go
```diff
@@ -0,0 +1,227 @@
+// Copyright 2015-present Oursky Ltd.
```
cheungpat: TODO(cheungpat): look at this file again

pkg/server/handler/common.go
cheungpat: Sometimes you don’t have the chance to come back and change it later. Plus it you define a function you will be sure there is no where trying to access the value from the `authdata` map if you remove it later.

pkg/server/handler/common.go
```diff
@@ -38,11 +39,19 @@ type AuthResponse struct {
 	LastSeenAt  *time.Time `json:"last_seen_at,omitempty"`
 }
 
-func NewAuthResponse(info skydb.AuthInfo, accessToken string) AuthResponse {
+func NewAuthResponse(info skydb.AuthInfo, authdata *skydb.AuthData, accessToken string) AuthResponse {
```
cheungpat: I still don’t understand why `authdata` needs to be a pointer

pkg/server/handler/me.go
```diff
@@ -26,6 +27,7 @@ type MeHandler struct {
 	Authenticator router.Processor `preprocessor:"authenticator"`
 	DBConn        router.Processor `preprocessor:"dbconn"`
 	InjectUser    router.Processor `preprocessor:"inject_user"`
+	InjectDB      router.Processor `preprocessor:"inject_db"`
```
cheungpat: Don’t use the `InjectDB`. Suppose the client sends `_database_id = "_private"` to the `me` endpoint your code will attempt to get user record from the private DB, which doesn’t exist.

cheungpat: Please also add `RequireUser` preprocessor here by the way.

pkg/server/handler/relation.go
```diff
@@ -351,3 +362,24 @@ func (h *RelationRemoveHandler) Handle(rpayload *router.Payload, response *route
 	}
 	response.Result = results
 }
+
+func fetchUserAuthData(db skydb.Database, conn skydb.Conn, userID string) (data map[string]interface{}, err error) {
```
cheungpat: TODO(cheungpat): not sure what this does

pkg/server/skydb/authinfo.go
cheungpat: You need a better name. `AuthData` and `AuthInfo` looks pretty similar too. People will keep asking this question and I think `AuthData` need a more descriptive name.

pkg/server/handler/authutil.go
```diff
@@ -0,0 +1,227 @@
...
+	txDB, ok := db.(skydb.Transactional)
+	if !ok {
+		return skyerr.NewError(skyerr.NotSupported, "database impl does not support transaction")
+	}
+
+	txErr := withTransaction(txDB, func() error {
+		// Check if AuthData duplicated only when it is provided
+		// AuthData may be absent, e.g. login with provider
+		if len(authData) != 0 {
+			if err := validateAuthData(authData); err != nil {
```
cheungpat: Validating auth data should be the job of the caller IMO

pkg/server/skydb/pq/migration/f0c53134d25d_rename_table_user_to_auth.go
```diff
@@ -0,0 +1,258 @@
...
+	FROM information_schema.tables
+	WHERE
+		table_schema = current_schema() AND
+		table_name = $1
+);
+		`, table).Scan(&exists)
+	return exists, err
+}
+
+func (r *revision_f0c53134d25d) IsColumnExisted(tx *sqlx.Tx, table string, column string) (bool, error) {
```
cheungpat: Existed should not be in past tense. Ditto elsewhere in this file.

pkg/server/handler/authutil.go
```diff
@@ -0,0 +1,227 @@
...
+			}
+
+			if err != skydb.ErrUserNotFound {
+				return skyerr.MakeError(err)
+			}
+		}
+
+		if err := ctx.DBConn.CreateAuth(info); err != nil {
+			if err == skydb.ErrUserDuplicated {
+				return errUserDuplicated
```
cheungpat: Have you considered the case where there is a auth object but the corresponding user record is (for some unknown reason) missing?

pkg/server/preprocessor/preprocessor.go
cheungpat: Agreed

Steven-Chan:
@cheungpat updated

pkg/server/handler/authutil.go
cheungpat: Steven told me the case is considered and they we may revisit this problem in a later PR.

pkg/server/handler/authutil.go
```diff
@@ -0,0 +1,218 @@
...
+			Operator: skydb.ILike,
+			Children: []interface{}{
+				skydb.Expression{Type: skydb.KeyPath, Value: keyPath},
+				skydb.Expression{Type: skydb.Literal, Value: value},
+			},
+		}
+	}
+
+	if username != "" {
+		predicates = append(predicates, makeILikePredicate("username", username))
```
cheungpat: You don’t need ILIKE when the column is citext.

Steven-Chan:
@cheungpat updated

Steven-Chan:
@cheungpat  I just found if a user sign up with username, the email would be empty instead of null, which makes the second empty email sign up request violate the unique constraint. 😢

PR #413, opened 2017-07-04
Move delete old generated mock files after go install

Steven-Chan:
This makes less pain to developer if the source can not build properly and the generated files would not be gone.


oursky-travis:
@Steven-Chan, thanks for your PR! By analyzing the history of the files in this pull request, we identified @cheungpat to be a potential reviewer.

cheungpat:
thanks @Steven-Chan!

PR #414, opened 2017-07-04
Use NewGoroutineAwareTestReporter handler tests

cheungpat:
This is required because the handler is executed in a different goroutine,
which the mock controller will call testing.T. testing.T should
only be called from the same goroutine.


oursky-travis:

rickmak:
Merge via 2b38d15

We may make more test case gorountine aware :)

Issue #418, opened 2017-07-11
Update for czmq v4


cheungpat:
The current zmq library is gocmzq version 1 which is compatible with czmq v3. Since czmq v4 is released we should update the dependency to make installation easier for the advanced users.

- Skygear Server Date/Version: 1.0
- Application Platform:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

cheungpat:
follows #417

PR #419, opened 2017-07-11
Update goczmq to v4

cheungpat:
refs #418


cheungpat:
This is easier than I have thought. Let me also update the other zmq dependencies and use pre-built binaries too. Have wanted to do that for ages.

cheungpat:
Should be okay

PR #420, opened 2017-07-11
New user profile design: auth request with auth data and response with user record

Steven-Chan:
connect #393
need #421

- Update signup and login request to accept `auth_data`
- Update signup, login, add relation and query relation response to return user record
- Treat empty string value in `auth_data` as valid value, so it would not be purged
  - `auth_data` is no longer restricted to string, if empty string is purged, then `0` and `false` would also be purged, which does not make sense
- Add InjectUserIfPresent preprocessor

Note:
`AUTH_RECORD_KEYS` is NOT implemented yet, the current implementation is equal to `AUTH_RECORD_KEYS=username,email`. Most likely, the next, and hopefully the last for the feature, PR would finish that.

### DIFF
--- a/pkg/server/handler/common.go
+++ b/pkg/server/handler/common.go
@@ -17,7 +17,10 @@ package handler
 import (
 	"time"
 
+	"github.com/skygeario/skygear-server/pkg/server/asset"
 	"github.com/skygeario/skygear-server/pkg/server/skydb"
+	"github.com/skygeario/skygear-server/pkg/server/skydb/recordutil"
+	"github.com/skygeario/skygear-server/pkg/server/skydb/skyconv"
 	"github.com/skygeario/skygear-server/pkg/server/uuid"
 )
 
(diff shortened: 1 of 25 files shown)


pkg/server/router/model.go
```diff
@@ -117,6 +117,8 @@ type Payload struct {
 
 	DBConn   skydb.Conn
 	Database skydb.Database
+
+	User *skydb.Record
```
cheungpat: 😱 well okay

pkg/server/skydb/authinfo.go
```diff
@@ -91,7 +92,26 @@ func (a AuthData) IsValid() bool {
...
-	return true
+	return !a.IsEmpty()
+}
+
+// IsEmpty would return true if
+//
+// 1. no entries or
+// 2. all values are null
+func (a AuthData) IsEmpty() bool {
+	emptyCount := 0
```
cheungpat: I would check, if the len(a) == 0 exit first
then exit immediately if you find a non empty field.

pkg/server/handler/common.go
```diff
@@ -28,25 +31,33 @@ var (
...
+	LastLoginAt *time.Time          `json:"last_login_at,omitempty"`
+	LastSeenAt  *time.Time          `json:"last_seen_at,omitempty"`
 }
 
-func NewAuthResponse(info skydb.AuthInfo, authdata skydb.AuthData, accessToken string) AuthResponse {
+type AuthResponseFactory struct {
+	AssetStore asset.Store `inject:"AssetStore"`
+}
+
+func (f AuthResponseFactory) NewAuthResponse(conn skydb.Conn, info skydb.AuthInfo, user skydb.Record, accessToken string) (AuthResponse, error) {
```
Steven-Chan: yup

pkg/server/handler/common.go
```diff
@@ -28,25 +31,33 @@ var (
...
 type AuthResponse struct {
-	UserID      string     `json:"user_id,omitempty"`
-	Username    string     `json:"username,omitempty"`
-	Email       string     `json:"email,omitempty"`
-	Roles       []string   `json:"roles,omitempty"`
-	AccessToken string     `json:"access_token,omitempty"`
-	LastLoginAt *time.Time `json:"last_login_at,omitempty"`
-	LastSeenAt  *time.Time `json:"last_seen_at,omitempty"`
+	UserID      string              `json:"user_id,omitempty"`
+	User        *skyconv.JSONRecord `json:"user,omitempty"`
```
Steven-Chan: `UserRecord`

pkg/server/router/model.go
Steven-Chan: why 😱?

Steven-Chan:
@cheungpat updated

PR #421, opened 2017-07-11
New user profile design: move handler/recordutil to skydb/recordutil

Steven-Chan:


pkg/server/handler/authutil.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/skygeario/skygear-server/pkg/server/asset"
 	"github.com/skygeario/skygear-server/pkg/server/plugin/hook"
 	"github.com/skygeario/skygear-server/pkg/server/skydb"
+	"github.com/skygeario/skygear-server/pkg/server/skydb/recordutil"
```
cheungpat: I think `recordutil` should exists outside of `skydb` or `handler` package. I’d prefer `pkg/server/util/record` or `pkg/server/recordutil`.

pkg/server/skydb/recordutil/recordutil.go
```diff
@@ -328,21 +328,33 @@ func RecordSaveHandler(req *RecordModifyRequest, resp *RecordModifyResponse) sky
...
+	return nil
+}
+
+type saveHookTriggerer struct {
+	Req               *RecordModifyRequest
+	Resp              *RecordModifyResponse
+	OriginalRecordMap map[skydb.RecordID]*skydb.Record
+}
+
+func (t saveHookTriggerer) triggerIfRegistered(records []*skydb.Record, kind hook.Kind, shouldLogErr bool) []*skydb.Record {
```
Steven-Chan: This method may or may not trigger the hook, and I think this should be reflected in the method name. If the logic of checking the registry is put outside this method, I would name it `trigger`. But I think it makes more sense to put the checking in the method to reduce the complexity of the save record handler.

(How do we name things)

pkg/server/skydb/recordutil/recordutil.go
```diff
@@ -328,21 +328,33 @@ func RecordSaveHandler(req *RecordModifyRequest, resp *RecordModifyResponse) sky
...
+	if t.Req.HookRegistry != nil {
+		return executeRecordFunc(records, t.Resp.ErrMap, func(record *skydb.Record) (err skyerr.Error) {
+			originalRecord, _ := t.OriginalRecordMap[record.ID]
+			err = t.Req.HookRegistry.ExecuteHooks(t.Req.Context, kind, record, originalRecord)
+			if shouldLogErr && err != nil {
```
Steven-Chan: I'm just refactoring the code and keeping the behaviour unchanged. If `shouldLogErr` is in the struct, there would be two separate `saveHookTriggerer` in the record save handler. I don't know which one is better.

pkg/server/skydb/recordutil/recordutil.go
```diff
@@ -328,21 +328,33 @@ func RecordSaveHandler(req *RecordModifyRequest, resp *RecordModifyResponse) sky
...
-			if err != nil {
+	records = hookTriggerer.triggerIfRegistered(records, hook.AfterSave, true)
+
+	resp.SavedRecords = records
+	resp.SchemaUpdated = schemaExtended
+
+	return nil
+}
+
+type saveHookTriggerer struct {
```
Steven-Chan: Agree.

pkg/server/skydb/recordutil/recordutil.go
cheungpat: > This method may or may not trigger the hook, and I think this should be reflected in the method name.

No. All code does or does not do something based on condition. The checking being included in the function doesn’t justify the name to include this fact.

pkg/server/skydb/recordutil/recordutil.go
cheungpat: Just put it in the struct and make it consistent. Thanks.

cheungpat: The longer version of my last comment is that I am not aware of any reason there were consistent behavior so let’s just make it consistent this time. Thanks.

pkg/server/skydb/recordutil/recordutil.go
Steven-Chan: > All code does or does not do something based on condition

All code does `something` or does `something else` based on condition, but not all are `does or does not do something`.

You may argue that the `xxxIfxxx` is not consistent in the project or this kind of function is not a good practice, which can be a reason to refactor the function. But IMO if a function has the behaviour that `it does {A} if {B}, or it does nothing if not {B}`, its name should be `{A}if{B}`. Otherwise, it is a trap which makes other developers to think that this function would always do {A}.

cheungpat: 1. If `{A}` and `{B}` does not imply one another, I would agree. However, in this case {A} refers to trigger hooks, and {B} refers to (if) hook is registered. If a hook is not registered, it doesn’t exist, and there is nothing to trigger.
2. The function is named `triggerIfRegistered`, but the code says `if t.Req.HookRegistry != nil`. Doesn’t look like name is a description of the logic. (`triggerIfHookRegistryExists`? No just kidding.) 

So I am not saying that `DoIfSomething` is always a bad practice, no. I am just saying that in this particular case, `triggerIfRegistered` is two words more than necessary. Anyway this is not something I would force on you.

pkg/server/skydb/recordutil/recordutil.go
```diff
@@ -328,21 +328,33 @@ func RecordSaveHandler(req *RecordModifyRequest, resp *RecordModifyResponse) sky
...
+	Req               *RecordModifyRequest
+	Resp              *RecordModifyResponse
+	OriginalRecordMap map[skydb.RecordID]*skydb.Record
+}
+
+func (t saveHookTriggerer) triggerIfRegistered(records []*skydb.Record, kind hook.Kind, shouldLogErr bool) []*skydb.Record {
+	if t.Req.HookRegistry != nil {
```
cheungpat: golang convention is to exit early

```
if t.Req.HookRegistry == nil {
  return records
}

// the rest of the code here, with one less indent
```

Steven-Chan:
@cheungpat updated

cheungpat:
@Steven-Chan Many thanks!

PR #423, opened 2017-07-13
Signup with user profile

Steven-Chan:
connect #393

### DIFF
--- a/pkg/server/handler/auth.go
+++ b/pkg/server/handler/auth.go
@@ -35,6 +35,7 @@ type signupPayload struct {
 	Password         string                 `mapstructure:"password"`
 	Provider         string                 `mapstructure:"provider"`
 	ProviderAuthData map[string]interface{} `mapstructure:"provider_auth_data"`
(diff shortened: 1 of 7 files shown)


pkg/server/handler/auth.go
```diff
@@ -53,6 +54,10 @@ func (payload *signupPayload) Validate() skyerr.Error {
 			return skyerr.NewInvalidArgument("invalid auth data", []string{"auth_data"})
 		}
 
+		if dupliatedKeys := payload.duplicatedKeysInAuthDataAndProfile(); len(dupliatedKeys) > 0 {
```
cheungpat: s/dupliatedKeys/duplicatedKeys/g

cheungpat: Do you think it is better to ignore the keys instead of throwing an error? @rickmak

rickmak: Since the name is signup, I think it will be better throw an error.

PR #426, opened 2017-07-17
Fix the false positive of hmac.equal check

rickmak:
Properly close the base64Encoder

connects #427


oursky-travis:
@rickmak, thanks for your PR! By analyzing the history of the files in this pull request, we identified @limouren to be a potential reviewer.

limouren:
It would be great to have at least one negative test case for `ParseSignature` and normal flow test for `SignedURL`.

PR #428, opened 2017-07-18
Config auth data for signup and login with AUTH_RECORD_KEYS in env

Steven-Chan:
connect #393

### DIFF
--- a/pkg/server/skydb/conn.go
+++ b/pkg/server/skydb/conn.go
@@ -176,11 +176,13 @@ type Conn interface {
 	// RecordEvent from the Conn implementation
 	Subscribe(recordEventChan chan RecordEvent) error
 
-	// EnsureAuthRecordKeysValid check if authRecordKeys exist in database,
-	// has valid type and unique
-	//
-	// Migration would be done in dev mode
-	EnsureAuthRecordKeysValid(authRecordKeys [][]string) error
+	// EnsureAuthRecordKeysExist check if authRecordKeys exist in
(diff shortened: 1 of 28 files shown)


Steven-Chan:
@rickmak @cheungpat updated

I have simplified the logic, down to only check key existence and index existence (literally two separate functions), newly created auth record keys column are string type, but accept any column type for existing column.

Steven-Chan:
Although those two new functions are only called in main.go, but I want to have test case connected to database, so I still put them in pq package.

pkg/server/skydb/conn.go
```diff
@@ -176,11 +176,13 @@ type Conn interface {
...
-	//
-	// Migration would be done in dev mode
-	EnsureAuthRecordKeysValid(authRecordKeys [][]string) error
+	// EnsureAuthRecordKeysExist check if authRecordKeys exist in
+	// user record schema
+	EnsureAuthRecordKeysExist(authRecordKeys [][]string) error
+
+	// EnsureAuthRecordKeysIndexesExist check if indexes of authRecordKeys
+	// exist in user record schema
+	EnsureAuthRecordKeysIndexesExist(authRecordKeys [][]string) error
```
rickmak: `EnsureAuthRecordKeysIndexesExist` -> `EnsureAuthRecordKeysIndexesMatch`

We should mention we will delete indexes here.

Steven-Chan:
@rickmak updated

PR #429, opened 2017-07-18
Include user profile in change password response

Steven-Chan:
connect #393

### DIFF
--- a/pkg/server/handler/auth.go
+++ b/pkg/server/handler/auth.go
@@ -529,7 +529,7 @@ type PasswordHandler struct {
 	DBConn        router.Processor `preprocessor:"dbconn"`
 	InjectAuth    router.Processor `preprocessor:"inject_auth"`
 	InjectUser    router.Processor `preprocessor:"inject_user"`
-	RequireUser   router.Processor `preprocessor:"require_user"`
+	RequireAuth   router.Processor `preprocessor:"require_auth"`
 	PluginReady   router.Processor `preprocessor:"plugin_ready"`
 	preprocessors []router.Processor
 }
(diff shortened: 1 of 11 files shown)


pkg/server/handler/auth.go
```diff
@@ -525,6 +527,7 @@ func (h *PasswordHandler) Setup() {
 		h.Authenticator,
 		h.DBConn,
 		h.InjectAuth,
+		h.InjectUser,
```
cheungpat: try to follow the order defined above (it is a custom)

pkg/server/handler/auth.go
```diff
@@ -580,8 +583,19 @@ func (h *PasswordHandler) Handle(payload *router.Payload, response *router.Respo
 		panic(err)
 	}
 
-	response.Result = AuthResponse{
-		UserID:      info.ID,
-		AccessToken: token.AccessToken,
+	user := payload.User
```
cheungpat: how about using the require user preprocessor? instead of checking if the user exists here? If you keep the tracking here, you need to return a good error with code rather than panic (difference is 4XX and 500)

Steven-Chan: Then I would return a ResourceNotFound error instead of panic, and instead of changing the behaviour of the user preprocessor as you suggested. This has the same behaviour with `InjectAuth`.

Steven-Chan:
@cheungpat updated

Steven-Chan:
@cheungpat updated

Steven-Chan:
@cheungpat updated AuthResponse handling with empty user record, `profile` field would be omitted

pkg/server/handler/auth_test.go
```diff
@@ -1164,5 +1164,28 @@ func TestPasswordHandlerWithProvider(t *testing.T) {
 			So(resp.Code, ShouldEqual, 200)
 		})
 
+		Convey("change password success, without user", func() {
```
cheungpat: It is actually better to return `profile: null` than missing the key.

“without user record”

Steven-Chan:
@cheungpat updated

Issue #430, opened 2017-07-19
New User Profile Design: Add api for get user role, get last login at


Steven-Chan:
Due to https://github.com/SkygearIO/features/issues/48, `User` object and user discovery api in SDKs are removed, so that roles-api, last login at and last seen at are also removed. 

To retain these functionalities, the following apis need to be added:
- `role:get`
  - input: user_id
  - output: roles of the user
  - permission: master_key or can only get the user's own roles
- Move `last_login_at` from `_auth` to `user` record
  - field acl: owner read only / public no access

PR #438, opened 2017-07-26
Mention the return record is copy of the original in RecordResultFilter

rickmak:
refs #430


cheungpat:
Let me make a change and merge that.

PR #441, opened 2017-07-28
Handle unique constraint violation error when signup

Steven-Chan:
connect #393

### DIFF
--- a/pkg/server/skydb/pq/recorddb.go
+++ b/pkg/server/skydb/pq/recorddb.go
@@ -23,10 +23,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/sirupsen/logrus"
 	"github.com/jmoiron/sqlx"
 	sq "github.com/lann/squirrel"
 	"github.com/lib/pq"
+	"github.com/sirupsen/logrus"
 	"github.com/skygeario/skygear-server/pkg/server/skydb"
 	"github.com/skygeario/skygear-server/pkg/server/skydb/pq/builder"
 	"github.com/skygeario/skygear-server/pkg/server/skyerr"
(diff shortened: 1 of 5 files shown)


pkg/server/skydb/pq/recorddb.go
```diff
@@ -150,6 +150,15 @@ func (db *database) Save(record *skydb.Record) error {
 
 	row := db.c.QueryRowWith(upsert)
 	if err = newRecordScanner(record.ID.Type, typemap, row).Scan(record); err != nil {
+		if pqErr, ok := err.(*pq.Error); ok {
```
cheungpat: `if isUniqueViolated {`

cheungpat:
generally okay but there are test errors that need fixing

Steven-Chan:
skydb/skydbtest/skydbmockutil.go
```
func ExpectDBSaveUser(db *mock_skydb.MockTxDatabase, extendedSchema *skydb.RecordSchema, assertSavedUserRecord interface{}, saveError error) {
	db.EXPECT().ID().Return("_public").AnyTimes()

	// no record found
	db.EXPECT().
		Get(gomock.Any(), gomock.Any()).
		Return(skydb.ErrRecordNotFound).
		AnyTimes()

        // db.Get(skydb.NewRecordID("user", "abc"), nil) === skydb.ErrRecordNotFound
        // it works pretty well

	// extend Schema
	if extendedSchema != nil {
		ExpectDBExtendSchema(db, *extendedSchema)
	}

	db.EXPECT().
		Save(gomock.Any()).
		Do(assertSavedUserRecord).
		Return(saveError).
		AnyTimes()

         // db.Save(nil) == nil
         // saveError != nil
         // WHAT THE HELL IS GOING ON
}
```

@cheungpat 
Because of this ridiculous behaviour of gomock, I am going to delete those test case for auth data duplication.

Steven-Chan:
@cheungpat updated

Issue #445, opened 2017-07-31
Remove admin account in server seed data


Steven-Chan:
Simplify database migration in server startup.

PR #446, opened 2017-07-31
Remove add admin account and admin role logic in migration seed data

Steven-Chan:
connect #445


rickmak:
:just asking: should we also modify `bce7089fca4`?

Steven-Chan:
I think we should avoid changing migration revision, unless they have critical bug.

PR #451, opened 2017-08-10
Use only annotated tags for version string

cheungpat:
This is because `latest` is also used as tag for release purposes and
`--tags` option makes `latest` appears in version string.


cheungpat:
Closed without prejudice.

PR #453, opened 2017-08-10
Update cmd for changing releasing version of SDKs

Steven-Chan:


cheungpat:
Can you also add `make update-version` to skygear-server’s Makefile, too?
And change all Makefile to remove `--tags` from git-describe command.

Issue #461, opened 2017-09-13
[ACL] Update user email username default ACL


carmenlau:
After discussion, decided to change it to read only and queryable for any user

rickmak:
Duplicated with https://github.com/SkygearIO/skygear-server/issues/457

Issue #457, opened 2017-08-16
Remove default field ACL of username and email of user record


Steven-Chan:
https://github.com/SkygearIO/skygear-server/issues/431

PR #462, opened 2017-09-13
Update user username and email field acl default for new app

Steven-Chan:
connect #461 
connect #457


rickmak:
Merge via 26fa86a

Issue #473, opened 2017-10-08
Email and Username of users, should have default as "Discoverable"


chpapa:
To avoid new users confusing, we have decided to make Email and Username of users by default discoverable (Public: Read Only, Queryable)

PR #474, opened 2017-10-10
Update user username and email field acl default for new app

Steven-Chan:
Update to read only and queryable for any user.

Updating commit 1471ffb61feff57250ba13f92a5dee4b55d4519a.

connect #473


rickmak:
Merge via dde0d70

PR #478, opened 2017-10-12
New auth handlers to support SSO features

carmenlau:
connects #459

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Changes since 1.0.0
+## 1.1.0 (2017-08-07)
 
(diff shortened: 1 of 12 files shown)


carmenlau:
Currently the auth data structure is 
```
{
    "facebook:<fb_id>": {
        "access_token": "....",
        .....
    }
}
```

After discussion, it is make more sense to put the id inside the dict. And we will also put the provider profile to the auth data.

```
{
    "facebook": {
        "principal_id": "<fb_id>",
        "access_token": "....",
        .....,
        "profile": <fb profile payload>
    }
}
```

Will reopen PR after updating

Issue #479, opened 2017-10-12
Field-based ACL behaviour


Steven-Chan:
Discussed with @rickmak and @ben181231 offline, there seems to be some mis-alignment on the behaviour of how the field-based ACL works. The discussion is based on the following scenario:

This is how the current v1.1.0 skygear-server works.

I am NOT logged in to the server, and I want to query user by user.username.

Initially, the field ACL setting of ALL record types is:

record|field|user role|access|discovery
---|---|---|---|---
user|username|owner|read write|queryable
user|username|public|no access|not queryable

Then when I query user by user.username, I would get error.

Now I delete the public one, the field ACL setting is:

record|field|user role|access|discovery
---|---|---|---|---
user|username|owner|read write|queryable

**So if I query user by user.username, will I still get the error?**

There are two side of opinions.

## 1

There are three levels of field ACL the server need to match:

1. `user.username`
1. `user.*`
1. `*.*`

Since an entry is found for the first condition, the server will match by that entry. The entries are worked by whitelist. So only owner can query by user.username, others will get error.

Result: Error, so having the `Public` entry or not DOES NOT affect the result

## 2

On top of the above three conditions, the server will also match the user role. So effectively the server will try to find entries that match `user.username` **AND** user role that can match not authenticated user.

Then no entries are matched. The default behaviour will be applied.

Result: Query successfully, so the `Public` entry DOES affect the result

---

The actual result is **2**.

#### Why there is such a discussion

There is a `Private` option in Skygear portal  schema table acl setting modal dialog. There are two entries, which are exactly the first table in the above scenario. Then after some offline discussion, we thought there should only be one entry, the `Public` one should be useless. 

But this is not correct in the current implementation.

cheungpat:
The [current design](https://github.com/SkygearIO/features/blob/master/features/48-user-profile-discovery/field-based-acl.md) and [implementation](https://github.com/SkygearIO/skygear-server/blob/master/pkg/server/skydb/access.go) is based on a rule-baesd matching algorithm and the default for unmatched rule is least-restrictive (i.e. public readwrite, queryable, aka *full* access).

The default is least-restrictive because the field-based ACL work hand-in-hand with record ACL, and having a default for least-restrictive is consistent with existing behaviour (i.e. we don’t want to break existing app nor do we want developers to add *public full access* in order to revert to previous behaviour).

[...]

PR #481, opened 2017-10-17
Update FieldACL Accessible matching logic

Steven-Chan:
New matching criteria is, record type and record field.
- If no matched entries, return true
- Otherwise if user role does not fullfil matched entries, return false
- Otherwise return the result of the matched entry

connect #479

### DIFF
--- a/pkg/server/skydb/access.go
+++ b/pkg/server/skydb/access.go
@@ -200,22 +200,23 @@ func (acl FieldACL) Accessible(
 	// the specified record type and field. The iterator will handle the
 	// scenario where wildcard record type and field.
 	iter := NewFieldACLIterator(acl, recordType, field)
-	for {
-		entry := iter.Next()
-		if entry == nil {
-			// There is no matching ACL entry, the fallback is to grant access.
-			return true
-		}
+	entry := iter.Next()
 
-		// Only the rules that matches the user role will be selected.
-		if !entry.UserRole.Match(authInfo, record) {
-			continue
-		}
-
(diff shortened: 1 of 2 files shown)


pkg/server/skydb/access.go
```diff
@@ -230,51 +231,58 @@ type FieldACLIterator struct {
...
 
-	nextRecordTypes []string
-	nextEntries     []FieldACLEntry
-	eof             bool
+	nextEntries []FieldACLEntry
 }
 
 // NewFieldACLIterator creates a new iterator.
 func NewFieldACLIterator(acl FieldACL, recordType, recordField string) *FieldACLIterator {
+	// Categorize three types of entry list
```
cheungpat: I understand that you want to categorise all the ACE in one of the three categories, and then return the category based on `recordType` and `recordField`.

Since the FieldACL is sorted already, the iterator doesn’t need to process all the ACEs first in order to find the required matches. If the iterator found a matched ACE, it can keep returning the matches ACE until the next ACE doesn’t match again.

i.e.

1. for each ACE, check if record type and record field matches the record
2. if matches, return the ACE, and remember the record type and record field pair
3. keep returning the ACE that has the same record type and record field pair
4. when the next ACE doesn't have the same record type and record field pair, return nil and stop iterating

would be great if you could change it back

Steven-Chan:
@cheungpat updated

Steven-Chan:
@cheungpat removed the defer 🤦‍♂️

cheungpat:
picked as 20eb6d6

PR #499, opened 2017-11-30
Rename docker-compose.{dev,make}.yml

cheungpat:
This avoid the confusion that the docker compose file could be used for
development.


carmenlau:
Merged via d8a41d5

PR #504, opened 2017-12-05
User audit

louischan-oursky:


rickmak:
I don't comment line by line, but follow is the area I want to pay attendtion to.

- in `authInfo.PasswordHistoryEnable` is something should not belong to `authInfo`.
- The param at `GetPasswordHistory ( histoySize history Days?)` is a bit ugly. i.e. passing non-changing variable.

I think the root cause is you try to not modify the `coon`, but in fact you can. Let put the password related setting into the `conn` and make application code looks better.

For how to put config into conn, you can refer to the previous commit on how we add `canMigrate`
refs: https://github.com/SkygearIO/skygear-server/commit/5e645e650

I proposing add in `config` (you can think of other name is good :))

<hr>
- autinfo IsPasswordExpired interface is design for easy test? That why you want to inject the time.Time, which in real application code will only be `time.now().UTC()`. I think we can use mock instead of inject in testing so the application code will look better.

<hr>

- When rename to PasswordChecker files should rename the file too.

But I think in this project, the convention is to put a minimal struct and related logical function into one file. 

<hr>
Only last things, `Trail`. I am not sure why we don't use log with tag filed but making yet another stuct for audit. What is the benefit here.

<hr>

Logistics wise, I think you will rebase to master and push to `user-audit` branch directly and open separate issue/PR for the above improvement.

rickmak:
Pushed to `user-audit` branch via a63811f

Issue #494, opened 2017-11-16
Login with custom token


carmenlau:
(Describe the issue here)

- Skygear Server Date/Version:
- Application Platform:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Steps to reproduce

1.
2.
3.

PR #506, opened 2017-12-12
Add new sso:custom_token:login handler

cheungpat:
No test case for the handler at the moment

connects #494

### DIFF
--- a/pkg/server/handler/sso_customtoken.go
+++ b/pkg/server/handler/sso_customtoken.go
@@ -91,15 +91,40 @@ func (payload *ssoCustomTokenLoginPayload) Validate() skyerr.Error {
 }
 
 /*
-SSOCustomTokenLoginHandler authenticate user with password
+SSOCustomTokenLoginHandler authenticates the user with a custom token
 
-The user can be either identified by username or password.
+An external server is responsible for generating the custom token which
+contains a Principal ID and a signature. It is required that the token
+has issued-at and expired-at claims.
+
+The custom token is signed by a shared secret and encoded in JWT format.
+
+The claims of the custom token is as follows:
+
+    {
+      "sub": "id1234567800",
+      "iat": 1513316033,
+      "exp": 1828676033,
+      "skyprofile": {
+        "name": "John Doe"
+      }
+    }
+
+When signing the above claims with the custom token secret `ssosecret` using
+HS256 as algorithm, the following JWT token is produced:
+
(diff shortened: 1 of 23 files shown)


pkg/server/skydb/conn.go
```diff
@@ -227,6 +226,14 @@ type Conn interface {
 	DeleteOAuth(provider string, principalID string) error
 
 	Close() error
+
+	CustomTokenConn
```
rickmak: I think we need to use this pattern to other API groupings.

pkg/server/handler/sso_customtoken.go
```diff
@@ -0,0 +1,294 @@
...
+/*
+SSOCustomTokenLoginHandler authenticate user with password
+
+The user can be either identified by username or password.
+
+curl -X POST -H "Content-Type: application/json" \
+  -d @- http://localhost:3000/ <<EOF
+{
+	"action": "sso:custom_token:login",
+	"token": "eyXXXXXXX",
```
rickmak: I think the purpose of the example is to provide a minimal working example with the default setting.

I suggest the following:

1. Add the setting with default to `.env.sample`
2. the token here is actually valid with default setting.

pkg/server/handler/sso_customtoken.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func (h *SSOCustomTokenLoginHandler) Handle(payload *router.Payload, response *router.Response) {
+	if h.CustomTokenSecret == "" {
+		response.Err = skyerr.NewError(
+			skyerr.NotConfigured,
+			"login with custom token requires CUSTOM_TOKEN_SECRET config",
+		)
+		return
+	}
```
rickmak: Logically I think this checking should happen at preprocessorr...?

pkg/server/handler/sso_customtoken.go
cheungpat: even if the token is correctly signed the request still wouldn’t be valid because the validity period will cause the token to become invalid

but I totally recognise we need a better example, so i will also include info on how this token is generated

pkg/server/handler/sso_customtoken.go
cheungpat: I am not sure about that... doing so we need to create a preprocessor to check if the custom token secret is configured, meaning that the preprocess have only one purpose and it is always used by this handler.

Okay?

pkg/server/handler/sso_customtoken.go
```diff
@@ -0,0 +1,319 @@
...
+	} else {
+		userRecordContext.BeforeSaveFunc = func(conn skydb.Conn, info *skydb.AuthInfo) error {
+			if err := conn.UpdateAuth(info); err != nil {
+				return skyerr.MakeError(err)
+			}
+			return nil
+		}
+		userRecordContext.BeforeCommitFunc = func(conn skydb.Conn, user *skydb.Record) error {
+			now := timeNow()
+			if err := conn.CreateCustomTokenInfo(&skydb.CustomTokenInfo{
```
carmenlau: Need further check if we need createCustomTokenInfo in this case, the other part looks great!:)

cheungpat:
thanks @carmenlau for catching the errors, i have updated the pull request.

PR #510, opened 2017-12-15
Fix timeNow does not return the current time

rickmak:


cheungpat:
cherry-picked as d522d62a

PR #515, opened 2018-01-03
Fix test failure after refactoring

louischan-oursky:
I wonder if I can avoid the getter and setter. I took a look at `reflect` but it does not allow me to set a package var reflectively.

### DIFF
--- a/pkg/server/skydb/skydb.go
+++ b/pkg/server/skydb/skydb.go
@@ -21,3 +21,14 @@ import (
 
 var log = logging.LoggerEntry("skydb")
 var timeNow = func() time.Time { return time.Now().UTC() }
+
+// SetTimeNowForTestingOnlyPleaseDoNotUseThis sets timeNow
+// for testing purpose only.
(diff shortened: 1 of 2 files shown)


pkg/server/skydb/skydb.go
```diff
@@ -21,3 +21,14 @@ import (
...
+// SetTimeNowForTestingOnlyPleaseDoNotUseThis sets timeNow
+// for testing purpose only.
+func SetTimeNowForTestingOnlyPleaseDoNotUseThis(f func() time.Time) {
+	timeNow = f
+}
+
+// GetTimeNowForTestingOnlyPleaseDoNotUseThis returns timeNow
+func GetTimeNowForTestingOnlyPleaseDoNotUseThis() func() time.Time {
+	return timeNow
+}
```
rickmak: We should put this method into `skydb_test.go`

rickmak: I found it is impossible. I submit a new PR on reducing the number of public func: https://github.com/SkygearIO/skygear-server/pull/518

PR #542, opened 2018-02-09
Add reset password handler

louischan-oursky:


pkg/server/handler/auth.go
```diff
@@ -712,3 +712,162 @@ func (h *PasswordHandler) Handle(payload *router.Payload, response *router.Respo
...
+	payload.authInfo = &skydb.AuthInfo{}
+	payload.userRecord = &skydb.Record{}
+	if err := payload.conn.GetAuth(payload.AuthID, payload.authInfo); err != nil {
+		return skyerr.NewError(skyerr.UnexpectedAuthInfoNotFound, err.Error())
+	}
+	if err := payload.db.Get(skydb.NewRecordID("user", payload.AuthID), payload.userRecord); err != nil {
+		return skyerr.NewError(skyerr.UnexpectedUserNotFound, err.Error())
+	}
+	return nil
+}
```
rickmak: Let keep the DB related check in the `Handle` func.

The responsibility of a payload should only include the the incoming format and existence. For DB related check, we are putting them in the Handler.handle func.

pkg/server/handler/auth.go
```diff
@@ -712,3 +712,162 @@ func (h *PasswordHandler) Handle(payload *router.Payload, response *router.Respo
...
+	var userData map[string]interface{}
+	if payload.userRecord != nil {
+		userData = map[string]interface{}(payload.userRecord.Data)
+	}
+	return payload.passwordChecker.ValidatePassword(audit.ValidatePasswordPayload{
+		AuthID:        payload.AuthID,
+		PlainPassword: payload.NewPassword,
+		UserData:      userData,
+		Conn:          payload.conn,
+	})
```
rickmak: I think for the same reason, the passwordChecker call should also in `Handle` func.

Let change the changePasswordHandler too.

pkg/server/handler/auth.go
```diff
@@ -712,3 +712,162 @@ func (h *PasswordHandler) Handle(payload *router.Payload, response *router.Respo
 	}.WithRouterPayload(payload))
 	h.PwHousekeeper.Housekeep(info.ID)
 }
+
+type resetPasswordPayload struct {
+	AuthID          string `mapstructure:"auth_id"`
+	NewPassword     string `mapstructure:"password"`
+	Invalidate      bool   `mapstructure:"invalidate"`
```
rickmak: Remove the `Invalidate` flag in the resetPasswordPayload.

PR #543, opened 2018-02-09
Add Baidu Push Notification Service

limouren:
Service URL: http://push.baidu.com/

Only android is supported. For iOS push notification support, developers could simply refer to APNS.

:D


pkg/server/push/baidu/baidu.go
```diff
@@ -0,0 +1,315 @@
+// Package baidu provides Baidu Push client implementation, and a push.Pusher
```
cheungpat: need copyright boilerplate

chpapa:
@limouren @cheungpat @carmenlau Sorry for my ignorance; May I ask what is the implication of this pull request? Does it meant the existing Android SDK can send push notification via Baidu push if the env var are configurated properly? Or it will requires extra changes and support on the Android Client SDK?

PR #564, opened 2018-03-18
Add verify code table

cheungpat:
connects SkygearIO/forgot_password#31
connects SkygearIO/forgot_password#32


carmenlau:
Merged via fe74198

Issue #553, opened 2018-03-07
Async hooks are not called in HTTP transport


yipsang:
- Skygear Server Date/Version: 1.3.3

#### Steps to reproduce

1. Signup a user with sso (user after_save hook is called)
2. Update user record (user after_save hook is not called)

The hook is defined in this way:
```
@skygear.after_save('user', async=True)
@with_session
def after_save_user(session, record, original_record, db):
    _send_verification_email_if_needed(session, record)
    return record
```

carmenlau:
@yipsang Tried locally and cannot reproduce the problem. In my testing I removed your `with_session ` decoration. Would like to see

1. Can you try remove `with_session` and see if it will be called?
2. Do you have this problem in your dev environment? If so, in the skygear server log can you see user's `afterSave` registered?

thanks >"<

carmenlau:
Verified it happens in http transport

limouren:
zeromq is da best!

cheungpat:
@limouren the same happens in ZMQ transport

cheungpat:
@yipsang It would be great if you could try it out : ) you can fetch this image by pulling `quay.io/skygeario/skygear-server:canary`.

yipsang:
@cheungpat cool! will try later

PR #565, opened 2018-03-20
Fix async hook not executed

cheungpat:
This happens because the hook execution is bound to the client request
context, which could finish sooner than the execution of an async
hook. This is fixed by copying the request context into a new background
context, which has a separate timeout.

connects #553


carmenlau:
Merged via 349582d

PR #566, opened 2018-03-21
Update logic for user checking in preprocessors

cheungpat:
If the user is not allowed to access resource because of policies on user
data, such as password expiry, token expiry, disabled or not verified,
The preprocessor will still allow access if user is not required. Preprocessor
will not set AuthInfo in this case.

connects #554


cheungpat:
@carmenlau fixed the lint and added some test cases

cheungpat:
@carmenlau need your help

cheungpat:
@carmenlau updated

Issue #536, opened 2018-01-31
APNS token should be removed from the db when apnsErrorStatus=410


cheungpat:
This error can occur in skygear server when the APNS return a error status 410 (The device token is no longer active for the topic.) https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html

Since the token no longer works, sending notification to the token will result in more errors. When encountering the error, the token should be removed from the device table.

- Skygear Server Date/Version: 0.240

cheungpat:
adding this to the next bug fix milestone

PR #571, opened 2018-04-09
Update log level on apns token invalid error

cheungpat:
If the error occurred for sending push notification is `Gone` or
`BadDeviceToken`, the log level will be Info instead of Error.

connects #536


carmenlau:
Merged via 61ac266

Issue #576, opened 2018-04-11
[Structured logging] Log formatting to support JSON and plain text log for different deployment


carmenlau:
- Log function to print log in JSON + plain text format by config
- Generate request id and pass to plugin context

PR #579, opened 2018-04-17
Improve request logging and format

cheungpat:
This changeset will add support for JSON log formatter. The request log
will become structural and a request ID is added to log and response header.

connects #576


cheungpat:
@carmenlau I wonder if this is going to the desired direction for making structural logging? Let me know if there are problems : )

cheungpat:
I am going to close this pull request because I need to rebase #590 and it has everything included in this pull request.

PR #581, opened 2018-04-20
Refactor how processors are injected to plugin handlers/lambdas

cheungpat:
This is required because we need to inject asset store to lambda in
order to support asset signing when we support returning assets.


carmenlau:
Merged via 7173ebe

Issue #577, opened 2018-04-11
[Structured logging] Review all skygear-server logs to use new format log functions


carmenlau:
We did the log function in issue https://github.com/SkygearIO/skygear-server/issues/576. In this issue, we will review all existing log and update

PR #590, opened 2018-05-16
Better log

cheungpat:
connects #577
connects #576


cheungpat:
@carmenlau the problem with “request log not printed” is fixed.

PR #591, opened 2018-05-17
Update README.md

tensiuyan:
Add how to get support


chpapa:
@tensiuyan There is already a support section in the original README.md, you should either edit that section, or if you think it needs to be more prominent, move up. Don't have duplicated section in the same document.

tensiuyan:
Sorry overlooked.

PR #595, opened 2018-05-23
WIP: Rename hook `async` to `async_mode`

cheungpat:
refs SkygearIO/py-skygear#196


cheungpat:
@carmenlau Do you think this will fix the issue? The plugin side will look like this:

```python
@before_save("note", async_mode=True)
```

Will update py and js and doc if you think this is okay.

cheungpat:
Will be superseded with another pull request on py-skygear.

PR #613, opened 2018-07-11
Update record:delete action for record ID

cheungpat:
connects #605


cheungpat:
@ben181231 @carmenlau updated the PR to bring back multiple record type deletion

PR #621, opened 2018-08-17
Fix `make vendor` error on case insensitive file system

ben181231:
During installation of dependencies on case insensitive file system, the following command would be executed one after another, which cause an error:

1. `mkdir $GOPATH/src/github.com/skygeario`
1. `mkdir $GOPATH/src/github.com/SkygearIO`


carmenlau:
Merged via db95e896

carmenlau:
Merged to v1-branch via 8d80ba1

PR #639, opened 2018-08-29
(Follow up of) Specification on dependency should be on auth not core

Steven-Chan:
connect #630 

follow up of #638


rickmak:
After second though, I think the `Inject` should not at server. It should at dependency or handler factory.

PR #642, opened 2018-08-31
Move dependency logic to handler factory

rickmak:
Server should not aware of the dependency, while handler factory should.
Here we try to move the mount path information to handler instead of
main.go. It is because the where is mount related more to the business
logic of a handler than how the server boot.

connect #630

### DIFF
--- a/cmd/auth/main.go
+++ b/cmd/auth/main.go
@@ -15,12 +15,13 @@ import (
 )
 
 func main() {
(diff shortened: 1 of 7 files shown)


Steven-Chan:
I think `AuthProviders` or `provider. AuthProviders` by itself is not a very good name.

Anyway, I got ur point.

rickmak:
`AuthInjector`?
`provider.Providers`?

PR #644, opened 2018-09-03
Provide TokenStore and AuthInfoStore in handler

Steven-Chan:
Use jwt token store as TokenStore and DBConn pq implementation as AuthInfoStore.

TODO: 
- Break down DBConn into multiple interfaces, and only include AuthInfo related API in core package

connect #631


pkg/core/server/server.go
```diff
@@ -29,24 +32,40 @@ func NewServer(addr string) Server {
...
-			DBConnectionStr: "public",
+		var configuration config.TenantConfiguration
+		if s.devMode {
+			configuration = envConfiguration
+		} else {
+			configuration = config.GetTenantConfig(r)
 		}
 
-		h := hf.NewHandler(configuration)
+		h := hf.NewHandler(r.Context(), configuration)
```
rickmak: If we use r.Context here, we should just remove `handler.Conext` as we discussed before?

It makes the Handler confront to stand interface.

Steven-Chan:
@rickmak removed the middleware

Issue #629, opened 2018-08-28
AuthRouter


rickmak:
Composing

*TenantConfiguration
* UserAuthentication
* UserAuthorization

PR #648, opened 2018-09-05
next: authn and authz in server

Steven-Chan:
connect #629 
after #644

TODO:
- ~~handle master key like the skygear-server~~


pkg/core/server/server.go
```diff
@@ -55,9 +64,22 @@ func (s *Server) Handle(path string, hf handler.Factory) *mux.Route {
...
 
-		h.Handle(handler.Context{
-			ResponseWriter: rw,
-			Request:        r,
-		})
+		var authInfo auth.AuthInfo
+		if s.authInfoResolverFactory != nil {
+			resolver := s.authInfoResolverFactory.NewResolver(r.Context(), configuration)
+			authInfo, _ = resolver.Resolve(r)
+		}
```
rickmak: Assume `s.authInfoResolverFactory` is required.

Make the default factory in `NewServer`.

pkg/core/auth/authz/policy.go
```diff
@@ -0,0 +1,145 @@
...
+	"time"
+
+	"github.com/skygeario/skygear-server/pkg/core/auth"
+	"github.com/skygeario/skygear-server/pkg/server/skyerr"
+
+	"github.com/skygeario/skygear-server/pkg/gateway/model"
+)
+
+type PolicyProvider interface {
+	ProvideAuthzPolicy(*http.Request) Policy
```
rickmak: `ProvideAuthzPolicy` don't need to depends on the Request.

The policy should be same for all kind of request in a specific handler?

Steven-Chan: This interface would be useful for restful api, and I imagine chat may be one of them.

Steven-Chan: I think I will remove passing the request as param, and maybe we have a clearer picture on how to deal with it when doing chat.

pkg/core/server/server.go
Steven-Chan: But the default requires `authDependency` to work, which the server itself has no knowledge of it.

Steven-Chan:
@rickmak updated

Issue #632, opened 2018-08-28
DBProvider


rickmak:
Providing App connection to gears

PR #649, opened 2018-09-05
[WIP]Add DBProvider to auth gear

rickmak:
Call for review on the separation. I am making the connection pooling at core.
Gear will only get the sql.Conn.

connects #632


pkg/core/db/db.go
```diff
@@ -2,24 +2,33 @@ package db
...
 }
 
-func (db DB) GetRecord(recordID string) string {
-	return db.ConnectionStr + ":" + recordID
+func (p DBProvider) Provide(tConfig config.TenantConfiguration) *sql.Conn {
+	db, err := sql.Open("postgres", tConfig.DBConnectionStr)
+	if err != nil {
+		panic(err)
+	}
+	conn, err := db.Conn(context.Background())
```
Steven-Chan: I suggest passing the context as arg of this function. Thus when a handler need to provide a db connection, it can use the request context, and the db connection would be bound to the request life cycle, i.e. when a http request is finished, the db connection would be cleaned up by the underlying db library.

cmd/auth/main.go
```diff
@@ -32,6 +32,7 @@ func main() {
 	envconfig.Process("", &configuration)
 
 	authDependency := provider.AuthProviders{
+		DB:            db.NewDBProvider("auth"),
```
Steven-Chan: duplicated `DB`

pkg/auth/provider/provider.go
```diff
@@ -3,20 +3,22 @@ package provider
...
-	DB            *db.DBProvider
 	TokenStore    *auth.TokenStoreProvider
 	AuthInfoStore *auth.AuthInfoStoreProvider
+	DB            *coreDB.DBProvider
 }
 
 func (d AuthProviders) Provide(dependencyName string, ctx context.Context, tConfig config.TenantConfiguration) interface{} {
 	switch dependencyName {
 	case "DB":
+		return d.ProvideDB(ctx, tConfig)
```
Steven-Chan: duplicated `return`

Steven-Chan:
Should the core library provide something like the `conn` struct in https://github.com/SkygearIO/skygear-server/blob/master/pkg/server/skydb/pq/conn.go#L79, instead of a raw sql.Conn?

The `conn` is already handling the app name name spacing for skygear-server, I guess the gear name spacing is somehow similar.

pkg/core/db/db.go
rickmak: agree

rickmak:
Updated.

I am thinking it should be in core or in auth. Tentatively, I am toward sqlx.DB at core and sqlx.Conn at auth.

PR #651, opened 2018-09-06
next: Read api key and config in middleware with dev mode

Steven-Chan:
- moved some functions from gateway to core
- fixed some constant key string

### DIFF
--- a/cmd/auth/main.go
+++ b/cmd/auth/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/skygeario/skygear-server/pkg/auth/provider"
 	"github.com/skygeario/skygear-server/pkg/core/auth"
 	"github.com/skygeario/skygear-server/pkg/core/db"
+	"github.com/skygeario/skygear-server/pkg/core/middleware"
 	"github.com/skygeario/skygear-server/pkg/core/server"
 )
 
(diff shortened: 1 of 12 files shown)


cmd/auth/main.go
```diff
@@ -37,7 +38,11 @@ func main() {
 		AuthInfoStore: &auth.AuthInfoStoreProvider{CanMigrate: true},
 	}
 
-	srv := server.NewServer("localhost:3000", configuration.DevMode)
+	srv := server.NewServer("localhost:3000")
+
+	if configuration.DevMode {
+		srv.Use(middleware.EnvTenantMiddleware{}.Handle)
+	}
```
rickmak: My preference is still use Hanlder(https://golang.org/pkg/net/http/#Handler) as middleware, not mux.MiddlewareFunc.

Steven-Chan: Do you mean like `StripPrefix` in https://golang.org/pkg/net/http/#StripPrefix?

But `net/http` itself is not consistent, see `TimeoutHandler` and `StripPrefix`.

Steven-Chan: Or do you mean like `CORSMiddleware` in skygear-server?

rickmak: I mean `CORSMiddleware` in skygear-server.

Steven-Chan: I prefer `mux.MiddlewareFunc`. 

Middleware and handler are practically not exchangeable, because take skygear-server middlewares as example, they all require the `Next` to be not nil. So if someone create a `CORSMiddleware` but does not pass a `Next` handler to it, it may create runtime error. Making it declarative (i.e. use something like a `mux.MiddlewareFunc` interface) can prevent this, also the code look cleaner.

node express and elixir phoenix also take a declarative approach on the middleware.

Steven-Chan:
@rickmak i've updated to use one TenantConfigurationMiddleware with different ConfigurationProvider.

Steven-Chan:
@rickmak rebased

Issue #624, opened 2018-08-28
Tenant configuration DB


rickmak:
* Schema
    * Migration
    * Accessor, connection pool

carmenlau:
Initial thought of the configuration db schema
![configdb 2](https://user-images.githubusercontent.com/388892/44713039-d0f85080-aae4-11e8-878b-8d9096857e6d.png)


* Determine the app by domain 
* config JSON
```
{
    "DATABASE_URL": "postgresql://...",
    "API_KEY": "",
    "MASTER_KEY": "",
    "TOKEN_STORE": "jwt",
    "TOKEN_STORE_SECRET": ""
}
```
* gears JSON - Enabled gears for given app plans. JSON example:
```
{
    "auth": true,
    "cms": false
}
```

PR #652, opened 2018-09-06
next: Add migration to gateway with Makefile

rickmak:
connect #624

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,33 @@ Next is the V2 of Skygear that aim to follow
         └── main.go
 ```
 
+## Gateway
+
+### Migration
+
+To provide db url by `MIGRATION_DB=`, default is `postgres://postgres:@localhost/postgres?sslmode=disable` for developement
+
+**Add a version**
+
+```
+cd ./scripts/gateway
+$ make add-version REVISION=<revision_description>
+```
+
+**Run migration**
+
+```
+cd ./scripts/gateway
+$ make migrate-up
+```
+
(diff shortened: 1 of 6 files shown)


Steven-Chan:
I think `WITH_DOCKER` is better than `NO_DOCKER` https://github.com/SkygearIO/skygear-server/blob/master/Makefile#L28, but as long as the option is there, i am fine.

rickmak:
WITH_DOCKER=0 or NO_DOCKER=1 that is the question. Wording itself is telling with is default IMO.

Steven-Chan:
The difference is `WITH_DOCKER` in the skygear-server example does not alter the cmd, but prepend docker cmd.

rickmak:
Seem it is not possible with the migrate/migrate docker image? I did not check details. @carmenlau ?

carmenlau:
`migrate/migrate` docker image itself is an executable with entry point calling the binary, so the command will be a bit different between with or without docker. So cannot simply prepend docker cmd, the binary path are different.

I think using docker should be the default? As I think in most of the case we don't want install extra command? :P 

How about change to `WITH_DOCKER` and `WITH_DOCKER ?= 1`? If it looks good, let me change it :)

Steven-Chan:
@carmenlau Then i'm ok with `NO_DOCKER`.

carmenlau:
Cool, let use Rick's commit. I picked rick's commit to my PR and did some small update for the schema in the PR. Let's continue there.

PR #659, opened 2018-09-12
next: Gear db migration

Steven-Chan:
original tables:

- `_auth`
- `_password_history`
- `_sso_custom_token`
- `_sso_oauth`

proposed tables:

- core gear:
  - `_sky_user`
    - contains auth info
    - `_auth` without `password` and `provider_info`
- auth gear:
  - `_auth_password_history`
    - replace `_password_history`
  - `_auth_provider_oauth`
    - replace `_sso_oauth`
  - `_auth_provider_password`
    - contains the auth data, which is needed when client call `loginWithAuthData`
  - `_auth_principal`
    - contains all principal, each principal represents a method to login
    - each user has many principals
    - custom auth provider will have a custom `provide` value here

Note:
- core tables start with `_sky_`
- gear tables start with `_{gear_name}_`, e.g. `_auth_xxx` and `_record_xxx`
- so we cannot have a gear named `sky` 🙊


README.md
```diff
@@ -57,6 +57,37 @@ cd ./scripts/gateway
...
 ```
 
+## Gear DB migration
+
+**Add a version**
+
+```
+$ export GEAR=<gear_name>
+$ export REVISION=<revision_description>
+$ make -f scripts/gateway/Makefile add-version MIGRATION_DIR=cmd/migrate/gear/${GEAR} REVISION=${REVISION}
```
carmenlau: It would be better if we have a `common.mk` in core, and gear has its own Makefile?

Steven-Chan:
I prefer having another program or another set of commands to manage the db status, because there must be some problem if the required condition is not reached, which maybe schema not created or core table version not updated, and these are some higher level control.

Maybe something like this:

```
skydbctl version -dburl xxx -app xxx [-gear xxx]
skydbctl create -dburl xxx [-app xxx [-gear xxx]]
skydbctl drop -dburl xxx [-app xxx [-gear xxx]]

// in non-force update, skydbctl would resolve and update the core version required for the gear version.
// force update would be equal to current migrate.
skydbctl update -dburl xxx -app xxx -gear xxx [-f] <version>
```

chpapa:
Does it still hold that gear tables start with underscore, to distinguish from user created table? If yes, shall it be renamed, for example from auth_password_history to _auth_password_history?

Steven-Chan:
> Does it still hold that gear tables start with underscore, to distinguish from user created table? If yes, shall it be renamed, for example from auth_password_history to _auth_password_history?

Didn't consider this in this PR, and I'm ok with this, let me update.

Steven-Chan:
@chpapa updated

Steven-Chan:
@carmenlau updated

cmd/migrate/gear/core/1536736169_initialize_schema.up.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+
+CREATE TABLE _sky_user (
+  id text PRIMARY KEY,
+  token_valid_since timestamp without time zone,
+  last_seen_at timestamp without time zone,
+  disabled boolean NOT NULL DEFAULT false,
+  disabled_message text,
+  disabled_expiry timestamp without time zone
+);
```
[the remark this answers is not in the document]
Steven-Chan: This table would be managed by core library.

`AuthInfoStore` in core library would provide `AuthInfo` crud.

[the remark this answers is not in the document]
rickmak: How about the public end-point of authinfo crud, likd disable user. I assume it will host at auth gear. So it will be auth gear responsible for http level, and DB level is by core? Am I correct?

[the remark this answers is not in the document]
Steven-Chan: > it will be auth gear responsible for http level, and DB level is by core

yes

README.md
```diff
@@ -62,6 +62,68 @@ $ make migrate-version
 1. Add migration to update plan table with gear enabled column
 2. Update `Plan` struct and `CanAccessGear` func in `pkg/gateway/model/plan.go`
 
+## Gear DB migration
+
+The following part is about gear db migration.
+
+If you come from skygear-server 0.x to 1.x, the biggest difference is that skygear next would not offer JIT db migration in most gears.
```
chpapa: It might be a bit misleading: It is because we don't expect/recommend gears developed with the SkyRecord API, so JIT DB migration is not offered.

The current statement makes it sounds like JIT DB Migration is another feature.

cmd/migrate/gear/core/1536736169_initialize_schema.up.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+
+CREATE TABLE _sky_user (
```
chpapa: What is the definition of core table (a.k.a. table starts with `_sky`)?

I guess it is not the responsibility of gears? (Because it would be hard not to understand `disabled` is the responsibility of auth instead of core lib), so maybe the definition is: **all gears' table shall not be accessed by other gears (even auth), except those tables under the `_sky` namespace.**?

Steven-Chan: > all gears' table shall not be accessed by other gears (even auth), except those tables under the _sky namespace.

yes.

Maybe some confusion here. See if these updates would make things better:

- Move the migration files from `cmd/migrate/gear/core/` to `cmd/migrate/core`
- Update the command to `go run cmd/migrate/main.go -path cmd/migrate/core -schema app_helloworld -core up`

Steven-Chan:
@chpapa @rickmak @carmenlau updated.

I've separate core migration paths and command from gears.

PR #672, opened 2018-09-13
next: Refactor next core library layout

Steven-Chan:
- Separate interface and implementation in different files but same package
- Rename AuthInfo to AuthContext, AuthInfoResolver to AuthContextResolver
- Move ProviderGraph to package inject

Preparation for moving skydb code to core library.


Steven-Chan:
I was trying to separate interface and implementation, but seems the result is not good, becoz `default` is not a good name as a package.

rickmak:
`default.go` is not very good, but I don't have a good name now.

Rebase first. Let me think about it.

Steven-Chan:
@rickmak rebased

PR #673, opened 2018-09-17
next: Extract AuthInfotore

Steven-Chan:


Steven-Chan:
Accidentally pressed Enter

PR #674, opened 2018-09-17
next: Extract and refactor AuthInfoStore from skydb/pq

Steven-Chan:
Require #672

- Rename ProviderGraph to DependencyMap
- Removed the individual Provider for auth info store and token store
- Removed the dummy pq db and db provider, replaced with SQLBuilder and SQLExecutor
  - SQLBuilder takes care of app name and gear namespace (not strictly)
  - SQLExecutor basically wraps db with logging (mostly copied from skydb/pq/conn_exec.go)

For db connection, with the current DI flow, each dependency will have a new connection for themselves, unless 

- we change the `DependencyMap` interface to let them share the same connection, or
- since it starts becoming a real dependency graph, we use those DI library


Steven-Chan:
@rickmak @carmenlau I've rebased this PR.

@rickmak I've reverted some of the changes you made in one of your previous commit about how the dependencies are assigned in `AuthContextResolver`.

If we want to test the `AuthContextResolver` itself, we can create one by assigning dependencies manually, while if we want to have higher level test, we can create a new `AuthContextResolverFactory` which assign another set of dependencies.

So I simplify the structure, that the factory would assign dependencies itself, instead of getting providers from the caller or the owner.

rickmak:
Good move. It is simplify the structure.

Issue #563, opened 2018-03-16
Support sort by key path


cheungpat:
for example, sort by `conversation.updatedAt` when querying `userConversation`

PR #678, opened 2018-09-25
Support sort by key path

MarkRunWu:
connect #563

### DIFF
--- a/pkg/server/skydb/pq/builder/factory.go
+++ b/pkg/server/skydb/pq/builder/factory.go
@@ -15,6 +15,7 @@
 package builder
 
 import (
+	"errors"
 	"fmt"
 
 	sq "github.com/lann/squirrel"
@@ -259,6 +260,41 @@ func (f *sqlizerFactory) newExpressionPredicateSqlizerForKeyPath(expr skydb.Expr
 	return newExpressionSqlizer(alias, field, expr), nil
 }
 
+func (f *sqlizerFactory) newExpressionSortForKeyPath(expr skydb.Expression) (string, error) {
+	if !expr.IsKeyPath() {
+		panic("expression is not a key path")
+	}
+
+	components := expr.KeyPathComponents()
+	keyPath := expr.Value.(string)
+	if len(components) > 2 {
+		return "", skyerr.NewErrorf(skyerr.RecordQueryInvalid,
+			`keypath "%s" with more than 2 components is not supported`, keyPath)
+	}
+
+	alias := f.primaryTable
+	fields, err := skydb.TraverseColumnTypes(f.db, f.primaryTable, keyPath)
+	if err != nil {
+		return "", skyerr.NewError(skyerr.RecordQueryInvalid, err.Error())
+	}
+
+	field := skydb.FieldType{}
+	for i, keyPathField := range fields {
+		isLast := (i == len(components)-1)
+		field = keyPathField
+		if field.Type == skydb.TypeReference && !isLast {
+			alias = f.createLeftJoin(field.ReferenceType, components[i], "_id")
+		}
+	}
+	lastComponent := components[len(components)-1]
+	sortKey := fullQuoteIdentifier(alias, lastComponent)
+	if alias != f.primaryTable {
+		sortKey = "_sort" + lastComponent + alias
+		f.addExtraColumn(sortKey, skydb.TypeReference, expr, alias)
+	}
+	return sortKey, nil
+}
+
 // createLeftJoin create an alias of a table to be joined to the primary table
 // and return the alias for the joined table
 func (f *sqlizerFactory) createLeftJoin(secondaryTable string, primaryColumn string, secondaryColumn string) string {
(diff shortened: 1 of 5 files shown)


pkg/server/skydb/pq/builder/factory.go
```diff
@@ -326,3 +332,34 @@ type joinedTable struct {
 func (a joinedTable) equal(b joinedTable) bool {
 	return a.secondaryTable == b.secondaryTable && a.primaryColumn == b.primaryColumn && a.secondaryColumn == b.secondaryColumn
 }
+
+func (f *predicateSqlizerFactory) NewSort(s skydb.Sort) (string, error) {
```
Steven-Chan: This function is doing the same thing as `SortOrderBySQL`, so you either need this `NewSort` or `SortOrderBySQL`. It is so unpredictable now, you cannot tell which code path will be executed for each type of expression. Please combine them into one single function.

pkg/server/skydb/pq/builder/factory.go
```diff
@@ -229,7 +230,7 @@ func (f *predicateSqlizerFactory) newExpressionSqlizer(expr skydb.Expression) (e
 		`unexpected expression type`)
 }
 
-func (f *predicateSqlizerFactory) newExpressionSqlizerForKeyPath(expr skydb.Expression) (expressionSqlizer, error) {
+func (f *predicateSqlizerFactory) newExpressionSqlizerForKeyPath(expr skydb.Expression, addJoinedFieldToExtraColumn bool) (expressionSqlizer, error) {
```
Steven-Chan: See `Avoid Side Effects` in https://medium.com/javascript-scene/the-single-biggest-mistake-programmers-make-every-day-62366b432308.

In short, if a function returns something except error, it is expected not to have any side effects.

pkg/server/skydb/pq/builder/factory.go
```diff
@@ -26,6 +26,7 @@ import (
 type PredicateSqlizerFactory interface {
 	UpdateTypemap(typemap skydb.RecordSchema) skydb.RecordSchema
 	AddJoinsToSelectBuilder(q sq.SelectBuilder) sq.SelectBuilder
+	NewSort(s skydb.Sort) (string, error)
```
Steven-Chan: This is a `predicate` `sqlizer` factory, and this new function fits neither.

pkg/server/skydb/pq/builder/factory.go
```diff
@@ -326,3 +332,34 @@ type joinedTable struct {
 func (a joinedTable) equal(b joinedTable) bool {
 	return a.secondaryTable == b.secondaryTable && a.primaryColumn == b.primaryColumn && a.secondaryColumn == b.secondaryColumn
 }
+
+func (f *predicateSqlizerFactory) NewSort(s skydb.Sort) (string, error) {
+	expr := s.Expression
+	if expr.IsKeyPath() && len(expr.KeyPathComponents()) == 2 {
```
Steven-Chan: Please handle `len(expr.KeyPathComponents()) > 2` explicitly.

pkg/server/skydb/pq/builder/sort.go
```diff
@@ -26,7 +26,11 @@ func SortOrderBySQL(alias string, sort skydb.Sort) (string, error) {
 
 	switch sort.Expression.Type {
 	case skydb.KeyPath:
-		expr = fullQuoteIdentifier(alias, sort.Expression.Value.(string))
+		if len(sort.Expression.KeyPathComponents()) == 2 {
```
Steven-Chan: Please handle `len(expr.KeyPathComponents()) > 2` explicitly.

pkg/server/skydb/pq/recorddb.go
```diff
@@ -599,8 +609,14 @@ func columnSqlizersForSelect(recordType string, typemap skydb.RecordSchema) map[
 				Value: column,
 			}
 		}
+		var sqlizer sq.Sqlizer
 
-		sqlizer := builder.NewExpressionSqlizer(recordType, fieldType, expr)
+		if fieldType.Type == skydb.TypeReference && len(expr.KeyPathComponents()) == 2 {
```
Steven-Chan: Please handle `len(expr.KeyPathComponents()) > 2` explicitly.

MarkRunWu: Updated

pkg/server/skydb/pq/builder/sort.go
MarkRunWu: It's not used, removed.

pkg/server/skydb/pq/builder/factory.go
MarkRunWu: Updated to not support expr.KeyPathComponents() > 2 for now same as the current predicate expression.

pkg/server/skydb/pq/builder/factory.go
MarkRunWu: Think the original factory owns joinedTables and extraFields and they are also referenced by assembling sort expression for sorting by reference.
So decide to extends the original factory capability and rename the `PredicateSqlizerFactory` to `SqlizerFactory`

pkg/server/skydb/pq/builder/factory.go
MarkRunWu: Create another method for sorting instead.

pkg/server/skydb/pq/builder/factory.go
MarkRunWu: Done at 31bf84d

Steven-Chan:
@MarkRunWu Thanks for the update!

Though, imo, the `NewSort` function does not really fit into `SqlizerFactory` interface, but that's the problem of existing code (or maybe just me who thinks its a problem :scream:), which may need refactoring on the `SqlizerFactory`.

So I think this PR is good enough to merge.

PR #680, opened 2018-09-27
next: Signup with password

Steven-Chan:
connect #663 

---

I've picked the name `password`, and the table name `_auth_provider_password`, which tries to match the oauth grant type password. But I just realised that we are calling this type of signup/login auth data, e.g. signupWithAuthData and loginWithAuthData in SDK.

@rickmak @carmenlau any thought on this, I suggest we decide it earlier about the naming thing to avoid refactoring and dealing with conflict.


pkg/auth/dependency/provider/password/password.go
```diff
@@ -0,0 +1,64 @@
...
+		sqlExecutor: executor,
+		logger:      logger,
+	}
+}
+
+func (p ProviderImpl) CreateEntry(
+	principalID string,
+	authData interface{},
+	plainPassword string,
+) error {
```
carmenlau: Currently user with same auth data can signup multiple time, I think we will further update this to prevent user duplication?

Steven-Chan: I've added a unique constraints to auth_data so it would be unique.

I think this would a lot simpler and robust than the old skygear-server, that creating and deleting unique constraints for individual keys. Though the downside is the unique key won't exist on user record anymore.

Steven-Chan: Also since we don't have database transection yet, duplicated auth data would create zombie `_auth_principal` data.

pkg/auth/inject.go
```diff
@@ -33,6 +37,18 @@ func (m DependencyMap) Provide(dependencyName string, ctx context.Context, tConf
...
+			db.NewSQLBuilder("auth", tConfig.AppName),
+			db.NewSQLExecutor(ctx, "postgres", tConfig.DBConnectionStr),
+			logging.CreateLogger(ctx, "auth_principal"),
+		)
+	case "PasswordAuthProvider":
+		return password.NewProvider(
+			db.NewSQLBuilder("auth", tConfig.AppName),
+			db.NewSQLExecutor(ctx, "postgres", tConfig.DBConnectionStr),
+			logging.CreateLogger(ctx, "provider_password"),
+		)
```
rickmak: I cannot make sense on creating this two provider.

Seem putting them in AuthInfoProvider will be simpler without much sacrifice.

Steven-Chan: In my mind, `AuthPrincipalStore` and `PasswordAuthProvider` should not combine with `AuthInfoStore` because `AuthInfoStore` is in the core library, which is meant to be imported by all gears, while the other two are in auth gear.

`AuthInfoStore` is the crud layer of `_core_user` (and maybe `_core_role`, `_core_user_role` later), while `AuthPrincipalStore` and `PasswordAuthProvider` are crud layer of `_auth_xxx`.

Steven-Chan: While AuthPrincipalStore and PasswordAuthProvider may indeed be combined to one interface. I think I may revise a little bit on how to better do polymorphism with the `_auth_principal` and `_auth_provider_xxx`.

Steven-Chan:
@rickmak updated

PR #684, opened 2018-09-28
next: Auth user profile config

Steven-Chan:
connect #663

### DIFF
--- a/pkg/auth/handler/signup.go
+++ b/pkg/auth/handler/signup.go
@@ -48,8 +48,7 @@ type SignupRequestPayload struct {
 	Password         string                 `json:"password"`
 	Provider         string                 `json:"provider"`
 	ProviderAuthData map[string]interface{} `json:"provider_auth_data"`
-	// TODO:
-	// RawProfile       map[string]interface{} `json:"profile"`
+	RawProfile       map[string]interface{} `json:"profile"`
 }
 
 func (p SignupRequestPayload) Validate() error {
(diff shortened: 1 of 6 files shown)


pkg/core/inject/inject.go
```diff
@@ -31,13 +32,34 @@ func injectDependency(
...
-		if dependencyName == "" {
+		dependencyTag := t.Field(i).Tag.Get("dependency")
+		if dependencyTag == "" {
 			continue
 		}
 
+		dependencyArgs := strings.Split(dependencyTag, ",")
+		dependencyName := dependencyArgs[0]
+		options := dependencyArgs[1:]
+		optionSet := arrayToSet(options)
```
rickmak: Why optionSet?

I think the arrayToMap that return `map[string]bool` will solve the problem.

pkg/auth/handler/signup.go
```diff
@@ -127,7 +127,11 @@ func (h SignupHandler) Handle(req interface{}, _ handler.AuthContext) (resp inte
 
 	authContext.AuthInfo = &info
 
-	// TODO: create user profile
+	if h.UserProfileStore != nil {
+		if err = h.UserProfileStore.CreateUserProfile(payload.RawProfile); err != nil {
+			return
```
rickmak: TODO: add error log.

pkg/core/inject/inject.go
Steven-Chan: https://www.davidkaya.sk/2017/12/10/sets-in-golang/

> However I prefer the first variant since struct{} takes 0 bytes

pkg/auth/handler/signup.go
Steven-Chan: @rickmak updated

pkg/core/inject/inject.go
rickmak: 💯

Issue #686, opened 2018-10-02
Fix lint error & test case


tensiuyan:

carmenlau:
I am going to help add handlers test case

PR #690, opened 2018-10-03
next: test case for core library

Steven-Chan:
connect #686 

test added:
- core/inject
- core/middleware
- core/auth/authinfo
- core/auth/authz/policy

### DIFF
--- a/pkg/core/auth/authinfo/authinfo_test.go
+++ b/pkg/core/auth/authinfo/authinfo_test.go
@@ -1,0 +1,73 @@
+package authinfo
+
+import (
+	"testing"
+	"time"
+
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestAuthInfo(t *testing.T) {
+	Convey("return new AuthInfo", t, func() {
+		info := NewAuthInfo()
+		So(info.ID, ShouldNotBeEmpty)
+	})
+}
+
+func TestDisabled(t *testing.T) {
+	mockedNow := time.Date(2017, 12, 2, 0, 0, 0, 0, time.UTC)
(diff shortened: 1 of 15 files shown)


pkg/core/auth/authinfo/authinfo_test.go
```diff
@@ -0,0 +1,73 @@
...
+	timeNow = func() time.Time {
+		return mockedNow
+	}
+
+	Convey("Test default is not disabled", t, func() {
+		info := AuthInfo{}
+		So(info.IsDisabled(), ShouldBeFalse)
+	})
+
+	Convey("Test IsDisabled", t, func() {
```
rickmak:
```
  Test IsDisabled 
    default should be false ✔
    should return true if Disabled=true ✔
    should return true if DisabledExpiry is in the future ✔
    should return false if DisabledExpiry is in the past ✔
```

Try to reword is to indicate it is testing `AuthInfo`

May be 

```
Test AuthInfo.IsDisabled 
```

?

pkg/core/middleware/tenant_test.go
```diff
@@ -0,0 +1,85 @@
...
+		req, _ = http.NewRequest("POST", "", nil)
+		return
+	}
+
+	targetMiddleware := TenantConfigurationMiddleware{
+		ConfigurationProvider: ConfigurationProviderFunc(provideConfiguration),
+	}
+	handler := targetMiddleware.Handle(GetTestHandler())
+
+	Convey("handle request without headers", t, func() {
```
rickmak: Group it by one more `Convey`? Or make this message complete sentence?

Steven-Chan:
@rickmak updated

PR #691, opened 2018-10-04
next: Fix api handler bypassed authz policy

Steven-Chan:
Move authz.PolicyProvider from handler to handler factory.

---

Update:

handler.Factory must implement authz.PolicyProvider, reason:

- Optional interface make debug hard, it's hard to trace if the policies are applied to one endpoint until test case or runtime.
- Authentication is mandatory already, so making authorisation mandatory seems also resonable.
- Adding `Allow everybody` policy is easy.
- Authorisation itself is stateless and lightweight, adding it to every handler should have very little performance hit.


Steven-Chan:
@carmenlau @rickmak I've updated the PR to make authz mandatory to every handler, pls see the updated comment for details.

carmenlau:
Looks good to me!:)

rickmak:
Merge via d113d98

PR #695, opened 2018-10-05
next: Signup payload validation

Steven-Chan:
connect #663


Steven-Chan:
It is already handling to user duplication, but the error message is the old one (i.e. `skydb: duplicated UserID or Username`) if displaying the skydb error in handler.

I suggesting doing it in new PR, because it requires moving old code to next packages.

PR #697, opened 2018-10-05
next: Handler error

Steven-Chan:
- encode error to json
- handle error recovered from panic

### DIFF
--- a/cmd/auth/main.go
+++ b/cmd/auth/main.go
@@ -42,6 +42,9 @@ func main() {
 	authContextResolverFactory := resolver.AuthContextResolverFactory{}
 	srv := server.NewServer("localhost:3000", authContextResolverFactory)
 
+	// RecoverMiddleware must come first
+	srv.Use(middleware.RecoverMiddleware{}.Handle)
+
 	if configuration.DevMode {
 		srv.Use(middleware.TenantConfigurationMiddleware{
 			ConfigurationProvider: middleware.ConfigurationProviderFunc(config.NewTenantConfigurationFromEnv),
--- a/cmd/auth/main.go
+++ b/cmd/auth/main.go
@@ -16,9 +16,9 @@ import (
 	"github.com/joho/godotenv"
 	"github.com/skygeario/skygear-server/pkg/auth/handler"
 	"github.com/skygeario/skygear-server/pkg/core/config"
+	"github.com/skygeario/skygear-server/pkg/core/logging"
 	"github.com/skygeario/skygear-server/pkg/core/middleware"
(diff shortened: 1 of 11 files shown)


pkg/core/handler/api.go
```diff
@@ -13,39 +14,42 @@ type APIHandler interface {
...
+		encoder := json.NewEncoder(rw)
+
+		defer func() {
+			if response.Err != nil {
+				httpStatus = nextSkyerr.ErrorDefaultStatusCode(response.Err)
+			}
+
+			rw.Header().Set("Content-Type", "application/json")
+			rw.WriteHeader(httpStatus)
+			encoder.Encode(response)
```
carmenlau: To prevent writing the response multiple times, should avoid writing response in defer function. We did it in `RecoverMiddleware`.

Steven-Chan:
@carmenlau updated

cmd/auth/main.go
```diff
@@ -42,6 +42,9 @@ func main() {
 	authContextResolverFactory := resolver.AuthContextResolverFactory{}
 	srv := server.NewServer("localhost:3000", authContextResolverFactory)
 
+	// RecoverMiddleware must come first
+	srv.Use(middleware.RecoverMiddleware{}.Handle)
```
rickmak: Seem we will not have case that don't want a `Recovery`.

I am thinking instead of leaving a comment to `RecoverMiddleware must come first`. Will we put it as a parameter of `NewServer` as `authContextResolverFactory`? It seem easier to enforce the existence and order.

Steven-Chan:
@rickmak updated

carmenlau:
The last update makes me think of should we also move `RequestIDMiddleware` to NewServer, and `RequestIDMiddleware` should be added before `RecoverMiddleware`. So in case we want to add log to `RecoverMiddleware`, we can have the request id to trace.

pkg/auth/handler/me.go
```diff
@@ -49,6 +49,7 @@ func (h MeHandler) DecodeRequest(request *http.Request) (payload handler.Request
 }
 
 func (h MeHandler) Handle(req interface{}, ctx context.AuthContext) (resp interface{}, err error) {
+	panic(http.ErrAbortHandler)
```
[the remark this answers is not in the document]
Steven-Chan: oops

Steven-Chan:
> Again I think the Middleware should be bound to HTTP semantic, hence the golang specific recover should not belong to that stage of the request processing.

I think for production code, we need a big try catch to at the entry point of the request, hence I put the recovery code in the middleware.

rickmak:
I dropped 374be2a and merge via e40a588

PR #699, opened 2018-10-09
next: reset password handler

kaojohnny:
connects #662

### DIFF
--- a/pkg/auth/dependency/provider/password/password.go
+++ b/pkg/auth/dependency/provider/password/password.go
@@ -89,14 +89,13 @@ func (p ProviderImpl) GetPrincipalByAuthData(authData interface{}, principal *Pr
 	if err != nil {
 		return
 	}
-	builder := p.sqlBuilder.Select("principal_id", "auth_data", "password").
+	builder := p.sqlBuilder.Select("principal_id", "password").
 		From(p.sqlBuilder.TableName("provider_password")).
 		Where(`auth_data @> ?::jsonb`, authDataBytes)
 	scanner := p.sqlExecutor.QueryRowWith(builder)
 
 	err = scanner.Scan(
 		&principal.ID,
-		&principal.AuthData,
 		&principal.HashedPassword,
 	)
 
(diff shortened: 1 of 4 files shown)


rickmak:
@kaojohnny our workflow allow multiple reviewers but one assignees.

pkg/auth/handler/reset_password.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (f ResetPasswordHandlerFactory) NewHandler(request *http.Request) handler.Handler {
+	h := &ResetPasswordHandler{}
+	inject.DefaultInject(h, f.Dependency, request)
+	return handler.APIHandlerToHandler(h)
+}
+
+func (f ResetPasswordHandlerFactory) ProvideAuthzPolicy() authz.Policy {
+	return policy.AllOf(
```
Steven-Chan: This api requires masterkey.

pkg/auth/handler/reset_password.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	authContext := context.AuthContext{}
+	authContext.AuthInfo = &authinfo
+
+	resp = response.NewAuthResponse(authContext, skydb.Record{}, token.AccessToken)
+
+	// Populate the activity time to user
+	now := timeNow()
+	authContext.AuthInfo.LastLoginAt = &now
```
Steven-Chan: According to the old skygear code, `LastLoginAt` and `LastSeenAt` does not have to be set.

pkg/auth/dependency/provider/password/password.go
```diff
@@ -14,7 +14,9 @@ const providerPassword string = "password"
 
 type Provider interface {
 	CreatePrincipal(principal Principal) error
-	GetPrincipal(authData interface{}, principal *Principal) error
+	GetPrincipalByAuthData(authData interface{}, principal *Principal) error
```
rickmak: In both login and signup, the `AuthData` is `map[string]interface{}` not `interface{}. We should go for a more specific one here.

Steven-Chan: I used `interface{}` instead of `map[string]interface{}` because I thought what this function requires is something that can be transformed to json.

rickmak: I think the interface here should not concern about the storing format.

Re the authData is `interface` or `map[string]interface{}`, I think it depends on our actual system and use-case. I think in our specification, it must be a `map[string]interface{}` https://github.com/SkygearIO/features/blob/master/features/48-user-profile-discovery/user-profile.md

Issue #682, opened 2018-09-28
Transaction management in next


rickmak:
The party involve are:

- HandlerFactory
- DependencyMap
- Handler

There is handler that require fine control on transcation depending on the actual payload. The transaction is management is application logic. We need to provide a way for handler to control over the transaction.

chpapa:
@Steven-Chan If it is only becoz Record API need to have multiple transactions in a single request and complicate everything, my gut feeling is that's a good tradeoff to keep thing simple but only support atomic request in record API... (just 2 cents)

Steven-Chan:
> If it is only becoz Record API need to have multiple transactions in a single request and complicate everything

Yes, but not entirely. I am not sure if the record save API is the only one (now or future) which require fine control on db transaction.

If we go for proposal 2, that will straightly shut down such ability. I don't think such sacrifice is actually worth, unless we find a way to allow handler to have ability to fine control, or we have a philosophy on skygear API design that each request must be atomic on db operation.

Steven-Chan:
Did some searching on current skygear-server code base.

At handler level:
- create auth db data and create user record in `authutil.go`
- `atomicModifyFunc` in record save handler

At preprocessor level:
- create user in preprocessor

At skydb/pq level:
- `SetRecordFieldAccess`
- migration
- schema `Extend`

---

Notes:

- Transaction management at `skydb/pq` level would be removed.
- Preprocessor is removed already
- We have not yet provided the exact same replacement of `create user in preprocessor` in next yet
- create auth and create user record in `authutil.go`
  - we have decided not to keep one transaction for this

--- 

TODO: check existing plugin apis

PR #701, opened 2018-10-09
next: DB transaction

Steven-Chan:
connect #682

I have made some big changes to the interface, since handler.Handler is removed and http.Handler is used instead. The AuthContext would need dependency to provide interface for access

### DIFF
--- a/pkg/auth/handler/login.go
+++ b/pkg/auth/handler/login.go
@@ -13,6 +13,7 @@ import (
 	"github.com/skygeario/skygear-server/pkg/core/auth/authtoken"
 	"github.com/skygeario/skygear-server/pkg/core/auth/authz"
 	"github.com/skygeario/skygear-server/pkg/core/auth/authz/policy"
+	"github.com/skygeario/skygear-server/pkg/core/db"
 	"github.com/skygeario/skygear-server/pkg/core/handler"
 	"github.com/skygeario/skygear-server/pkg/core/inject"
 	"github.com/skygeario/skygear-server/pkg/core/server"
@@ -39,7 +40,7 @@ type LoginHandlerFactory struct {
 func (f LoginHandlerFactory) NewHandler(request *http.Request) http.Handler {
 	h := &LoginHandler{}
 	inject.DefaultInject(h, f.Dependency, request)
-	return handler.APIHandlerToHandler(h)
+	return handler.APIHandlerToHandler(h, h.TxContext)
 }
 
 func (f LoginHandlerFactory) ProvideAuthzPolicy() authz.Policy {
@@ -73,6 +74,11 @@ type LoginHandler struct {
 	PasswordAuthProvider password.Provider           `dependency:"PasswordAuthProvider"`
 	UserProfileStore     dependency.UserProfileStore `dependency:"UserProfileStore,optional"`
 	AuditTrail           *audit.Trail                `dependency:"AuditTrail"`
+	TxContext            db.TxContext                `dependency:"TxContext"`
+}
+
+func (h LoginHandler) WithTx() bool {
+	return true
 }
 
 // ProvideAuthzPolicy provides authorization policy
--- a/pkg/core/db/context.go
+++ b/pkg/core/db/context.go
@@ -1,0 +1,141 @@
+package db
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/jmoiron/sqlx"
+)
+
+type contextKey string
+
+var (
+	keyContainer = contextKey("container")
+)
+
+// DB provides the interface for retrieving an interface to execute sql
+type Context interface {
+	DB() ExtContext
+}
+
+// Tx provides the interface for managing transaction
+type TxContext interface {
+	HasTx() bool
+	BeginTx() error
+	CommitTx() error
+	RollbackTx() error
+}
+
+// TODO: handle thread safety
+type contextContainer struct {
+	db *sqlx.DB
+	tx *sqlx.Tx
+}
+
+type dbContext struct {
+	context.Context
+	dbOpener func() (*sqlx.DB, error)
+}
+
+// InitRequestDBContext initialize db context for the request
+func InitRequestDBContext(req *http.Request) *http.Request {
+	container := &contextContainer{}
+	return req.WithContext(context.WithValue(req.Context(), keyContainer, container))
+}
+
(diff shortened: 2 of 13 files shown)


Steven-Chan:
WIP: I want to revert some functions that rely on http.Request back to rely on context.Context.

pkg/auth/handler/login.go
```diff
@@ -72,6 +72,12 @@ type LoginHandler struct {
 	AuthInfoStore        authinfo.Store              `dependency:"AuthInfoStore"`
 	PasswordAuthProvider password.Provider           `dependency:"PasswordAuthProvider"`
 	UserProfileStore     dependency.UserProfileStore `dependency:"UserProfileStore,optional"`
+	TxContext            db.TxContext                `dependency:"TxContext"`
+}
+
+// WithTx describes if the api handler requires transaction
+func (h LoginHandler) WithTx() bool {
+	return true
 }
```
rickmak: Let don't define this at every handler as dependency?

I prefer this optional at the respective Factory.NewHandler.

pkg/core/db/context.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (d *dbContext) tx() *sqlx.Tx {
+	return d.container().tx
+}
+
+func (d *dbContext) lazydb() *sqlx.DB {
+	db := d.db()
+	if db == nil {
+		var err error
+		if db, err = d.dbOpener(); err != nil {
```
rickmak: This assume the `d.db()` only have one DB as per context. However, the `dbOpener` is pass in func, that can be open to different DB.

It will have bug if different Store are opening different pg connection.

pkg/core/server/server.go
```diff
@@ -44,22 +46,25 @@ func NewServer(
 // Handle delegates gorilla mux Handler, and accept a HandlerFactory instead of Handler
 func (s *Server) Handle(path string, hf handler.Factory) *mux.Route {
 	return s.router.NewRoute().Path(path).Handler(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		r = auth.InitRequestAuthContext(r)
+		r = db.InitRequestDBContext(r)
```
rickmak: This init is required?

How about if a handler is not related to an DB operation.

pkg/auth/handler/login.go
Steven-Chan: I actually prefer having this function explicitly.

If later someone want to figure out whether a handler is wrapped inside a transaction, this function can tell immediately.

rickmak:
There is also some concern on the implicit control on `Store` transaction in handler.

Only handler can control the transaction. It is good to avoid Deadlock with the `Store`. However, it impose the transaction properties to Store implicitly.

For example,

- Record Handler decode request
- determine weather or not to open an transaction
- Perform AuthProfileStore. (this Store don't know it is within an transaction or not)

This raise some concern. But I cannot think of an actual case that have problem yet.

pkg/core/server/server.go
Steven-Chan: This init does really interact with the DB,

It can be removed actually if everything is lazily created.

Steven-Chan: Oh, sorry, this init is required.

Because context in request is immutable, creating the container lazily would not work.

Steven-Chan:
> Perform AuthProfileStore. (this Store don't know it is within an transaction or not)

I think this is the root of problem of creating transaction Deadlock.

When someone developing a `Store` which would execute multiple SQL, the immediate thought would be to make a transaction to ensure the SQL are executed atomically.

I made this design principal based on the idea that `Store` should not have control on transaction.

Steven-Chan:
I would try to split this PR into two, one for changing the handler interface, one for db transaction, so that this PR would not block other issues.

Steven-Chan:
Here is a diagram try to explain how the db transaction works in this PR.

```
                                 +---------------+
                                 |               |
                                 |    Handler    |
   +----------------------------->    Factory    |
   |                             |               |
   |                             |               |
   |                             +-------+-------+
   |                                     | Create
   |                                     |
   |                             +-------v-------+
   |                             |               |
   |                             |               |
   |                             |    Handler    |
   |                             |               |
   |                             |               |
   |                             +-------+-------+
   |                                     | Has a
   |                                     |
   |                        +------------+------------+
   |                        |                         |
   |                        |                         |
   |                +-------v-------+         +-------v-------+
   | Initiate       |               |         |               |
   | Factory to     |               |         |               |
   | create Handler |    AStore     |         |    BStore     |
+--+------------+   |               |         |               |
|               |   |               |         |               |
|               |   +-------+-------+         +-------+-------+
|    Request    |           | Has a                   | Has a
|               |           |                         |
|               |   +-------v-------+         +-------v-------+
+--+------------+   |               |         |               |
   | Has a          |               |         |               |
   |                |  SQLExecutor  |         |  SQLExecutor  |
   |                |               |         |               |
   |                |               |         |               |
   |                +-------+-------+         +-------+-------+
   |                        | Has a                   | Has a
   |                        |                         |
   |                +-------v-------+         +-------v-------+
   |                |               |         |               |
   |          Has a |               |         |               |
   |         +------+  DB Context   |         |  DB Context   |
   |         |      |               |         |               |
   |         |      |               |         |               |
   |         |      +-------+-------+         +-------+-------+
   |         |              |                         |
   |         |              |          Access         |
   |         |              +------------+------------+
   |         |                           |
   |         |                           |
+--v---------v--+                +-------v-------+  For the same request,
|               |                |               |  all Store get access to
|    Request    | Has a          |    DB / Tx    |  the same container.
|    Context    +---------------->    Container  |
|               |                |               |  Thus they execute SQL on
|               |                |               |  the same DB / Tx.
+---------------+                +---------------+

```

Steven-Chan:
@rickmak I'm thinking about doing a less magical version, that use the following interface.

```
type Store interface {
	CreateAuth(ctx context.Context, authinfo *AuthInfo) error
	GetAuth(ctx context.Context, id string, authinfo *AuthInfo) error
	UpdateAuth(ctx context.Context, authinfo *AuthInfo) error
	DeleteAuth(ctx context.Context, id string) error
}
```

rickmak:
Let merge it with the document attached as the part of PR. I think the design is fine if it is well informed.

Steven-Chan:
@rickmak rebased

Issue #220, opened 2016-11-08
Sanitize log output for sensitive information


cheungpat:
Since people will be posting their log for debug, it is better if we sanitize the log output for hiding sensitive information (such as access token, api key, master key and password) to protect their personal data.

tensiuyan:
note: we need a full list of sensitive information

kaojohnny:
https://oursky.quip.com/1rezANjuu4Yt

chpapa:
@kaojohnny thanks for putting the list and plan together!

A reminder is I think it might be better for the document to be on Github as comment (or a spec in skygeario/feature) because it is suppose to be an open source project (mostly sponsor by Oursky lol) we should try to keep information to public instead of private document within Oursky Quip.

I know we have been doing quite a lot of it for NEXT refactor, so it is just a reminder for all of us to beware of what should live in opensource and what should be in the company.

PR #704, opened 2018-10-11
next: sanitize log

kaojohnny:
supports:

1. master key, API key
2. JWT access token

connects #220

### DIFF
--- a/pkg/auth/inject.go
+++ b/pkg/auth/inject.go
@@ -92,8 +92,5 @@ func (m DependencyMap) Provide(dependencyName string, r *http.Request) interface
 
 func sensitiveLoggerValues(r *http.Request) []string {
 	tConfig := config.GetTenantConfig(r)
-	return []string{
-		tConfig.APIKey,
-		tConfig.MasterKey,
-	}
+	return tConfig.DefaultSensitiveLoggerValues()
 }
--- a/pkg/auth/inject.go
+++ b/pkg/auth/inject.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 
 	"github.com/jmoiron/sqlx"
+	"github.com/sirupsen/logrus"
 	"github.com/skygeario/skygear-server/pkg/auth/dependency"
 	"github.com/skygeario/skygear-server/pkg/auth/dependency/provider/password"
 	coreAudit "github.com/skygeario/skygear-server/pkg/core/audit"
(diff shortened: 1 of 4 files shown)


kaojohnny:
@Steven-Chan Rebased. Thanks.

cmd/auth/main.go
```diff
@@ -22,7 +24,10 @@ import (
 )
 
 type configuration struct {
-	DevMode bool `envconfig:"DEV_MODE"`
+	DevMode           bool   `envconfig:"DEV_MODE"`
+	APIKey            string `envconfig:"API_KEY"`
+	MasterKey         string `envconfig:"MASTER_KEY"`
+	CustomTokenSecret string `envconfig:"CUSTOM_TOKEN_SECRET"`
```
carmenlau: In next, api key, master key and custom token is provided by the tenant config which is different per request. After discussion, we support provide the tenant config when creating logger and the formatter should be set there.

pkg/core/logging/MaskFormatter.go
```diff
@@ -0,0 +1,55 @@
...
+	return regexp.Compile(p)
+}
+
+type MaskFormatter struct {
+	MaskPatterns     []*regexp.Regexp
+	DefaultFormatter logrus.Formatter
+}
+
+func (f *MaskFormatter) Format(entry *logrus.Entry) ([]byte, error) {
+	message := entry.Message
```
carmenlau: Besides the message, we should also check and mask the log fields

pkg/core/logging/MaskFormatter.go
```diff
@@ -0,0 +1,55 @@
+// Copyright 2015-present Oursky Ltd.
```
carmenlau: File name in go should be lower case with underscore.

pkg/auth/inject.go
```diff
@@ -53,13 +54,15 @@ func (m DependencyMap) Provide(dependencyName string, r *http.Request) interface
...
 		return password.NewProvider(
 			db.NewSQLBuilder("auth", tConfig.AppName),
 			db.NewSQLExecutor(r.Context(), db.NewContextWithContext(r.Context(), openDB(tConfig))),
-			logging.CreateLogger(r, "provider_password"),
+			logging.CreateLogger(r, "provider_password", maskFormatter),
 		)
 	case "HandlerLogger":
-		return logging.CreateLogger(r, "handler")
+		maskFormatter := logging.CreateMaskFormatter(sensitiveLoggerValues(r), &logrus.TextFormatter{})
+		return logging.CreateLogger(r, "handler", maskFormatter)
```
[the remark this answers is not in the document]
kaojohnny: I pass `maskFormatter` as parameter due to not sure I think we may want to configure formatter in some cases. For example, we may want to use `JSONFormatter` in some module in stead of `TextFormatter`. 

And second, I think `logger` module should work without knowing application-level logic.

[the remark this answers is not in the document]
rickmak: I think the logger module should work without knowing the gear specific logic.

Gear is our application. So if a logic is univerial enough that will apply too all skygear gear, it can be the logic in the core lib.

So the maskFormatter fall into this category, and it can be in core lib.

The discussion should focus on “are different mask is required for different gear and different tanent?”
My answer is yes, we may allow the tanent to configure log to remote logger/without the mastk.

pkg/auth/inject.go
```diff
@@ -84,3 +87,11 @@ func (m DependencyMap) Provide(dependencyName string, r *http.Request) interface
...
 	}
 }
+
+func sensitiveLoggerValues(r *http.Request) []string {
+	tConfig := config.GetTenantConfig(r)
+	return []string{
+		tConfig.APIKey,
+		tConfig.MasterKey,
+	}
+}
```
rickmak: Missed this, this should belong to core, not auth.

kaojohnny: @rickmak updated. thanks.

PR #705, opened 2018-10-12
next: get role

kaojohnny:
connects #670

## Note

Note that, due to some implementation detail is not confirmed yet, this endpoint will only support admin only temporarily. Developers should fix this API to support user to get its own roles later.

# Not confirmed implementation:
For some special cases, API may have different behavior when incoming request arguments are different. Take `/role/get` as an example, the API has different permission policy according to request access key type (master key or non master key). We should come up with an elegant design to support such behavior.

### DIFF
--- a/pkg/auth/handler/get_role.go
+++ b/pkg/auth/handler/get_role.go
@@ -1,0 +1,112 @@
+package handler
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/skygeario/skygear-server/pkg/auth"
+	"github.com/skygeario/skygear-server/pkg/core/auth/authinfo"
+	"github.com/skygeario/skygear-server/pkg/core/auth/authz"
+	"github.com/skygeario/skygear-server/pkg/core/auth/authz/policy"
+	"github.com/skygeario/skygear-server/pkg/core/db"
+	"github.com/skygeario/skygear-server/pkg/core/handler"
+	"github.com/skygeario/skygear-server/pkg/core/inject"
+	"github.com/skygeario/skygear-server/pkg/core/server"
+	"github.com/skygeario/skygear-server/pkg/server/skyerr"
+)
+
+func AttachGetRoleHandler(
+	server *server.Server,
+	authDependency auth.DependencyMap,
+) *server.Server {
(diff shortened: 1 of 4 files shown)


pkg/auth/handler/get_role.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (f GetRoleHandlerFactory) NewHandler(request *http.Request) http.Handler {
+	h := &GetRoleHandler{}
+	inject.DefaultInject(h, f.Dependency, request)
+	return handler.APIHandlerToHandler(h)
+}
+
+func (f GetRoleHandlerFactory) ProvideAuthzPolicy() authz.Policy {
+	return policy.AllOf(
+		// FIXEM: this endpoint should hanlde request with master key or with access key
```
rickmak: s/FIXEM/FIXME/

pkg/auth/handler/get_role.go
```diff
@@ -0,0 +1,79 @@
...
+
+type GetRoleRequestPayload struct {
+	UserIDs []string `json:"users"`
+}
+
+func (p GetRoleRequestPayload) Validate() error {
+	return nil
+}
+
+// GetRoleHandler handles signup request
```
rickmak: // GetRoleHandler handles signup request
->
// GetRoleHandler allow admin user to query user roles.

rickmak: Please add an example input and outpur as in https://github.com/SkygearIO/skygear-server/blob/next/pkg/auth/handler/role_assign.go#L62

pkg/auth/handler/get_role.go
```diff
@@ -0,0 +1,79 @@
...
+	AuthInfoStore authinfo.Store `dependency:"AuthInfoStore"`
+}
+
+func (h GetRoleHandler) DecodeRequest(request *http.Request) (handler.RequestPayload, error) {
+	payload := GetRoleRequestPayload{}
+	err := json.NewDecoder(request.Body).Decode(&payload)
+	return payload, err
+}
+
+// Handle getting roles of users specified by user IDs. Users can only
```
rickmak: Mark TODO, since this comment is not valid yet.

rickmak:
Merge via 48d28ec

PR #707, opened 2018-10-15
next: me handler

Steven-Chan:
connect #664

### DIFF
--- a/pkg/auth/handler/me.go
+++ b/pkg/auth/handler/me.go
@@ -4,13 +4,18 @@ import (
 	"net/http"
 
 	"github.com/skygeario/skygear-server/pkg/auth"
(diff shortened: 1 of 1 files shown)


pkg/auth/handler/me.go
```diff
@@ -57,6 +64,26 @@ func (h MeHandler) DecodeRequest(request *http.Request) (payload handler.Request
 }
 
 func (h MeHandler) Handle(req interface{}) (resp interface{}, err error) {
-	resp = h.AuthContext.AuthInfo()
+	authInfo := h.AuthContext.AuthInfo()
+
+	// refresh access token with a newly generated one
```
rickmak: Please move this inline comment to descriptive comment at Handler.

Also, please add example input and output.

Steven-Chan:
@rickmak updated

PR #711, opened 2018-10-15
next: add CORS middleware to gear

kaojohnny:
connects #710


Steven-Chan:
> 1. Should we have different CORS setting per app? If yes, we may want to get it from the tenant config in request.
>2. In order to let the gear can work in dev mode, maybe we should put the middleware in the gear level?

agree

kaojohnny:
@carmenlau @Steven-Chan Updated, please help review. Thanks.

Issue #708, opened 2018-10-15
Record save handler


tensiuyan:

Steven-Chan:
Note on some offline discussion with @rickmak and @carmenlau about naming on database table for record gears.

In v1, we have a convention that `_xxx` are skygear reserved tables and others are user defined tables.

In next, we have a convention that `_{gear}_xxx` are tables for each gear. While record gear has some reserved tables itself, thus we need to define a format to distinguish user defined tables.

The conclusion is, 
- `_record_xxx` are record gear reserved tables
- `_record__xxx` are user defined table, e.g. `_record__user`, `_record__category`, `_record__label`

chpapa:
@Steven-Chan @rickmak @carmenlau 
Have we considered the option of anything not started with `_` are user-defined tables? So it would be simply `user`, `category`, `label` etc... I thought it is better because:

1. People directly connect to DB is easier / type a few fewer characters.
1. Avoid breaking some convention when other plugins need to access user-defined tables, that they are actually accessing the `_record` namespace.

Issue #424, opened 2017-07-13
Revamp recordSaveHandler in recordutil to a more flatten style


Steven-Chan:
I found that there are multiple layers of function that are initiating transaction which may cause database locking / exception thrown (it depends on how the transaction is started). For example, with dev mode, atomic save record while the save would cause schema extend.

I suggest to move the responsibility of making db transaction to handler functions.

But then I find the current implementation of recordSaveHandler in recordutil is a hard to refactor for this. So I suggest to re-organize the recordSaveHandler to a more flatten style, which would make db transaction managament easier.

Here is the pseudo-code:
```
before refactor:

func RecordSaveHandler(records):
  for record in records:
    fetchOrCreate(record)
    updateMetaData(record)

  for record in records:
    applyDefaultACL(record)

  for record in records:
    triggerBeforeSaveHook(record)

  beginTx()   // You don't see this in current RecordSaveHandler, they are inside db.Extend(...)
  extendSchema(getSchemaMergeMap(records))
  endTx()     // You don't see this in current RecordSaveHandler, they are inside db.Extend(...)

  for record in records:
    save(getRecordDelta(record))

  for record in records:
    injectAsset(record)

  for record in records:
    triggerAfterSaveHook(record)

func HigherOrderRecordSaveHandler():
  if atomic:
    beginTx()

  RecordSaveHandler(recordsToSave)

  if atomic:
    endTx()
```
```
after refactor:

func RecordSaveHandler(record):
  fetchOrCreate(record)
  updateMetaData(record)

  applyDefaultACL(record)

  triggerBeforeSaveHook(record)

  extendSchema(getSchemaMergeMap(record))  // no more transaction inside db.Extend(...)

  save(getRecordDelta(record))

  injectAsset(record)

  triggerAfterSaveHook(record)


func RecordsSaveHandler(records):
  if atomic:
    beginTx()
    for record in records:
      RecordSaveHandler(record)
    endTx()
  else:
    for record in records:
      beginTx()
      RecordSaveHandler(record)
      endTx()
```

The biggest difference is that, for non-atomic record save, the transaction would live in the db actions for each single record to save. So if one record save failed, all the transaction of that particular record save would be rolled back.

The implication is the trigger point of before/after-save would be changed, and what db schema the plugin would see in the before/after-save. I feel a bit uncomfortable if this change would break anything.

===

[...]

PR #718, opened 2018-10-17
wip: next: record save handler

Steven-Chan:
connect #708 

This is a super large PR, because record save involves the largest part of skygear-server.

Note:

- db migration script need to be run

in progress:

- asset store
- fully remove database id


Steven-Chan:
To run test and lint on travis correctly, continue to deal with this PR after #725.

Steven-Chan:
@rickmak Please review e2376ec542ed0d4b9eca05585bb689fa86d03211, it's about transaction handling in record save handler.

rickmak:
@Steven-Chan May be merge 36ac9bb into master first? I think the transaction part will need some updates.

pkg/record/handler/save.go
```diff
@@ -361,13 +372,62 @@ func (f RecordFetcher) FetchOrCreateRecord(recordID record.ID, authInfo *authinf
...
 // 4. Populate meta data and save the record (like updated_at/by)
 // 5. Execute after save hooks with original record and new record
-func RecordSaveHandler(req *RecordModifyRequest, resp *RecordModifyResponse) skyerr.Error {
+// nolint: gocyclo
+func RecordSaveHandler(req *RecordModifyRequest, resp *RecordModifyResponse) (err error) {
 	records := req.RecordsToSave
 
-	fetcher := NewRecordFetcher(req.Context, req.RecordStore, req.WithMasterKey)
+	// Open trasaction for whole record save operation
+	if req.Atomic {
```
rickmak: Is it possible to move all atomic related code to the `RecordSaveHandler.Handle`? Keep this RecordSaveHandler transaction unaware may be better separate of concern.

This `RecordSaveHandler` already complicated enough.

pkg/record/handler/save.go
```diff
@@ -361,13 +372,62 @@ func (f RecordFetcher) FetchOrCreateRecord(recordID record.ID, authInfo *authinf
 // 3. Clean up some transport only data (sequence for example) away from record
 // 4. Populate meta data and save the record (like updated_at/by)
 // 5. Execute after save hooks with original record and new record
-func RecordSaveHandler(req *RecordModifyRequest, resp *RecordModifyResponse) skyerr.Error {
+// nolint: gocyclo
```
rickmak: 💣

pkg/record/handler/save.go
```diff
@@ -683,3 +744,93 @@ func scrubRecordFieldsForRead(authInfo *authinfo.AuthInfo, r *record.Record, fie
 		}
 	}
 }
+
+func ExtendRecordSchema(recordStore record.Store, logger *logrus.Entry, records []*record.Record) (bool, error) {
```
rickmak: I think `ExtendRecordSchema` can be transaction aware?

Steven-Chan:
> May be merge 36ac9bb into master first? I think the transaction part will need some updates.

@rickmak done

Steven-Chan:
@rickmak I really want to bring this issue back https://github.com/SkygearIO/skygear-server/issues/424 😢 , though I know u don't want to make too much change in next.

Steven-Chan:
@rickmak updated

Steven-Chan:
@rickmak @carmenlau @kaojohnny I've pushed a new commit, which fixed a bug I found with test case, and seems I found a better way to organise the transaction code.

pkg/record/handler/save.go
```diff
@@ -182,75 +182,74 @@ func (h SaveHandler) Handle(req interface{}) (resp interface{}, err error) {
...
+
+	return
+}
+
+// ExtendRecordSchemaWithTx ensure the operation is within a transaction.
+// When the request is inatomic, the operation would be wrapped in a new transaction.
+func (h SaveHandler) ExtendRecordSchemaWithTx(payload SaveRequestPayload) (err error) {
 	if !payload.Atomic {
-		if err = h.TxContext.CommitTx(); err != nil {
+		if err = h.TxContext.BeginTx(); err != nil {
```
kaojohnny: I am a little confused when the operation is atomic, `BeginTx` in `Handler`, when the operation is inatomic, when have `BeginTx` in `ExtendRecordSchemaWithTx`.

Steven-Chan: `ExtendRecordSchema` executes multiple sql statement, and this function should be atomic no matter if the record save is atomic or not.

At handler level, it should not be aware of whether it need to begin transaction for the underlying store. Either I don't think the store implementation should be aware of that.
 
But I can't find a simple way to do it yet, that's why I put the transaction here 😢.

kaojohnny: Thanks, I can understand now.

Do you think it would be better if we have an `ensureTx()`? so `ExtendRecordSchemaWithTx` won't be related to `Atomic` idea? 

Just an idea.

Steven-Chan: The `ExtendRecordSchemaWithTx` imo should be related to `atomic`, since at handler level, only when the request is inatomic the function should begin a separate transaction. If the `HasTx` return false in `ExtendRecordSchemaWithTx ` when the request is atomic, it should be considered as programming error.

Steven-Chan: I guess maybe eventually I will just add a `ensureTx()` in every underlying function that execute more than one sql statement 😢.

kaojohnny: Thanks. I have no further problems. See if @rickmak okay for the changes.

Steven-Chan:
To @kaojohnny , we've a short offline discussion on this topic. I suggested that the underlying store functions could be wrapped with a transaction if there does not exist one when entering the function. This could make the functions behave atomically, which from the point of encapsulation, makes more sense than having functions that may throw errors and stay at unexpected state.

However, we think this would result in some magic that we do not want to have, which make us hard to trace where transactions are began.

We came up with a solution that, instead we throw error if transaction does not exist when entering functions. And thus handler code, where transaction is required, would be the same as current implementation. For code that does not have transaction now, would require beginning transaction. I expect this would be quite a lot of code change. So I propose to have another PR to deal with it.

cc @rickmak @carmenlau

pkg/record/handler/save.go
```diff
@@ -182,20 +182,37 @@ func (h SaveHandler) Handle(req interface{}) (resp interface{}, err error) {
...
-			err = myerr
+	// Open transaction for whole operation if atomic save
+	if payload.Atomic {
+		if err = h.TxContext.BeginTx(); err != nil {
 			return
 		}
 
-		err = skyerr.NewError(skyerr.IncompatibleSchema, "failed to migrate record schema")
+		defer func() {
+			if txErr := db.EndTx(h.TxContext, err); txErr != nil {
```
rickmak: Don't resue the `err`. This defer only intended to catch error below. Separate the variable before and after.

Steven-Chan:
@rickmak updated

PR #719, opened 2018-10-17
next: record gear setup

Steven-Chan:
connect #708

### DIFF
--- a/cmd/record/main.go
+++ b/cmd/record/main.go
@@ -21,7 +21,8 @@ import (
 )
 
 type configuration struct {
-	DevMode bool `envconfig:"DEV_MODE"`
+	DevMode bool   `envconfig:"DEV_MODE"`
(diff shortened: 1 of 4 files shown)


cmd/record/main.go
```diff
@@ -0,0 +1,79 @@
...
+	configuration := configuration{}
+	envconfig.Process("", &configuration)
+
+	// logging initialization
+	logging.SetModule("record")
+
+	recordDependency := record.NewDependencyMap()
+
+	authContextResolverFactory := resolver.AuthContextResolverFactory{}
+	srv := server.NewServer("localhost:3000", authContextResolverFactory)
```
carmenlau: Record gear is using same port as the auth gear. It is ok for deployment as they are running in different container. But may be inconvenient for development?
Ideally, it would be great to be configurable.

Steven-Chan:
@carmenlau updated

PR #720, opened 2018-10-18
next: sign up anonymously

kaojohnny:
connects #716

### DIFF
--- a/pkg/auth/dependency/provider/anonymous/anonymous.go
+++ b/pkg/auth/dependency/provider/anonymous/anonymous.go
@@ -1,0 +1,48 @@
+package anonymous
+
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/skygeario/skygear-server/pkg/core/db"
+)
+
+const providerAnonymous string = "anonymous"
+
+type Provider interface {
+	CreatePrincipal(principal Principal) error
+}
+
+type ProviderImpl struct {
+	sqlBuilder  db.SQLBuilder
+	sqlExecutor db.SQLExecutor
+	logger      *logrus.Entry
+}
+
+func NewProvider(builder db.SQLBuilder, executor db.SQLExecutor, logger *logrus.Entry) *ProviderImpl {
+	return &ProviderImpl{
+		sqlBuilder:  builder,
+		sqlExecutor: executor,
+		logger:      logger,
+	}
+}
+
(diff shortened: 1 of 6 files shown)


pkg/auth/dependency/provider/anonymous/anonymous.go
```diff
@@ -0,0 +1,48 @@
+package anonymous
+
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/skygeario/skygear-server/pkg/core/db"
+)
+
+const providerPassword string = "anonymous"
```
Steven-Chan: `providerPassword` => `providerAnonymous`

pkg/auth/dependency/provider/anonymous/anonymous.go
```diff
@@ -0,0 +1,48 @@
...
+	// TODO: log
+
+	// Create principal
+	builder := p.sqlBuilder.Insert(p.sqlBuilder.TableName("principal")).Columns(
+		"id",
+		"provider",
+		"user_id",
+	).Values(
+		principal.ID,
+		providerPassword,
```
rickmak: I remember the V1 implementation of "anonymous" user is a user with empty password.

Am I correct? I think in this next refractor should not change it? It will make the data migration before difficult if not make it impossible.

pkg/auth/dependency/provider/anonymous/anonymous.go
rickmak: Since it is private, I can be just `providerName`. But for consistent with provider/password. it can be providerAnonymous.

pkg/auth/dependency/provider/anonymous/anonymous.go
Steven-Chan: > I remember the V1 implementation of "anonymous" user is a user with empty password.

Yes.

But actually what I intended to do when designing the next auth database is the same as what this PR does.

Steven-Chan:
> A bit concern to chance too much on the DB semantic.

@rickmak since when we split up core, auth and record, the `user` has been quite different than that of v1.

To me, as long as the API and feature set is maintained, the underlying change should still be controllable.

Steven-Chan:
Though, I think `"anonymous" user is a user with empty password` can still fall into the next auth principal design. I prefer this PR over keeping the semantics in v1, because the new design is more clear.

PR #721, opened 2018-10-18
next: record save (part 1)

Steven-Chan:
connect #708 
require #719 

This PR is mainly about moving code from skydb to record gear with some small refactoring to make it compile successfully.


Steven-Chan:
Let me rebase first, because the sqlbuilder has some update, auth gear should not be working.

PR #742, opened 2018-10-30
next: refactor transaction management

Steven-Chan:
connect #682

There is a safeXXXStore for each resource that interact with database, which would panic if function called without transaction.


pkg/auth/dependency/provider/anonymous/provider.go
```diff
@@ -11,14 +11,18 @@ type providerImpl struct {
...
+func newProvider(builder db.SQLBuilder, executor db.SQLExecutor, logger *logrus.Entry) *providerImpl {
 	return &providerImpl{
 		sqlBuilder:  builder,
 		sqlExecutor: executor,
 		logger:      logger,
 	}
 }
 
+func NewProvider(builder db.SQLBuilder, executor db.SQLExecutor, logger *logrus.Entry) Provider {
+	return NewProvider(builder, executor, logger)
```
carmenlau: s/NewProvider/newProvider/

pkg/core/skyerr/recover.go
```diff
@@ -1,13 +1,15 @@
...
 import (
+	"fmt"
 	"runtime/debug"
 
 	"github.com/sirupsen/logrus"
 	"github.com/skygeario/skygear-server/pkg/server/skyerr"
 )
 
 func ErrorFromRecoveringPanic(r interface{}) skyerr.Error {
+	fmt.Printf("%+v\n", r)
```
carmenlau: Debug print?

pkg/record/handler/save.go
```diff
@@ -522,10 +525,44 @@ func RecordSaveHandler(req *RecordModifyRequest, resp *RecordModifyResponse) (er
...
+			return do()
+		})
+		if txErr != nil {
+			err = skyerr.MakeError(txErr)
+		}
+
+		return
+	}
+
+	err = do()
```
carmenlau: Should we also wrap this error with `skyerr.MakeError`? So the behaviour of atomic and non-atomic is the same

Steven-Chan:
> Should we also wrap this error with `skyerr.MakeError`? So the behaviour of atomic and non-atomic is the same

@carmenlau  I have updated the code to remove `skyerr.MakeError` when inatomic, because APIHandler does not require the handle function to return a skygear error, it is the responsibility of the transformation function to make it skygear error.

PR #747, opened 2018-10-31
next: create default profile store

kaojohnny:
Default store is used when record gear is not configured.

connects #754


pkg/auth/dependency/userprofile/store.go
```diff
@@ -0,0 +1,89 @@
...
+	(*userProfile)["_id"] = "user/" + userID
+	(*userProfile)["_type"] = "record"
+	(*userProfile)["_recordID"] = userID
+	(*userProfile)["_recordType"] = "user"
+	(*userProfile)["_access"] = nil
+	(*userProfile)["_ownerID"] = userID
+	(*userProfile)["_created_at"] = createdAt
+	(*userProfile)["_created_by"] = userID
+	(*userProfile)["_updated_at"] = updatedAt
+	(*userProfile)["_updated_by"] = userID
```
Steven-Chan: The implementation of `CreateUserProfile` and `GetUserProfile` works on the param `userProfile` differently, which lead to an unexpected calling pattern in the signup handler.

`CreateUserProfile` treats the dictionary `userProfile` as `data` in table `user_profile`, while `GetUserProfile` put the whole `user_profile` in the dictionary `userProfile`.

In signup handler, I did not expect that it need to call `GetUserProfile` for the response data. It looks redundant, because even if `CreateUserProfile` is expected to update the input, its output should reflect the result instead of making you need to call `GetUserProfile`.

Take a look at how record gear works with record, it is the responsibility of handler to update those `created_at`, `updated_at` field, I think it is the same case for user profile here.

pkg/auth/dependency/userprofile/store.go
```diff
@@ -0,0 +1,89 @@
...
+func newUserProfileStore(builder db.SQLBuilder, executor db.SQLExecutor, logger *logrus.Entry) *storeImpl {
+	return &storeImpl{
+		sqlBuilder:  builder,
+		sqlExecutor: executor,
+		logger:      logger,
+	}
+}
+
+func (u storeImpl) CreateUserProfile(userID string, userProfile map[string]interface{}) (err error) {
+	now := time.Now().UTC()
```
rickmak: The inline use of time.Now().UTC is not testable.

After the discussion on, in skygear project we adopted this: https://github.com/SkygearIO/skygear-server/pull/518/files

pkg/auth/dependency/userprofile/userprofile.go
```diff
@@ -0,0 +1,6 @@
+package userprofile
+
+type Store interface {
+	CreateUserProfile(userID string, userProfile map[string]interface{}) error
+	GetUserProfile(userID string, userProfile *map[string]interface{}) error
```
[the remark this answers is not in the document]
kaojohnny: @Steven-Chan Just do some searches, find `map` will not be passed by reference, so I think we should keep `GetUserProfile(userID string, userProfile *map[string]interface{}) error`

Please refer:
https://dave.cheney.net/2017/04/29/there-is-no-pass-by-reference-in-go
https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it

Feel free to correct him :)

kaojohnny:
@Steven-Chan I made some modifications of the interface:

```
type Data map[string]interface{}
type UserProfile map[string]interface{}

CreateUserProfile(userID string, data Data) (UserProfile, error)
GetUserProfile(userID string) (UserProfile, error)
```

Hope it can resolve your concerns about `userProfile` in a different pattern in the previous version.

Though I prefer not putting `created_at`, `updated_at` related login into handler, I think it is `userprofile.Store`'s responsibility to generate corresponding data, and handler needn't know implementation details of the `userprofile.Store`. 

In my mind, we will have two `userprofile.Store` finally:

- `store`: a simple version record based user profile store
- `recordStore`: a real record based user profile store, it sends a record save request to the record gear.

So from handler's perspective, it shouldn't care which store is using, both stores are treated equally and needn't special handling for anyone of them.

Steven-Chan:
I think the updated interface looks better.

The direction is a bit different than my very original thought, which the auth gear would just save the data in a json field and no meta field.

For me, I think the implementation in this PR is a bit oversimplifying, but I think `So from handler's perspective, it shouldn't care which store is using, both stores are treated equally and needn't special handling for anyone of them` is a good direction.

So see @rickmak and @carmenlau has any comments?

rickmak:
The definition of `UserProfile` may need to be at `core` package?

But I think it fine to merge this PR first.

PR #749, opened 2018-11-01
safeXXXStore should invoke impl implementation

kaojohnny:
connects #682


Steven-Chan:
I added a commit 587ce6995e1df23873ab5fb578083668e1ac1e5b.

PR #751, opened 2018-11-01
next: add record save handler test case

Steven-Chan:
connect #708


rickmak:
I think we should write other test as JSON in/out as this one. How do you think @carmenlau ?

carmenlau:
Yes, using this format is good. Easier to read for the result comparison. I created an issue for updating auth gear test https://github.com/SkygearIO/skygear-server/issues/758

carmenlau:
Merge via 0669f06

Issue #746, opened 2018-10-31
Asset download link not URL encoded when using FS for asset store


jasonho215:
Step to reproduce
---
1. Create a skygear server with file system as storage
2. upload an asset named `Screenshot 2018-10-30 at 5.45.23 PM.png`
3. sign the uploaded asset, got a link `https://{FQDN}/files/{UUID}-Screenshot 2018-10-30 at 5.45.23 PM.png?expiredAt=1540959007&signature=XIN3ob3MZBLdmZ0CoLuh_AmXgc25McM_0arVoIpXNPY=`

The signed URL does not conform to protocol [RFC1378](https://tools.ietf.org/html/rfc1738)

Expected result
---
1. URL encoded file name (e.g. `abc;touch hello.jpg` -> `abc%3Btouch%20hello.jpg`)
2. Signed URL to access asset stored in file system and the URL conforms to RFC1378

Related code
---
https://github.com/SkygearIO/skygear-server/blob/2760036bb867b342807e4db8ab9d3ea358e9c971/pkg/server/handler/asset.go#L94

PR #756, opened 2018-11-05
Fix fs asset filename encoding issue

Steven-Chan:
connect #746

Tested with S3 and it does not have the same problem.


carmenlau:
Merged to v1-branch via 860cef5

carmenlau:
Merged to master via 9459ba2

Issue #755, opened 2018-11-05
next: user profile record store


kaojohnny:

Steven-Chan:
Side note:
We have decided to let auth gear to import record gear code directly.

This would imply that, auth gear would access record gear database directly, **migration for record gear must also be run** if auth gear use record gear to store user profile.

PR #757, opened 2018-11-05
next: add skygear url to tenant configuration

kaojohnny:
connects #755 

1. add `SKYGEAR_URL` to `TenantConfiguration`, this is used in auth gear to generate a request to record gear.
2. add `record_enabled` to `Plan` table, it helps gateway to know if it routes record request of an app to record gear.

### DIFF
--- a/pkg/core/config/tenant.go
+++ b/pkg/core/config/tenant.go
@@ -26,7 +26,8 @@ type TokenStoreConfiguration struct {
 }
 
 type UserProfileConfiguration struct {
-	ImplName string `msg:"IMPLEMENTATION" envconfig:"USER_PROFILE_IMPL_NAME" json:"IMPLEMENTATION"`
+	ImplName     string `msg:"IMPLEMENTATION" envconfig:"USER_PROFILE_IMPL_NAME" json:"IMPLEMENTATION"`
+	ImplStoreURL string `msg:"IMPL_STORE_URL" envconfig:"USER_PROFILE_IMPL_STORE_URL" json:"IMPL_STORE_URL"`
 }
 
 type UserAuditConfiguration struct {
--- a/pkg/core/config/tenant_gen.go
+++ b/pkg/core/config/tenant_gen.go
@@ -97,6 +97,11 @@ func (z *TenantConfiguration) DecodeMsg(dc *msgp.Reader) (err error) {
 					if err != nil {
 						return
 					}
+				case "IMPL_STORE_URL":
+					z.UserProfile.ImplStoreURL, err = dc.ReadString()
+					if err != nil {
+						return
+					}
 				default:
 					err = dc.Skip()
 					if err != nil {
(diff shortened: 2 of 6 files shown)


Steven-Chan:
Also add a configuration for auth gear to know which user profile implementation chosen?

kaojohnny:
@Steven-Chan We already have 

```
userProfileStore := tConfig.UserProfile.ImplName
```

if `userProfileStore == "record"`, then auth gear will use record gear as user profile store. Or do you consider another configuration?

Steven-Chan:
omg, me so forgetful 😢.

Steven-Chan:
Configuration are already grouped by feature, e.g. `ASSET_STORE_XXX`, `TOKEN_STORE_XXX`.

First of all, I think the url added by this PR should be under `USER_PROFILE_XXX` instead of a generic `SKYGEAR_URL`.
Second, to @rickmak, I think grouping by feature is already enough.

kaojohnny:
@Steven-Chan I think `SKYGEAR_URL=http://localhost:3001` is a general configuration for all gears, in current use case (save user profile to record gear), it is used when `auth` gear sends a request to `record` gear, something like:

```
resp, err := goreq.Request{
	Method: "POST",
	Uri:    u.skygearURL + "record/save",
	Body:   body,
}.Do()
```

It is more like a base configuration for communication between gears. Whereas, `USER_PROFILE_RECORD_STORE_URL=http://localhost:3001/record/` is also good to me.

Steven-Chan:
Actually, I prefer `USER_PROFILE_SKYGEAR_URL` over `USER_PROFILE_RECORD_STORE_URL`.

pkg/core/config/tenant.go
```diff
@@ -10,14 +10,15 @@ import (
...
-	UserAudit       UserAuditConfiguration   `json:"USER_AUDIT" msg:"USER_AUDIT"`
+	DBConnectionStr           string                   `msg:"DATABASE_URL" envconfig:"DATABASE_URL" json:"DATABASE_URL"`
+	APIKey                    string                   `msg:"API_KEY" envconfig:"API_KEY" json:"API_KEY"`
+	MasterKey                 string                   `msg:"MASTER_KEY" envconfig:"MASTER_KEY" json:"MASTER_KEY"`
+	AppName                   string                   `msg:"APP_NAME" envconfig:"APP_NAME" json:"APP_NAME"`
+	CORSHost                  string                   `msg:"CORS_HOST" envconfig:"CORS_HOST" default:"*"`
+	TokenStore                TokenStoreConfiguration  `json:"TOKEN_STORE" msg:"TOKEN_STORE"`
+	UserProfile               UserProfileConfiguration `json:"USER_PROFILE" msg:"USER_PROFILE"`
+	UserAudit                 UserAuditConfiguration   `json:"USER_AUDIT" msg:"USER_AUDIT"`
+	UserProfileRecordStoreURL string                   `msg:"USER_PROFILE_RECORD_STORE_URL" envconfig:"USER_PROFILE_RECORD_STORE_URL" json:"USER_PROFILE_RECORD_STORE_URL"`
```
Steven-Chan: This config should be under `UserProfileConfiguration`.

For config that have more than one implementation, I think you can take skygear v1 asset store as reference.

https://github.com/SkygearIO/skygear-server/blob/master/pkg/server/skyconfig/config.go#L152
https://github.com/SkygearIO/skygear-server/blob/master/pkg/server/skyconfig/config.go#L441

---

`UserProfileSkygearURL` or `UserProfileRecordStoreURL`? @carmenlau @rickmak any comment?

kaojohnny: @Steven-Chan I updated it to 

```
type UserProfileConfiguration struct {
	ImplName     string `msg:"IMPLEMENTATION" envconfig:"USER_PROFILE_IMPL_NAME" json:"IMPLEMENTATION"`
	ImplStoreURL string `msg:"IMPL_STORE_URL" envconfig:"USER_PROFILE_IMPL_STORE_URL" json:"IMPL_STORE_URL"`
}
```

let me know if it is a proper name. Thanks.

PR #759, opened 2018-11-05
next: record delete handler

Steven-Chan:
connect #730


kaojohnny:
@Steven-Chan Please help resolve conflicts, thanks.

PR #766, opened 2018-11-07
next: save user profile to record gear

kaojohnny:
connects #755

## note ##
1. Based on #757
2. For record to record communication, handler should commit transaction before the request initiate.

### DIFF
--- a/pkg/auth/dependency/userprofile/tx_provider.go
+++ b/pkg/auth/dependency/userprofile/tx_provider.go
@@ -10,6 +10,7 @@ type safeUserProfileImpl struct {
 	txContext db.SafeTxContext
 }
 
+// NewSafeProvider returns a auth gear user profile store implementation
 func NewSafeProvider(
 	builder db.SQLBuilder,
 	executor db.SQLExecutor,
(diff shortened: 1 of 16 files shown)


pkg/auth/handler/signup.go
```diff
@@ -223,20 +262,65 @@ func (h SignupHandler) verifyPayload(payload SignupRequestPayload) (err error) {
...
+	if err = h.AuthInfoStore.DeleteAuth(authInfoID); err != nil {
+		return
+	}
+
+	if err = h.TxContext.CommitTx(); err != nil {
+		return
+	}
+
+	if err != nil {
+		err = h.TxContext.RollbackTx()
```
Steven-Chan: Should rollback the transaction for early return if error exists.

pkg/auth/dependency/userprofile/tx_provider.go
```diff
@@ -22,12 +23,17 @@ func NewSafeProvider(
...
 	return s.impl.CreateUserProfile(userID, data)
 }
 
-func (s *safeUserProfileImpl) GetUserProfile(userID string) (profile UserProfile, err error) {
+func (s *safeUserProfileImpl) GetUserProfile(userID string, accessToken string) (profile UserProfile, err error) {
 	s.txContext.EnsureTx()
 	return s.impl.GetUserProfile(userID)
 }
+
+func (s *safeUserProfileImpl) WithInTx() (withIn bool) {
```
Steven-Chan: I thought `WithInTx` was saying if the store is now in transaction.

I suggest to use a function name that describe the store characteristics, something like `IsOperationRunInDB` or `CanWithInTx`.

kaojohnny:
@Steven-Chan Thanks. Updated.

kaojohnny:
Auth gear will use record gear API to create records directly (avoid gear to gear request).

PR #769, opened 2018-11-08
next: field acl get and update handler

Steven-Chan:
connect #738
connect #737

### DIFF
--- a/pkg/record/handler/field_access_update.go
+++ b/pkg/record/handler/field_access_update.go
@@ -53,7 +53,7 @@ func (p FieldAccessUpdateRequestPayload) Validate() error {
 }
 
 /*
-FieldAccessUpdateHandler fetches the entire Field ACL settings.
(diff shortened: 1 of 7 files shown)


pkg/record/handler/field_access_update.go
```diff
@@ -0,0 +1,116 @@
...
+	RawAccess []map[string]interface{} `json:"access"`
+	FieldACL  record.FieldACL
+}
+
+func (p FieldAccessUpdateRequestPayload) Validate() error {
+	return nil
+}
+
+/*
+FieldAccessUpdateHandler fetches the entire Field ACL settings.
```
kaojohnny: should "updates" not fetch.

Steven-Chan: 😱

Steven-Chan: updated

Issue #744, opened 2018-10-31
Completely remove `type/` support in next


Steven-Chan:
A followup of  #605.

PR #770, opened 2018-11-08
next: remove old format record id

Steven-Chan:
connect #744

### DIFF
--- a/pkg/record/handler/save.go
+++ b/pkg/record/handler/save.go
@@ -84,7 +84,8 @@ curl -X POST -H "Content-Type: application/json" \
   -d @- http://localhost:3000/save <<EOF
 {
     "records": [{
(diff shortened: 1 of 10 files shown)


pkg/record/handler/save.go
```diff
@@ -101,12 +102,14 @@ curl -X POST -H "Content-Type: application/json" \
   "records": [
     {
       "collection": {
-        "$type": "ref",
-        "$id": "collection/10"
+				"$type": "ref",
+				"$recordType": "collection",
+        "$recordID": "10"
```
carmenlau: Please update the curl example with the same indentation.

Steven-Chan:
@carmenlau updated

carmenlau:
💯

PR #792, opened 2018-11-12
next: Send dummy welcome email when signup and test api

Steven-Chan:
connect #779 

I will open another PR for reading content from template.

Also, I didn't figure out how to write test for go routine in http handler yet.

---

For those who want to test this PR, remember to set smtp settings and welcome email settings.
You may create a testing smtp server on https://mailtrap.io/.


pkg/auth/handler/welcome_email.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/skygeario/skygear-server/pkg/core/server"
+	"github.com/skygeario/skygear-server/pkg/server/skyerr"
+)
+
+// AttachWelcomeEmailHandler attaches WelcomeEmailHandler to server
+func AttachWelcomeEmailHandler(
+	server *server.Server,
+	authDependency auth.DependencyMap,
+) *server.Server {
+	server.Handle("/welcome_email/test", &WelcomeEmailHandlerFactory{
```
kaojohnny: May I know when will us use this handler? Or it is for test only.

Steven-Chan: This is for portal user to test before they confirm and save the setting.

Steven-Chan: https://github.com/SkygearIO/forgot_password/blob/master/forgot_password/handlers/welcome_email.py#L126

kaojohnny: Thanks.

carmenlau:
@Steven-Chan The PR looks good. For testing go routine part, how about add a channel field in handler and we will only assign channel when constructing handler for test?

Steven-Chan:
> @Steven-Chan The PR looks good. For testing go routine part, how about add a channel field in handler and we will only assign channel when constructing handler for test?

I will open another PR for the test case.

carmenlau:
Sure, thanks @kaojohnny

PR #793, opened 2018-11-13
Use record gear to save profile

kaojohnny:
connects #755

### DIFF
--- a/pkg/auth/dependency/userprofile/recordstore.go
+++ b/pkg/auth/dependency/userprofile/recordstore.go
@@ -1,10 +1,8 @@
 package userprofile
 
 import (
-	"encoding/json"
 	"time"
 
-	"github.com/franela/goreq"
 	"github.com/sirupsen/logrus"
 	"github.com/skygeario/skygear-server/pkg/core/asset"
 	"github.com/skygeario/skygear-server/pkg/core/auth"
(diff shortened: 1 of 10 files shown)


pkg/auth/dependency/userprofile/recordstore.go
```diff
@@ -0,0 +1,159 @@
+package userprofile
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/franela/goreq"
```
Steven-Chan: why still need `goreq` 🤔

Steven-Chan:
@kaojohnny is this PR ready to review? I find the `getUserProfile` is calling record gear api instead of importing record code.

Other than this I would merge the PR. I would open an issue about refactoring the record store. Because the UserProfileStore is now importing code from handler, which looks very bad.

Steven-Chan:
@rickmak @carmenlau any comment?

kaojohnny:
@Steven-Chan I made a update of using record gear API to query user profile record. Please help review. Thanks!

rickmak:
Can someone write down the discussion point? Seem @Steven-Chan don't provide any argument here but @kaojohnny does.

kaojohnny:
@rickmak After some consideration, I think that use two different invoking styles (importing record gear API and go through public API) may confuse other developers, so I decide to import record gear API to both endpoints to avoid confusion.

Issue #786, opened 2018-11-09
next: SSO get provider login url handler


Steven-Chan:
In SSO plugin, `sso/{provider name}/login_auth_url`

PR #797, opened 2018-11-15
next: SSO get provider login url

kaojohnny:
connects #786


pkg/core/config/tenant.go
```diff
@@ -35,6 +36,14 @@ type UserAuditConfiguration struct {
 	TrailHandlerURL string `msg:"TRAIL_HANDLER_URL" envconfig:"USER_AUDIT_TRAIL_HANDLER_URL" json:"TRAIL_HANDLER_URL"`
 }
 
+type SSOConfiguration struct {
+	Name         string `msg:"NAME" envconfig:"SSO_CONFIG_NAME" json:"NAME"`
+	Enabled      bool   `msg:"ENABLED" envconfig:"SSO_CONFIG_ENABLED" json:"ENABLED"`
+	ClientID     string `msg:"CLIENT_ID" envconfig:"SSO_CONFIG_CLIENT_ID" json:"CLIENT_ID"`
+	ClientSecret string `msg:"CLIENT_SECRET" envconfig:"SSO_CONFIG_CLIENT_SECRET" json:"CLIENT_SECRET"`
+	Scope        string `msg:"SCOPE" envconfig:"SSO_CONFIG_SCOPE" json:"SCOPE"`
```
carmenlau: The overall direction of PR looks fine. But for the sso config, can we provide different config based on providers by environment variable? It would be better if we can have .env.example for auth gear too, so that we can set this up easily for development.

pkg/core/config/tenant.go
```diff
@@ -132,5 +132,27 @@ func SetTenantConfig(i interface{}, t TenantConfiguration) {
 func NewTenantConfigurationFromEnv(_ *http.Request) (TenantConfiguration, error) {
 	c := TenantConfiguration{}
 	err := envconfig.Process("", &c)
+	c.SSOConfigs = appendSSOConfigs()
```
kaojohnny: @carmenlau I add a function which reads env SSO provider configs to tenant config, please help review. Thanks.

pkg/auth/handler/login_auth_url.go
```diff
@@ -0,0 +1,151 @@
...
+//
+// {
+//     "result": "<auth_url>"
+// }
+type LoginAuthURLHandler struct {
+	TxContext         db.TxContext           `dependency:"TxContext"`
+	GoogleProvider    sso.Provider           `dependency:"GoogleSSOProvider,optional"`
+	FacebookProvider  sso.Provider           `dependency:"FacebookSSOProvider,optional"`
+	InstagramProvider sso.Provider           `dependency:"InstagramSSOProvider,optional"`
+	LinkedInProvider  sso.Provider           `dependency:"LinkedInSSOProvider,optional"`
```
carmenlau: Currently we will inject all the providers to the handler, that makes both the `DependencyMap` and all SSO handlers need to know the list of providers. When we want to add 1 new provider, we will need to update many places. Maybe we can inject one provider only by checking the route in `DependencyMap.Provide`? Handlers don't need to know about what providers we have.

pkg/core/config/tenant.go
```diff
@@ -105,5 +132,33 @@ func SetTenantConfig(i interface{}, t TenantConfiguration) {
...
+	return
+}
+
+func appendSSOConfigs() []SSOConfiguration {
+	configs := make([]SSOConfiguration, 0)
+	SSOProviderNames := []string{
+		"google",
+		"facebook",
+		"instagram",
+		"linkedin",
```
carmenlau: Maybe we can put the provider lists in a centralised place? It would be easier for us to add new provide in future.

pkg/core/config/tenant.go
carmenlau: Looks good:)

kaojohnny:
@carmenlau Thanks for your review. Updated.

pkg/auth/handler/login_auth_url.go
```diff
@@ -38,13 +39,8 @@ func (f LoginAuthURLHandlerFactory) NewHandler(request *http.Request) http.Handl
...
 	h.ProviderName = vars["provider"]
-	providers := map[string]sso.Provider{
-		"google":    h.GoogleProvider,
-		"facebook":  h.FacebookProvider,
-		"instagram": h.InstagramProvider,
-		"linkedin":  h.LinkedInProvider,
-	}
-	if provider, ok := providers[h.ProviderName]; ok {
+	providerDependencyName := strings.Title(h.ProviderName) + "SSOProvider"
+	if provider, ok := f.Dependency.Provide(providerDependencyName, request).(sso.Provider); ok {
```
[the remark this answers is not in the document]
kaojohnny: @carmenlau Updated.

pkg/auth/dependency/sso/authurl.go
```diff
@@ -0,0 +1,37 @@
...
+		return "", err
+	}
+	v := url.Values{}
+	v.Set("response_type", "code")
+	v.Add("client_id", params.clientID)
+	v.Add("redirect_uri", RedirectURI(params.urlPrefix, params.prividerName))
+	v.Add("state", encodedState)
+	v.Add("scope", strings.Join(params.scope, " "))
+	v.Add("access_type", "offline")
+	v.Add("prompt", "select_account")
```
carmenlau: I tried generate google auth url and found `access_type` and `prompt` params are repeated.

1. `access_type` and `prompt` should be for google only
1.  if any of the query params are provided in `params.options` we should use them, here should provide the default value only.

pkg/auth/handler/login_auth_url_test.go
```diff
@@ -0,0 +1,172 @@
...
+		Convey("should convert unknown string to sso.Undefined", func() {
+			modeFromStr := sso.UXModeFromString("some_string")
+			So(modeFromStr, ShouldEqual, sso.Undefined)
+		})
+	})
+}
+
+func TestLoginAuthURLHandler(t *testing.T) {
+	Convey("Test TestLoginAuthURLHandler", t, func() {
+		h := &LoginAuthURLHandler{}
```
carmenlau: For the test case format, we had discussion before and thought JSON in/out format is clearer.

Our discussion: https://github.com/SkygearIO/skygear-server/pull/751

I will merge this PR after update `authURL` function, we can update this test later but please use JSON in/out format in the future PR.

Thanks @kaojohnny

pkg/auth/dependency/sso/authurl.go
kaojohnny: @carmenlau `access_type` and `prompt` have added in `google.go` already, I accidentally added two lines of code again in `authurl.go`. Thanks for spotting this issue.

pkg/auth/dependency/sso/utils.go
```diff
@@ -0,0 +1,66 @@
...
+	BaseURLs = map[string]string{
+		"google":    "https://accounts.google.com/o/oauth2/v2/auth",
+		"facebook":  "https://www.facebook.com/dialog/oauth",
+		"instagram": "https://api.instagram.com/oauth/authorize",
+		"linkedin":  "https://www.linkedin.com/oauth/v2/authorization",
+	}
+)
+
+// CustomCliams is the type for jwt encoded
+type CustomCliams struct {
```
Steven-Chan: `CustomCliams` -> `CustomClaims`

pkg/core/config/tenant.go
```diff
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	"github.com/kelseyhightower/envconfig"
+	"github.com/skygeario/skygear-server/pkg/auth/dependency/sso" // tolerant nextimportslint: auth
```
Steven-Chan: 😱 😨 😱

Steven-Chan: I think we can avoid importing gear package in this case, although the config parsing may be different.

cc @carmenlau

Steven-Chan: A short discussion with @carmenlau.

We suggest to change the config format like this:

```
SSO_PROVIDERS=google,facebook,skygear

SSO_GOOGLE_CLIENT_ID=my_google_app_id
SSO_GOOGLE_CLIENT_SECRET=my_google_app_secret
SSO_GOOGLE_SCOPE=https://www.googleapis.com/auth/userinfo.email

SSO_FACEBOOK_CLIENT_ID=my_facebook_app_id
SSO_FACEBOOK_CLIENT_SECRET=my_facebook_app_secret
SSO_FACEBOOK_SCOPE=email,user_friends

SSO_SKYGEAR_CLIENT_ID=my_facebook_app_id
SSO_SKYGEAR_CLIENT_SECRET=my_facebook_app_secret
SSO_SKYGEAR_SCOPE=email,user_friends
```

Obviously, skygear is not an auth provider, but the config parsing would not throw error on this, instead the auth package sso struct would throw when it does not recognise the provider name.

.env.example
```diff
@@ -159,6 +159,7 @@ CUSTOM_TOKEN_SECRET="ssosecret"
 ########################################################################
 # SSO provider configurations
 ########################################################################
+#SSO_PROVIDERS=google,facebook,instagram,linkedin
```
Steven-Chan: `SSO_xxx_ENABLED` can be removed, because when it presents in `SSO_PROVIDERS`, its enabled.

kaojohnny: Updated. Thanks.

pkg/auth/handler/login_auth_url.go
```diff
@@ -0,0 +1,136 @@
...
+// }
+// EOF
+//
+// {
+//     "result": "<auth_url>"
+// }
+type LoginAuthURLHandler struct {
+	TxContext    db.TxContext           `dependency:"TxContext"`
+	AuthContext  coreAuth.ContextGetter `dependency:"AuthContextGetter"`
+	Provider     sso.Provider           `dependency:"SSOProvider,optional"`
```
Steven-Chan: `Provider` is not actually `optional` here, because throwing error is not a normal flow, otherwise any dependency can be optional and the handlers would be polluted with error handling.

I think the error for unrecognised provider should be thrown at `inject.go`.

kaojohnny: @Steven-Chan Updated. Thanks.

Issue #795, opened 2018-11-15
next: Fix signup auth record keys checking


Steven-Chan:
Current impl:

- There is a unique constraint on `_auth_provider_password.auth_data`.
- Default auth record keys setting is `username,email`.

Behaviour:

1. Signup with `{"username": "user1", "email": "user1@oursky.com"}` -> success
1. Signup with `{"username": "user2", "email": "user1@oursky.com"}` -> in v1, this should fail, but now success

According to the original design of auth record keys, each key or key combination (e.g. `a,b,(c,d,e)`) should be unique in the table.

We need to update auth data handling to match the correct auth record keys behaviour.

~~Since we don't have JIT db migration now, we can't change constraint when settings get update. We have a few approaches:~~
- ~~When signup, check unique constraint matched auth record keys setting~~
- ~~When signup, query with provided auth data, and check if there is a user existed already~~
- ~~When login, query with provided auth data, and check if more than one user found~~

Steven-Chan:
I've come up with a more sensible solution, which does not require setting constraints based on the auth record keys setting.

The auth data json should respect the auth record keys setting. E.g. the default setting `username,email`, it means user should either use username or email to login.

If, for the simplest case, a user signup with username only, there should be one signup entry:

- `auth_data`: `{username: xxxx}` + `password`: `xxxx`

If a user signup with **both** username and email, there should be two signup entries:

- `auth_data`: `{username: xxxx}` + `password`: `xxxx`
- `auth_data`: `{email: xxxx}` + `password`: `xxxx`

In next, there should be two entries of principal.

If the auth record keys settings is like `username,(email,nickname)` and a user signup with data with all three fields, the backend should save:

- `auth_data`: `{username: xxxx}` + `password`: `xxxx`
- `auth_data`: `{email: xxxx, nickname: xxxx}` + `password`: `xxxx`

This above changes implies that:

- we need to remove the constraint that each user can only have one principal of one provider
- when changing password for one user, all the password principals for that user need to be updated 
- any underlying function that needs to get principals of one user, it must return an array of principal instead of a single principal

tensiuyan:
https://github.com/SkygearIO/features/blob/master/features/48-user-profile-discovery/user-profile.md#authentication-record-configurations

Steven-Chan:
Try translating the config to a logic may help understand the behaviour.

## Case 1
[...]

PR #803, opened 2018-11-21
next: Fix auth record checking

kaojohnny:
connects #795


pkg/auth/inject.go
```diff
@@ -42,9 +42,13 @@ func (m DependencyMap) Provide(dependencyName string, r *http.Request) interface
 	case "AuthInfoStore":
 		tConfig := config.GetTenantConfig(r)
 		return coreAuth.NewDefaultAuthInfoStore(r.Context(), tConfig)
+	case "AuthDataKeys":
```
Steven-Chan: I think directly injecting the config into the handler is violating the current injection pattern.

I can think of two approaches:
- put the `ToValidAuthDataList` function in `AuthDataChecker` (and maybe give it a different name), OR
- merge `AuthDataChecker` into `PasswordAuthProvider` and put all auth record keys logic in it, because imo the handler does not even have to know how to process the given auth data

pkg/auth/dependency/provider/password/authdata.go
```diff
@@ -0,0 +1,100 @@
...
+// - if authData is { "username": "john.doe", "email": "john.doe@example.com" }, output is [{ "username": "john.doe" }, { "email": "john.doe@example.com" }]
+//
+// example 2: authRecordKeys = [["username", "nickname"], ["email"]]
+// - if authData is { "username": "john.doe", "email": "john.doe@example.com", "nickname": "john.doe" },
+// output is [{ "username": "john.doe", "nickname": "john.doe" }, { "email": "john.doe@example.com" }]
+//
+// example 3: authRecordKeys = [["username", "email"], ["nickname"]]
+// - if authData is { "username": "john.doe", "nickname": "john.doe" },
+// output is [{ "nickname": "john.doe" }}]
+func toValidAuthDataList(authRecordKeys [][]string, authData map[string]interface{}) []map[string]interface{} {
```
Steven-Chan: This function looks very similar to `isValid`, I suggest you to put this in `defaultAuthDataChecker` and replace the implementation of `isValid` with calling this function.

PR #806, opened 2018-11-22
next: forgot password handler

Steven-Chan:
connect #777


pkg/auth/inject.go
```diff
@@ -121,6 +122,12 @@ func (m DependencyMap) Provide(dependencyName string, r *http.Request) interface
 			panic(err)
 		}
 		return trail
+	case "ForgotPasswordEmailSender":
+		tConfig := config.GetTenantConfig(r)
+		return forgotpwdemail.NewDefaultSender(tConfig, mail.NewDialer(tConfig.SMTP))
+	case "TestForgotPasswordEmailSender":
```
kaojohnny: Will it change like PR #805's "WelcomeEmailSendTask" to support async sender?

Steven-Chan: Hmm, the original python plugin sends it synchronously, so I think I will just keep it like this.

kaojohnny:
@Steven-Chan Please help resolve conflicts. Thanks.

Steven-Chan:
@kaojohnny rebased, I've changed `GetPrincipalByEmail` to `GetPrincipalsByEmail` which return multiple entries.

pkg/auth/handler/forgot_password.go
```diff
@@ -0,0 +1,235 @@
...
+			err = skyerr.NewError(skyerr.ResourceNotFound, "user not found")
+			return
+		}
+		// TODO: more error handling here if necessary
+		err = skyerr.NewResourceFetchFailureErr("auth_data", authData)
+		return
+	}
+
+	userID := principals[0].UserID
+	hashedPassword := principals[0].HashedPassword
```
kaojohnny: `principals` could be an empty list, need to check its length before use it.

Steven-Chan:
@kaojohnny updated

carmenlau:
After review this PR, we found that there is a design problem of this api. `GetPrincipalsByEmail` will return multiple principals only if the auth_key is a combined keys (e.g. `(email,user_type)`). If we only send email to the first one by `principals[0]`, user will not be able to reset the password of second account. This problem exists in the forgot password plugin.

After discussed with steven. To solve this, this api should send multiple emails if more than 1 accounts with the same email. Each email has different reset link, and we can identify the user record from the link. Developer can use the email template to let the user know which account is this email referring to.

Steven-Chan:
@carmenlau @kaojohnny updated

Steven-Chan:
@carmenlau updated with config default value.

Steven-Chan:
@carmenlau updated with secure match, and i've moved the handling of principals not found in provider.

Issue #783, opened 2018-11-09
next: SSO get config handler


Steven-Chan:
In SSO plugin, `sso/config`

kaojohnny:
In SSO plugin, `sso/config` is defined as

```python
def register_config_lambda(allowed_callback_urls):
    @skygear.op('sso/config')
    def sso_config_lambda(**kwargs):
        return {
            'authorized_urls': allowed_callback_urls
        }
```

Should we change it to `sso/allowed_callback_urls`?

carmenlau:
We used a generic name `sso/config` for this endpoint, it is because we want to give room for extension in future. We can put more information into this api, but currently the only config that we have are `allowed_callback_urls`.

kaojohnny:
@carmenlau @Steven-Chan Note that SSO handlers response is in following format: `{'result': '...'}` to align other auth handlers. Not `{'authorized_urls': '...'}` or `{'auth_url': '....'}`.

PR #813, opened 2018-11-27
next: SSO get config

kaojohnny:
connects #783

### DIFF
--- a/pkg/auth/handler/ssohandler/config.go
+++ b/pkg/auth/handler/ssohandler/config.go
@@ -1,4 +1,4 @@
-package handler
+package ssohandler
 
 import (
 	"net/http"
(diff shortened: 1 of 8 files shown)


pkg/auth/handler/ssohandler/config.go
```diff
@@ -0,0 +1,57 @@
...
+// <<EOF
+// {
+// }
+// EOF
+//
+// {
+//     "result": [
+//         "http://localhost",
+//         "http://127.0.0.1"
+//     ]
```
carmenlau: IMO

1. If we want this api extensible in future, we can use `/sso/config` a more generic name. But we may want to put the result in a dictionary. So that we can add new key with other new information, and keep the api compatible with the old version SDK, this is important. The payload may looks like this.
    ```
    {
        "result": {
            "authorized_urls": [
                "http://localhost",
                "http://127.0.0.1"
            ]
        }
    }
    ```

1.  If we plan to keep this api simple and will not extent it in the future. For any new feature that the SDK need to get config from server, we will add new api. We should rename this api to a more specific name like what you mentioned `/sso/authorized_urls`. And the payload can be like this.

    ```
    {
        "result": [
            "http://localhost",
            "http://127.0.0.1"
        ]
    }
    ```

What do you think?

kaojohnny: @carmenlau I updated response structure, I consider `/sso/config` should have a more generic response. Please help review. Thanks.

Issue #787, opened 2018-11-09
next: SSO get provider link url handler


Steven-Chan:
In SSO plugin, `sso/{provide name}/link_auth_url`

PR #815, opened 2018-11-28
next: SSO get provider link url

kaojohnny:
connects #787
requires #783


kaojohnny:
@Steven-Chan @carmenlau conflicts resolved.

PR #823, opened 2018-12-03
next: signup auto verify (preview of async task)

Steven-Chan:
require #814 

Please don't merge this PR yet. 

This PR is for @rickmak to review the new async task design.

In short,
- task are registered in gear `main.go`
- task logic are put in `pkg/{gear}/task/{task_name}.go`
- handler inject ~~`TaskClient`~~ `TaskQueue` to submit task by name and param


rickmak:
Do you have any class diagram for overview?

Steven-Chan:
The `pkg/core/async` contains all the parts for async task. I try to follow other job queue system the when designing the task server and client structure.

### `pkg/core/async`

- `Task` interface
- `TaskContext` provides the context of a task, which carries the information of the request that spawn the task

### `pkg/core/async/server`

- `TaskServer`
  - `func Register` is meant to be called at server **boot time** for registering tasks
  - `func Handle` would be called whenever the server receive a task request, which is submitted by `TaskClient` at **runtime**

The current implementation of `TaskServer` is a very basic one, which spawn a new go routine each time `Handle` is called.

For different dedicated behaviours, like long running go routine, limited concurrent tasks and any queueing logic, are meant to be implemented in the `TaskServer`.

### `pkg/core/async/client`

- `TaskClient`
  - `func Submit` is for submitting task by name and passing param

`TaskClient` is meant to be **injected in handler** like other dependencies.

cmd/auth/main.go
```diff
@@ -39,7 +41,15 @@ func main() {
 	// logging initialization
 	logging.SetModule("auth")
 
-	authDependency := auth.NewDependencyMap()
+	authDependency := auth.DependencyMap{}
+
+	asyncTaskServer := asyncServer.NewTaskServer()
+	task.AttachVerifyCodeSendTask(asyncTaskServer, authDependency)
```
rickmak: The `asyncTaskServer` configuration should be pass in as parameter at `NewTaskServer`.

I am against to use dependency injection here. The objective the the dependency injection is to fulfil the requirement of Handler factory. Making the dependency inject have extra responsibility introduce complexity without proportional reward. My judgement is the `TaskServer` need to KNOW what task to run and what dependency it involves.

As your comment said, we try to `So when we later switch to a real job queue system, the interfaces should be very similar.`

So a job queue system commonly compose `queue`, `worker` and `task`.
1. There can be more than one queue in a system
2. Not all worker can work on all task.
3. Task should be serializable.

Point 1 and 2 are design dependency on needs. Some important factor is the does all task are in same run time, i.e. all run in python or some python or js.
Commonly, one queue with one kid of worker already solved 90% of the problem I have experienced during web development.

So here, you are mixing up queue and worker in golang, which can be fine as it does not related to the interfacing of how a task is trigger.

However, assuming there is one worker and it will have ability to process all kind of task is too bold.

So the worker/server dependency should be actively maintain by themselves here. Which tsk it can handle should be decided during initialization. Can be DI here, but I don't see the need as compare to handler.

The `AsyncTaskServer` ( I would call it queue ) should be a dependency at DependencyMap, not a special one.

pkg/auth/handler/signup.go
```diff
@@ -245,14 +258,18 @@ func (h SignupHandler) createPrincipal(payload SignupRequestPayload, authInfo au
...
-		default:
-			panic("unexpcted send welcome email request no receiver")
+		h.WelcomeEmailSendTask.Execute(email, userProfile, h.Logger)
+	}
+}
+
+func (h SignupHandler) sendUserVerifyRequest(userProfile userprofile.UserProfile) {
+	for _, key := range h.UserVerifyKeys {
+		if value, ok := userProfile.Data[key].(string); ok {
+			h.TaskClient.Submit(task.VerifyCodeSendTaskName, task.VerifyCodeSendTaskParam{
```
rickmak: `TaskClient.submit` to `TaskQueue.submit`?

cmd/auth/main.go
Steven-Chan:
```
So the worker/server dependency should be actively maintain by themselves here. Which tsk it can handle should be decided during initialization.
```

I don't think that's the case. All the dependency can only be created when the task is enqueued, because the worker is multi-tenant. The tenant configuration (or app id) is part of the task parameter.

This case is similar to how we change the DI from v1 to next, we moved dependency creation to request handling.

Steven-Chan: > So here, you are mixing up queue and worker in golang, which can be fine as it does not related to the interfacing of how a task is trigger.
> 
> However, assuming there is one worker and it will have ability to process all kind of task is too bold.

There is no concept of `queue` and `worker` in my mind when doing this, because in fact they are not, that's just something run in a go routine. The wording `client` and `server` is to identify that the two parts are expected to be two completely separated components and the `server` employs a different life cycle to the original skygear `server`.

But indeed I didn't take the naming to match existing job queue system seriously. If the wording is confusing, I can change it.

rickmak: > All the dependency can only be created when the task is enqueued, because the server (worker) is multi-tenant.

All dependency can be inject or re-create when the task is execute. Holding the dependency while the task is waiting for execution seem not a good idea.

> There is no concept of queue and worker

Designing the interface with those concept will probably make the transition to a real queue system more smooth in future. I hate `server` since we already have a web `server` which will make the reading difficut. In current case, we can call it task `executor`.

Steven-Chan: > All dependency can be inject or re-create when the task is execute.

This is exactly what I've done in the PR.

Steven-Chan:
@rickmak I've renamed the server and client to executor and queue.

Steven-Chan:
@rickmak @carmenlau I've added a commit which should reflect what we've discussed earlier.

Issue #788, opened 2018-11-09
next: SSO handle provider oauth callback handler


Steven-Chan:
In SSO plugin `sso/{provide name}/auth_handler`

PR #828, opened 2018-12-04
next: sso auth handler

kaojohnny:
connects #788 
connects #829

## Note
1. `process_userinfo_hook` is not implemented yet
2. I think `auto_link_keys` is not useful if `process_userinfo_hook` is not implemented, so it's not included in this PR too. Let me know if I have any misunderstandings, I can add it back. (5309dd9)


pkg/auth/handler/ssohandler/auth_handler.go
```diff
@@ -0,0 +1,398 @@
...
+	payload.Code = q.Get("code")
+	payload.Scope = strings.Split(q.Get("scope"), " ")
+	payload.EncodedState = q.Get("state")
+
+	return payload, nil
+}
+
+func (h AuthHandler) Handler() http.Handler {
+	// reference from APIHandlerToHandler
+	return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
```
carmenlau: Since we are not using APIHandler in this handler, we don't need to have a `Handle` function to return `HandlerFunc`. `AuthHandler` can be the handle function directly, see sso `IFrameHandler`.

pkg/auth/handler/ssohandler/auth_handler.go
```diff
@@ -0,0 +1,398 @@
...
+			respErr = skyerr.MakeError(err)
+			return
+		}
+
+		if err := payload.Validate(); err != nil {
+			respErr = skyerr.MakeError(err)
+			return
+		}
+
+		if h.WithTx() {
```
carmenlau: We can open the transaction here directly and remove `WithTx`. This is for `APIHandler` only so we can have a generic setup to decide whether the handler need auto transaction. If it is always true here, we can just remove it, to make the handler simple.

pkg/auth/handler/ssohandler/auth_handler.go
```diff
@@ -0,0 +1,398 @@
...
+			respErr = skyerr.MakeError(err)
+			return
+		}
+		authResp, err := h.getAuthResp(oauthAuthInfo)
+		if h.TxContext != nil {
+			h.TxContext.CommitTx()
+		}
+
+		// handle callback url by ux_mode
+		tConfig := config.GetTenantConfig(r)
```
carmenlau: Handler should not read the tenant config, the design of dependency injection is to handle the multi-tenant logic, so handler doesn't need to have knowledge of multi tenant thing. Maybe we should have a sso dependency that hold the SSO setting and help do the callback urls validation.

kaojohnny:
@carmenlau @Steven-Chan Updated, please help review. Thanks.

pkg/auth/handler/sso/auth_handler.go
```diff
@@ -0,0 +1,456 @@
...
+			}
+			err = skyerr.NewError(skyerr.ResourceNotFound, "User not found")
+			return
+		}
+	}
+	return
+}
+
+func (h AuthHandler) authLinkUser(oauthAuthInfo sso.AuthInfo) (*oauth.Principal, error) {
+	principals, e := h.PasswordAuthProvider.GetPrincipalsByAuthData(oauthAuthInfo.ProviderAuthData)
```
carmenlau: In the original SSO plugin, auto link user is done by the configuration `SSO_AUTO_LINK_PROVIDER_KEYS` dynamically. User can disable this by providing empty config.

Also, the case becomes more complicated in the new password provider models. 
    1. this function should not return multiple principals
    2. we should create password provider entry with empty password to occupy the email... etc

I think it is a bit too much to handle it here in this PR? Maybe we should remove the auto link feature from this PR and handle it in https://github.com/SkygearIO/skygear-server/issues/837

pkg/auth/handler/sso/auth_handler.go
```diff
@@ -0,0 +1,456 @@
...
+			response := handler.APIResponse{
+				Err: respErr,
+			}
+			handler.WriteResponse(rw, response)
+		}
+	}()
+
+	payload, err := h.DecodeRequest(r)
+	if err != nil {
+		respErr = skyerr.MakeError(err)
```
carmenlau: We can convert the error to skygear Error object in the defer, and don't need to repeat `skyerr.MakeError`?

pkg/auth/dependency/sso/utils.go
```diff
@@ -60,7 +99,7 @@ func GetScope(scope Scope, defaultScope Scope) Scope {
 // RedirectURI generates redirect uri from URLPrefix and provider name
 func RedirectURI(URLPrefix string, providerName string) string {
 	u, _ := url.Parse(URLPrefix)
-	path := fmt.Sprintf("/sso/%s/auth_handler", providerName)
+	path := fmt.Sprintf("/auth/sso/%s/auth_handler", providerName)
```
carmenlau: Auth gear should not need to have knowledge of the `/auth` prefix? If developer use the auth gear without auth gear, it should work. When there is the gateway, we should config the `URLPrefix` to be `http://api-domain/auth`? Does it work?

kaojohnny:
@carmenlau Updated. When you have time, please help review. Thanks.

pkg/auth/handler/sso/auth_handler.go
```diff
@@ -0,0 +1,295 @@
...
+func (h AuthHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	// reference from APIHandlerToHandler
+	var err error
+
+	defer func() {
+		if err != nil {
+			response := handler.APIResponse{
+				Err: skyerr.MakeError(err),
+			}
+			handler.WriteResponse(rw, response)
```
carmenlau: Since this api is called by user browser which is not an api call, for all the errors we can show the error message directly like `validateCallbackUR` error? 

Would like to generalize the error handling of this handler, so we don't need special handling for error of `validateCallbackURL`. Thought?

pkg/auth/handler/sso/auth_handler.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	// handle callback url by ux_mode
+	UXMode := oauthAuthInfo.State.UXMode
+	callbackURL := oauthAuthInfo.State.CallbackURL
+	allowedCallbackURLs := h.SSOSetting.AllowedCallbackURLs
+	err = h.validateCallbackURL(allowedCallbackURLs, callbackURL)
+	if err != nil {
+		err = nil
+		http.Error(rw, err.Error(), http.StatusBadRequest)
```
carmenlau: err is clear in line 177

pkg/auth/dependency/sso/utils.go
```diff
@@ -99,7 +99,7 @@ func GetScope(scope Scope, defaultScope Scope) Scope {
 // RedirectURI generates redirect uri from URLPrefix and provider name
 func RedirectURI(URLPrefix string, providerName string) string {
 	u, _ := url.Parse(URLPrefix)
-	path := fmt.Sprintf("/auth/sso/%s/auth_handler", providerName)
+	path := fmt.Sprintf("%s/sso/%s/auth_handler", u.Path, providerName)
```
carmenlau: We need to handle user input with tailing `/`. Both `http://localhost:3000/auth` and `http://localhost:3000/auth/` should be valid.

pkg/auth/handler/sso/sso.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != skydb.ErrUserNotFound {
+		// some other error
+		return resp, err
+	}
+
+	// check if user is already linked
+	userID := oauthAuthInfo.State.UserID // skygear userID
+	_, err = h.OAuthAuthProvider.GetPrincipalByUserID(userID)
+	if err == nil {
+		err = skyerr.NewError(skyerr.InvalidArgument, "provider account already linked with existing user")
```
carmenlau: We should check the oauth principal with both user id and provider name, one user should be enable to connect with both google and facebook account. But they cannot connect with 2 different google accounts.

pkg/auth/dependency/sso/authInfo.go
```diff
@@ -0,0 +1,85 @@
+package sso
+
+type authHandler struct {
```
Steven-Chan: The name `authHandler` looks so general, and I thought it is the base class of different provider, but it is not. I suggest you to refactor this struct.

By the convention of `skygear` as I know, to share common logic, it should be called something like `getAuthInfoRequest` if you put those param (code, scope and encodedState) in the struct body.

Also, it does not look so golang with those function variable, they can be easily replaced by interface.

pkg/auth/handler/sso/auth_handler.go
Steven-Chan: I find there are multiple functions that write things to the `ResponseWriter` which makes the flow hard to follow.

And errors are ignored in `handleSessionResp` and `handleRedirectResp`.

pkg/auth/handler/sso/auth_handler.go
```diff
@@ -0,0 +1,295 @@
...
+	// handle authResp by UXMode
+	switch oauthAuthInfo.State.UXMode {
+	case sso.WebRedirect.String(), sso.WebPopup.String():
+		h.handleSessionResp(rw, r, UXMode, callbackURL, resp)
+	case sso.IOS.String(), sso.Android.String():
+		h.handleRedirectResp(rw, r, UXMode, callbackURL, resp)
+	}
+
+	if h.TxContext != nil {
+		h.TxContext.CommitTx()
```
Steven-Chan: I think it should rollback the transaction if error exists?

pkg/auth/handler/sso/auth_handler.go
```diff
@@ -0,0 +1,295 @@
...
+	if err = payload.Validate(); err != nil {
+		return
+	}
+
+	if err := h.TxContext.BeginTx(); err != nil {
+		panic(err)
+	}
+
+	defer func() {
+		if h.TxContext.HasTx() {
```
Steven-Chan: This should check for `err` instead of whether an  transaction exists.

It seems copied from `apiHandler`, but that is actually not so appropriate.

pkg/auth/handler/sso/auth_handler.go
kaojohnny: Maybe not? in `WriteResponse` it uses `nextSkyerr.ErrorDefaultStatusCode(response.Err)` to generate corresponding http error code from different `skyerr`, if we generalize error here, it could be `http.StatusBadRequest` for all errors.

carmenlau: I think what we need to decide is the format of error response, json or plan text. If we want plan text, we can still use `nextSkyerr.ErrorDefaultStatusCode(response.Err)` to convert the status code. But if we use `WriteResponse` it also convert the response to json response.

kaojohnny:
@carmenlau @Steven-Chan Thanks for your review, I had made some updates. When you have time, please help review again. Thanks.

pkg/auth/handler/sso/auth_handler.go
```diff
@@ -0,0 +1,304 @@
...
+	if err != nil {
+		return
+	}
+
+	// handle authResp by UXMode
+	switch oauthAuthInfo.State.UXMode {
+	case sso.WebRedirect.String(), sso.WebPopup.String():
+		err = h.handleSessionResp(rw, r, UXMode, callbackURL, resp)
+	case sso.IOS.String(), sso.Android.String():
+		err = h.handleRedirectResp(rw, r, UXMode, callbackURL, resp)
```
carmenlau: @kaojohnny I double checked the code of sso plugin, `handleSessionResp` and `handleRedirectResp` should have ability to handle some of the errors(e.g. invalid code, client.. etc, see `callback_redirect_response` decorator in sso plugin).

We should print the error to the response body only for some critical case, e.g. callback url is not valid we should not redirect the result to the provided callback url. Or any other case we cannot redirect the result.

Redirect the result (both success and error) back to the app is important, so that user can go back to the original app. SDK will parse the result and call the success or error callback accordingly.

kaojohnny:
@carmenlau @Steven-Chan I added some commits when I testing with JS SDK, when you have time, please help review:

- Should SetCookie for both redirect or popup mode c3a259d
- Should set path when set redirect login cookie 58b41c4
- Set correct API endpoint to iframe_html and auth_handler_html a8b8584
- Reverse proxy should also pass query parameters a847bac

pkg/auth/handler/sso/auth_handler.go
carmenlau: 1. The callback url validation handling is correct in this PR.

2. `callback_redirect_response ` is a decorator used by `auth_handler`. For all skygear errors raise in `auth_handler`, `callback_redirect_response` will catch it and change it to response object in Line 124. `handler_redirect_response` and `handler_session_response` will encode the error response, set it in cookies and redirect or render html.

The other parts of the PR looks good and the PR becomes huge, maybe let me merge it first and open another PR for the error handling?

Feel free ping me if we need discuss directly. Thanks Johnny!

kaojohnny: Thanks, if you think it is okay, please help merge it first, I will have another PR to handle errors.

PR #831, opened 2018-12-07
next: Forgot password reset password handler

Steven-Chan:
connect #778
requires #823


Steven-Chan:
@carmenlau Rebased, and updated the following

- update `token_valid_since`
- return with auth response instead of OK
- add audit trail

Issue #781, opened 2018-11-09
next: Forgot password verification code submission handler (form and api)


Steven-Chan:
Note:
Need to update user verification state, which was implemented with save hook.

PR #846, opened 2018-12-19
next: Verify code form handler

Steven-Chan:
connect #781


kaojohnny:
@Steven-Chan Please help rebase, thanks.

PR #849, opened 2018-12-20
next: SSO should support auto link existed user account

kaojohnny:
connects #841 

1. If app's authRecordKeys contains ['email'], then auto connect procedure will be executed.
2. If auto connect is executed, it will find corresponding password principal base on authData (which is a map contains provider email), if it finds one, then will use it as the user principal.
3. When signup a new user, and app's authRecordKeys contains ['email], it will generate an empty password principal as well.

### DIFF
--- a/pkg/auth/handler/sso/sso.go
+++ b/pkg/auth/handler/sso/sso.go
@@ -126,9 +126,13 @@ func (h respHandler) handleLogin(
 		err = nil
 	}
 
-	if valid := h.PasswordAuthProvider.IsAuthDataMatching(oauthAuthInfo.ProviderAuthData); valid {
-		// provider authData match app authRecordKeys,
-		// start auto-connect procedure
+	if valid := h.PasswordAuthProvider.IsAuthDataValid(oauthAuthInfo.ProviderAuthData); valid {
+		// provider authData matches app's authRecordKeys,
+		// then it starts auto-link procedure.
+		//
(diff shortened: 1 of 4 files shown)


pkg/auth/handler/sso/sso.go
```diff
@@ -126,7 +126,14 @@ func (h respHandler) handleLogin(
 		err = nil
 	}
 
-	// TODO: handle auto link user
+	if valid := h.PasswordAuthProvider.IsAuthDataMatching(oauthAuthInfo.ProviderAuthData); valid {
```
Steven-Chan: Seems it does not restrict the auth record key to `email` 🤔

kaojohnny: I am thinking since `oauthAuthInfo.ProviderAuthData` contains `email` for now, and if `IsAuthDataMatching` is true, it means `AuthRecordKeys` contains `["email"]`. And in the future, if skygear next support hook mechanism when process `AuthData`, it will return correct `authData` by the hook function.

Do you think such idea correct? Or I miss anything.

kaojohnny: I added some comments, see if it is helpful. `oauthAuthInfo.ProviderAuthData` is the same structure of `authData`, like `{email: "john.doe@example.com"}`

Steven-Chan:
@kaojohnny I find the same problem of test setup in login with custom token, which I've just opened a PR to fix it https://github.com/SkygearIO/skygear-server/pull/854

PR #852, opened 2018-12-21
next: Refactor error handling in sso auth_handler

kaojohnny:
- when callback url is missing, return 400 bad request
- in other cases, should send back error response to the client via redirect

connects #788

### DIFF
--- a/pkg/auth/dependency/sso/authinfo.go
+++ b/pkg/auth/dependency/sso/authinfo.go
@@ -15,6 +15,16 @@ type getAuthInfoRequest struct {
 }
 
 func (h getAuthInfoRequest) getAuthInfo() (authInfo AuthInfo, err error) {
+	authInfo = AuthInfo{
+		ProviderName: h.providerName,
+	}
+
+	state, err := DecodeState(h.stateJWTSecret, h.encodedState)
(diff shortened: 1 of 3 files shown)


carmenlau:
1. Before running `h.Provider.GetAuthInfo`, `oauthAuthInfo.State.CallbackURL` is empty. All the errors occur before running `h.Provider.GetAuthInfo` will become `Missing callback url`? 

1. In the sso auth handler, there is 2 defer functions. One is for rendering the response and the other one is for transaction commit or rollback handlings. The transaction defer function will run before the rendering defer function. If there is error in the rendering function, e.g. invalid callback url. The transaction will still be committed which is not expected. I had similar discussion with @Steven-Chan when working on `ForgotPasswordResetFormHandler`, we thought that transaction defer function should wrap the whole handler. If there is error when render, we should rollback the transaction too.

kaojohnny:
@carmenlau I made some update, when you have time, please help review. Thanks.

carmenlau:
@kaojohnny Would you mind help update the error message of `InvalidClient` in this PR too. Now is `auth_data or password incorrect` which is not appropriate. The refactor looks good, thanks so much!

Issue #858, opened 2019-01-03
next: sso login the same user again create duplicate `_auth_princial` 


carmenlau:
When user use sso login the same account again

#### Expected Results
- Login with the same core user, no new `_auth_princial` or `_auth_provider_oauth` should be created. Origin `_auth_provider_oauth` record `token_response` and `profile` should be updated.

#### Actual Results
- Login with the same core user, create duplicate `_auth_princial` or `_auth_provider_oauth` records

PR #860, opened 2019-01-04
next: Should link user only when oauth principal can't be found

kaojohnny:
connects #858

### DIFF
--- a/pkg/auth/handler/sso/sso.go
+++ b/pkg/auth/handler/sso/sso.go
@@ -215,6 +215,8 @@ func (h respHandler) authLinkUser(oauthAuthInfo sso.AuthInfo) (*oauth.Principal,
 			return nil, err
 		}
 		return &oauthPrincipal, nil
+	} else if e != skydb.ErrUserNotFound {
+		return nil, e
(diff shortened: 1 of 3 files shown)


pkg/auth/handler/sso/sso.go
```diff
@@ -196,17 +179,42 @@ func (h respHandler) handleLogin(
...
+		// for example, if oauthAuthInfo.ProviderAuthData is {"email", "john.doe@example.com"},
+		// it will be a valid authData if authRecordKeys is [["username"], ["email"]] or [["email"]]
+		// so, the oauthAuthInfo.ProviderAuthDat can be used as a password principal authData
+		return h.authLinkUser(oauthAuthInfo)
+	}
+
+	return nil, nil
+}
+
 func (h respHandler) authLinkUser(oauthAuthInfo sso.AuthInfo) (*oauth.Principal, error) {
```
carmenlau: Should `authLinkUser` return error if `h.PasswordAuthProvider.GetPrincipalByAuthData` is not `ErrUserNotFound`?

Issue #857, opened 2019-01-03
next: should not create auth data if the auth keys value are empty


carmenlau:
When the auth keys are ["email", "username"].

### Case 1
If user signup with data
```
{
    "auth_data": {
        "username": "",
        "email": ""
    },
    "password": "123456",
    "profile": {}
}
```

#### Expected Results
- Should fail the signup with error auth data is missing

#### Actual Results
- Signup successfully with 2 _auth_provider_password records, `{"email": ""}` and `{"username": ""}`

### Case 2
If user signup with data
```
{
    "auth_data": {
        "username": "user",
        "email": ""
    },
    "password": "123456",
    "profile": {}
}
```

#### Expected Results
- Signup successfully with 1 _auth_provider_password record, `{"username": "user"}`

#### Actual Results
- Signup successfully with 2 _auth_provider_password records, `{"email": ""}` and `{"username": "user"}`

PR #861, opened 2019-01-04
next: Should validate zero value in authData

kaojohnny:
connects #857

### DIFF
--- a/pkg/auth/dependency/provider/password/authdata.go
+++ b/pkg/auth/dependency/provider/password/authdata.go
@@ -1,21 +1,19 @@
 package password
 
-import "reflect"
-
 type authDataChecker interface {
-	isValid(authData map[string]interface{}) bool
-	isMatching(authData map[string]interface{}) bool
+	isValid(authData map[string]string) bool
+	isMatching(authData map[string]string) bool
 }
 
 type defaultAuthDataChecker struct {
 	authRecordKeys [][]string
 }
 
-func (c defaultAuthDataChecker) isValid(authData map[string]interface{}) bool {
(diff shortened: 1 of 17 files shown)


pkg/auth/dependency/provider/password/authdata.go
```diff
@@ -55,14 +57,21 @@ var (
...
 
 	for _, ks := range authRecordKeys {
 		m := make(map[string]interface{})
 		for _, k := range ks {
 			for dk := range authData {
-				if k == dk && authData[dk] != nil {
+				if k == dk &&
+					authData[dk] != nil &&
+					// for non-nil interface but underlying value is a zero value, it won't be considered as a valid authData
+					!reflect.DeepEqual(authData[dk], reflect.Zero(reflect.TypeOf(authData[dk])).Interface()) {
```
carmenlau: In the original design, we haven't restrict the type the auth data. But there are some problems.
1. For some data type like boolean, it will be rejected here. But it should be valid.
2. Since there is no restriction, user may pass some strange data here like json object. It is not the use case we want to support.
3. We don't have user defined validation to auth data, there will have problem if user pass some non-string data to email when sending welcome email.

We had a short chat on this, and we think string can fulfill most of the use case already. In order to keep it simple, maybe we can restrict user using string in auth data. Thought?

kaojohnny: I think it is okay if we don't expect `authData` in other data type. Do you want me to replace all `authData map[string]interface{}` to `authData map[string]string` in current code base? or just do type check in this function?

PR #865, opened 2019-01-10
add tcp keepalive option for APNS connection

hidden-dimensions:
APNS use HTTP/2 and multiplex multiple payload on same TCP
connection, which default disabled TCP keepalive mechanic,
or otherwise rely on system-level setting of like 7200 seconds.

Such connection, when idle, could be dropped by NAT or firewall,
and we will only receive TCP read error after a long timeout
of 3~30 minutes, furthermore any PN queued in such connection
will fail and discarded.

To deal with this, we shall allow setting a tcp keepalive option.

### DIFF
--- a/main.go
+++ b/main.go
@@ -589,6 +589,7 @@ func initCertBasedAPNSPusher(
 		push.GatewayType(config.APNS.Env),
 		cert,
 		key,
+		config.APNS.Keepalive,
 	)
 	if err != nil {
 		logger.Fatalf("Failed to set up push sender: %v", err)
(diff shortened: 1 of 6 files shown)


hidden-dimensions:
The keep alive setting might be different per environment, for example, azure would drop idle connection at 4 minutes, aws is 350s, it would be too opinion based to hard-code an internal value for it.

Also, the original behavior is no keep alive, and this PR gives exactly same behavior if config is not set (zero value).

rickmak:
First, skygear as a software that aimed to boost developer productivity, our default should sensible to make the developer don't need to know the azure and aws tcp drop time is different. We should provide a sensible default so the developer will just provide a `true` and things will works.

Second, in existing skygear config, we don't use zero value as turn off. We use explicit boolean config for the enable of a services. You may reference to `UserAudit.Enable`, `GCM.Enable`, and `APNS.Enable`.

Finally, why not turn the keepalive on by default since we found this problem? I cannot think of a normal usecase would like the keepalive is off by default.

carmenlau:
Merged to master via 9d4b6f57
Merged to v1-branch via e74dc7ea

PR #902, opened 2019-03-05
Cloud function routing

Steven-Chan:
Notes:

- Gear route is changed from `/{gear}/*` to `/_{gear}/*`.
- I've come up with the name `CloudCode` / `cloud_code` in code level as a general name for cloud function and microservices, in response to https://github.com/SkygearIO/features/pull/278#discussion_r258941485.
- `Context` is used to pass data between middlewares and handlers, which is against what we have discussed at the very early stage.

connect #898

### DIFF
--- a/pkg/gateway/handler/cloudcode.go
+++ b/pkg/gateway/handler/cloudcode.go
@@ -1,0 +1,33 @@
+package handler
+
+import (
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+
+	"github.com/skygeario/skygear-server/pkg/gateway/config"
+	"github.com/skygeario/skygear-server/pkg/gateway/model"
+)
+
+func NewCloudCodeHandler(routerConfig config.RouterConfig) http.HandlerFunc {
+	proxy := newCloudCodeReverseProxy(routerConfig)
+	return proxy.ServeHTTP
+}
+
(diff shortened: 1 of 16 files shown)


Steven-Chan:
> `Context` is used to pass data between middlewares and handlers, which is against what we have discussed at the very early stage.

I think passing data with context is suitable in the case of this PR. However, if we want to ensure type safety, which is the biggest argument for not using context, I suggest we use a struct to contain data for the whole gateway and pass it in context.

Steven-Chan:
Update: defined struct for context.

carmenlau:
@kaojohnny See if it looks good for you too when you have time?

pkg/gateway/handler/cloudcode.go
```diff
@@ -0,0 +1,33 @@
...
+		query := req.URL.RawQuery
+		fragment := req.URL.Fragment
+		var err error
+		u, err := url.Parse(routerConfig.CloudCodeGatewayURL)
+		if err != nil {
+			panic(err)
+		}
+		cloudCode := model.CloudCodeFromContext(req.Context())
+		req.URL = u
+		req.URL.Path = "/function" + cloudCode.TargetPath
```
kaojohnny: @Steven-Chan Just correct me if I misunderstand anything :)

1. `/function` here means the type of cloud function? and we will have something like `/http-handler` in the future?
2. Forward path seems missing when forward to cloud function gateway?

PR #906, opened 2019-03-12
Fix gear routing match order and fix context nil exception

kaojohnny:
auth gear `PathPrefix` should put before cloud function

### DIFF
--- a/pkg/gateway/middleware/app.go
+++ b/pkg/gateway/middleware/app.go
@@ -20,7 +20,7 @@ func (f FindAppMiddleware) Handle(next http.Handler) http.Handler {
 			return
 		}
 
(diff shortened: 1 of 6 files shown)


pkg/gateway/middleware/app.go
```diff
@@ -20,7 +20,7 @@ func (f FindAppMiddleware) Handle(next http.Handler) http.Handler {
 			return
 		}
 
-		ctx := gatewayModel.GatewayContextFromContext(r.Context())
+		ctx := gatewayModel.Context{}
```
Steven-Chan: I think a better approach is to make `gatewayModel.GatewayContextFromContext` lazy, i.e. return `gatewayModel.Context{}` if nil.

kaojohnny:
@Steven-Chan Updated.

Issue #880, opened 2019-02-01
next: [skygear-server] auth gear's user object response should include user's metadata


kaojohnny:
Before

```json
{
    "result": {
        "user_id": "e44f80ee-06af-4dd9-89dd-42735ef63792",
        "profile": {
            "_access": null,
            "_created_at": "2019-02-18T08:09:33.190914Z",
            "_created_by": "e44f80ee-06af-4dd9-89dd-42735ef63792",
            "_id": "user/e44f80ee-06af-4dd9-89dd-42735ef63792",
            "_ownerID": "e44f80ee-06af-4dd9-89dd-42735ef63792",
            "_recordID": "e44f80ee-06af-4dd9-89dd-42735ef63792",
            "_recordType": "user",
            "_type": "record",
            "_updated_at": "2019-02-18T08:09:33.190914Z",
            "_updated_by": "e44f80ee-06af-4dd9-89dd-42735ef63792",
            "email": "123456@example.com",
            "username": 123456
        },
        "access_token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJqdGkiOiJiZGNmNTFlMC1iYTFlLTQ1OTgtOGM1MC1lNDQ4MDQ1YzE1YzEiLCJpYXQiOjE1NTI0NDY1NTgsImlzcyI6Im15YXBwIiwic3ViIjoiZTQ0ZjgwZWUtMDZhZi00ZGQ5LTg5ZGQtNDI3MzVlZjYzNzkyIn0.NNfcr9ZyeukilhDBOskiHvWfkdmCVjSkQruRo3LBBxk",
        "last_seen_at": "2019-03-13T02:55:04.27151Z",
        "verified": false,
        "verify_info": {}
    }
}
```

After

```
{
    user_id: <id>,
    created_at: <created_at>,
    updated_at: <updated_at>,
    last_seen_at: <last_seen_at>,
    last_login_at: <last_login_at>,
    verified: <verified>,
    verify_info: <verified_info>,
    roles: [<role>, <role>, <role>, ...],
    access_token: <token>,
    metadata: {
      // loginIDs
      username: <username>,
      email: <email>,
      // custom user attributes
      avatar_url: <avatarUrl>,
      name: <name>,
      nickname: <nickname>,
      birthday: <birthday>,
      preferred_lang: <preferred_lang>,
    }
}
```

PR #909, opened 2019-03-13
next: Update auth response

kaojohnny:
connects #880


pkg/auth/dependency/userprofile/userprofile.go
```diff
@@ -46,47 +32,13 @@ type Store interface {
 }
 
 func (u UserProfile) MarshalJSON() ([]byte, error) {
-	return json.Marshal(u.ToMap())
+	return json.Marshal(u.Data)
```
Steven-Chan: Why is the marshal and unmarshal function only use `u.Data` but not the whole `u`?

kaojohnny: Originally we had record alike`user.profile`, that is:
```
        "profile": {
            "_access": null,
            "_created_at": "2019-02-18T08:09:33.190914Z",
            "_created_by": "e44f80ee-06af-4dd9-89dd-42735ef63792",
            "_id": "user/e44f80ee-06af-4dd9-89dd-42735ef63792",
            "_ownerID": "e44f80ee-06af-4dd9-89dd-42735ef63792",
            "_recordID": "e44f80ee-06af-4dd9-89dd-42735ef63792",
            "_recordType": "user",
            "_type": "record",
            "_updated_at": "2019-02-18T08:09:33.190914Z",
            "_updated_by": "e44f80ee-06af-4dd9-89dd-42735ef63792",
            "email": "123456@example.com",
            "username": 123456
        },
```
For the latest design spec, we decide to make user `metadata` contains changeable attributes only, that is:
```
        "metadata": {
            "email": "123456@example.com",
            "username": 123456
        },
```
which is the `u.Dta

pkg/auth/dependency/userverify/sender.go
```diff
@@ -110,17 +110,17 @@ func prepareVerifyRequestContext(
...
-		"record_key":   verifyCode.RecordKey,
-		"record_value": verifyCode.RecordValue,
-		"user_id":      userProfile.RecordID,
-		"user":         userProfile.ToMap(),
-		"code":         verifyCode.Code,
+		"appname":       appName,
+		"record_key":    verifyCode.RecordKey,
+		"record_value":  verifyCode.RecordValue,
+		"user_id":       userProfile.ID,
+		"user_metadata": userProfile.Data,
```
Steven-Chan: I think the context for email template should include the whole user object for the key `user`, use case like displaying user created time or update time.

Same for other email template context.

pkg/auth/response/response.go
```diff
@@ -23,30 +23,34 @@ import (
...
-	UserID      string                  `json:"user_id,omitempty"`
-	Profile     userprofile.UserProfile `json:"profile"`
-	Roles       []string                `json:"roles,omitempty"`
-	AccessToken string                  `json:"access_token,omitempty"`
-	LastLoginAt *time.Time              `json:"last_login_at,omitempty"`
-	LastSeenAt  *time.Time              `json:"last_seen_at,omitempty"`
-	Verified    bool                    `json:"verified"`
-	VerifyInfo  map[string]bool         `json:"verify_info"`
+	UserID      string           `json:"user_id,omitempty"`
+	Metadata    userprofile.Data `json:"metadata,omitempty"`
```
Steven-Chan: It seems better to keep metadata empty dictionary instead of using `omitempty`. This would be easier for client side to decode the data, because it can always assume the key `metadata` there.

PR #916, opened 2019-03-22
Provide a basic interface for updating metadata

kaojohnny:
connects #877

### DIFF
--- a/pkg/auth/dependency/provider/password/mock_password.go
+++ b/pkg/auth/dependency/provider/password/mock_password.go
@@ -4,7 +4,6 @@ import (
 	"reflect"
 
 	"github.com/skygeario/skygear-server/pkg/core/skydb"
-	"github.com/skygeario/skygear-server/pkg/core/utils"
 	"golang.org/x/crypto/bcrypt"
 )
 
@@ -42,10 +41,13 @@ func (m *MockProvider) IsAuthDataMatching(authData map[string]string) bool {
 }
 
 func (m *MockProvider) GetLoginIDMetadataFlattenedKeys() []string {
-	var output []string
+	output := make([]string, 0, len(m.loginIDMetadataKeys))
+	bookkeeper := make(map[string]bool)
+
 	for _, keys := range m.loginIDMetadataKeys {
 		for _, key := range keys {
-			if !utils.SliceContains(output, key) {
+			if _, ok := bookkeeper[key]; !ok {
+				bookkeeper[key] = true
 				output = append(output, key)
 			}
 		}
--- a/pkg/auth/dependency/provider/password/mock_password.go
+++ b/pkg/auth/dependency/provider/password/mock_password.go
@@ -4,6 +4,7 @@ import (
 	"reflect"
 
 	"github.com/skygeario/skygear-server/pkg/core/skydb"
+	"github.com/skygeario/skygear-server/pkg/core/utils"
 	"golang.org/x/crypto/bcrypt"
 )
 
@@ -38,6 +39,19 @@ func (m *MockProvider) IsAuthDataValid(authData map[string]string) bool {
 
 func (m *MockProvider) IsAuthDataMatching(authData map[string]string) bool {
 	return m.authDataChecker.isMatching(authData)
+}
+
+func (m *MockProvider) GetLoginIDMetadataFlattenedKeys() []string {
+	var output []string
+	for _, keys := range m.loginIDMetadataKeys {
+		for _, key := range keys {
+			if !utils.SliceContains(output, key) {
+				output = append(output, key)
+			}
+		}
+	}
+
+	return output
 }
 
 // CreatePrincipalsByAuthData creates principals by authData
--- a/pkg/auth/handler/update_metadata.go
+++ b/pkg/auth/handler/update_metadata.go
@@ -10,12 +10,15 @@ import (
 	"github.com/skygeario/skygear-server/pkg/auth"
 	"github.com/skygeario/skygear-server/pkg/auth/response"
 	coreAuth "github.com/skygeario/skygear-server/pkg/core/auth"
+	"github.com/skygeario/skygear-server/pkg/core/auth/authinfo"
 	"github.com/skygeario/skygear-server/pkg/core/auth/authz"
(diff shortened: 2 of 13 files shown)


pkg/core/utils/utils.go
```diff
@@ -63,3 +67,26 @@ func strAt(slice []string, str string) int {
 	}
 	return -1
 }
+
+func SliceContains(in interface{}, elem interface{}) bool {
+	inValue := reflect.ValueOf(in)
+	inType := inValue.Type()
+
+	if inType.Kind() == reflect.Slice {
```
Steven-Chan: Using reflect here looks overkill, the function name has already suggested that the input is a slice.

pkg/auth/dependency/provider/password/mock_password.go
```diff
@@ -40,6 +41,19 @@ func (m *MockProvider) IsAuthDataMatching(authData map[string]string) bool {
 	return m.authDataChecker.isMatching(authData)
 }
 
+func (m *MockProvider) GetLoginIDMetadataFlattenedKeys() []string {
+	var output []string
+	for _, keys := range m.loginIDMetadataKeys {
+		for _, key := range keys {
+			if !utils.SliceContains(output, key) {
```
Steven-Chan: This ends up with a 3-level nested loop, which the last layer is not needed.

Something like this: https://kylewbanks.com/blog/creating-unique-slices-in-go

pkg/auth/handler/update_metadata.go
```diff
@@ -0,0 +1,137 @@
...
+	inMetadata map[string]interface{},
+) map[string]interface{} {
+	loginIDMetadataKeys := h.PasswordAuthProvider.GetLoginIDMetadataFlattenedKeys()
+
+	output := make(map[string]interface{})
+	for k := range orgMetadata {
+		output[k] = orgMetadata[k]
+	}
+	for k, v := range inMetadata {
+		if !utils.SliceContains(loginIDMetadataKeys, k) {
```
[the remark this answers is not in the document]
kaojohnny: This PR is mainly for providing very basic functionality for update metadata for JS SDK, after https://github.com/SkygearIO/features/pull/295 is confirmed, these codes should be removed afterward.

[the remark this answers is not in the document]
Steven-Chan: Got it, i suggest you add comment to this.

kaojohnny:
@Steven-Chan I don't think we should support update someone else's metadata, do you think it is okay to change `skygear.auth.updateMetadata(user)` to `skygear.auth.updateCurrentUserMetadata()`?

pkg/core/utils/utils.go
```diff
@@ -63,3 +67,23 @@ func strAt(slice []string, str string) int {
 	}
 	return -1
 }
+
+func SliceContains(in interface{}, elem interface{}) bool {
+	inValue := reflect.ValueOf(in)
```
Steven-Chan: > Using reflect here looks overkill, the function name has already suggested that the input is a slice.

I didn't look clearly enough to realise the whole function is using reflection. I was trying to say using reflection for this function looks overkill. 

We don't actually have a case that needs to find a `interface{}` in a `[]interface{}`, so having a `StringSliceContains` is good enough for this case. This is where golang suffers when we want to use generic.

Steven-Chan:
> @Steven-Chan I don't think we should support update someone else's metadata, do you think it is okay to change `skygear.auth.updateMetadata(user)` to `skygear.auth.updateCurrentUserMetadata()`?

IMO, there must be a way for the developer to update a specific user's metadata, e.g. user management for admin. And if we don't open up the db to the developers, we will need to provide an API to do it.

I was thinking `skygear.auth.updateMetadata({ metadata: user.metadata })` may be better in the example.

chpapa:
> IMO, there must be a way for the developer to update a specific user's metadata, e.g. user management for admin. And if we don't open up the db to the developers, we will need to provide an API to do it.
> 
> I was thinking `skygear.auth.updateMetadata({ metadata: user.metadata })` may be better in the example.

Probably as an admin only functions?

pkg/auth/handler/update_metadata.go
```diff
@@ -125,11 +125,11 @@ func (h UpdateMetadataHandler) mergeMetadata(
...
 	output := make(map[string]interface{})
-	for k := range orgMetadata {
-		output[k] = orgMetadata[k]
-	}
 	for k, v := range inMetadata {
-		if !utils.StringSliceContains(loginIDMetadataKeys, k) {
+		output[k] = v
+	}
+	for k, v := range orgMetadata {
+		if utils.StringSliceContains(loginIDMetadataKeys, k) {
```
Steven-Chan: Looks weird, why update with old data after new data?

I thought the previous implementation is able to delete attribute by setting value to `null`? This is the same behaviour as v1 record save. I think you may write a more complete test case, deleting non login id keys.

kaojohnny: My original intention is to copy all attributes from new data and overwrite the original loginID value.

I update the implementation as:

```
	output := make(map[string]interface{})
	// copy loginID to the profile
	for _, v := range loginIDMetadataKeys {
		if pv, ok := orgMetadata[v]; ok {
			output[v] = pv
		}
	}
	// copy rest attributes to the profile
	for k, v := range inMetadata {
		if !utils.StringSliceContains(loginIDMetadataKeys, k) {
			output[k] = v
		}
	}
```

Feel free to Let me know your comments about this. Thanks.

kaojohnny:
> IMO, there must be a way for the developer to update a specific user's metadata, e.g. user management for admin. And if we don't open up the db to the developers, we will need to provide an API to do it.

> I was thinking skygear.auth.updateMetadata({ metadata: user.metadata }) may be better in the example.

>> Probably as an admin only functions?

Probably we can adjust the payload of `/update_metadata` to 

```
{
  user_id: <user_id>,
  metadata: <metadata>
}
```

1. In JS SDK, the API is `updateMetadata(metadata: user.metadata)`, `user_id` is current user id.
2. From cloud function, a developer can set `user_id` and assign master key to grant to update specific user's metadata.

Issue #904, opened 2019-03-11
auth gear should trigger cf module when user auth events fired


kaojohnny:
auth gear should trigger cf module when certain user auth events fired.

following are the list of events:

phase 1:
- [x] signup

phase 2:
- [ ] login
- [ ] logout
- [ ] roles
- [ ] enable
- [ ] password
- [ ] verify
- [ ] metadata

Each event includes sync, async, before and after. For sync event, if cf module returns failed, DB should roll back and return the formatted error to the client.

#### Expected Results

A developer can create auth hooked cloud functions to handle auth events.

--- 

- [x] Request Payload
- [ ] X-Skygear-Webhook-Req-ID
- [ ] X-Skygear-Webhook-Async
- [ ] X-Skygear-Webhook-Signature

PR #925, opened 2019-04-12
next: Implement auth hooks

kaojohnny:
require #922 
connects #904


kaojohnny:
@Steven-Chan @carmenlau 

For auth hooks, I plan to add a table in `app_<app_name>`, this may affect `skycli` and `skygear-controller` design, want to get your feedbacks. 

```sql
CREATE TABLE _auth_hooks (
  id text PRIMARY KEY,
  async boolean NOT NULL DEFAULT true, --sync, async
  event_name text NOT NULL, --after_signup, before_login
  url text NOT NULL,
  deleted boolean NOT NULL DEFAULT false, -- hook is deleted,
  timeout integer NOT NULL -- sync 5~10 secs, async 60 ~ 300 secs
);
```

How to manage secrets is missing, may you help give me some advice?
- `X-Skygear-Webhook-Signature` header: for webhook, it is used for verifying the request.

kaojohnny:
@Steven-Chan @carmenlau After discussion, I will use auto hooks configurations in tenant config, followings are the structure:

```
type TenantConfiguration struct {
        ......
	AuthHooks       []AuthHook                  `json:"AUTH_HOOKS" msg:"AUTH_HOOKS"`
}

type AuthHook struct {
	CloudCodeID string `msg:"CLOUD_CODE_ID" envconfig:"CLOUD_CODE_ID" json:"CLOUD_CODE_ID"`
	Async       bool   `msg:"ASYNC" envconfig:"ASYNC" json:"ASYNC"`
	Event       string `msg:"EVENT" envconfig:"EVENT" json:"EVENT"`
	URL         string `msg:"URL" envconfig:"URL" json:"URL"`
	TimeOut     int    `msg:"TIME_OUT" envconfig:"TIME_OUT" json:"TIME_OUT"`
}
```

kaojohnny:
@Steven-Chan @carmenlau When you have time, this is a basic version of auth hook handling, when you have time, please help do a quick review to make sure we are in the same direction. 

Thanks.

pkg/core/config/tenant.go
```diff
@@ -175,6 +176,14 @@ type NexmoConfiguration struct {
 	From      string `msg:"FROM" envconfig:"NEXMO_FROM" json:"FROM"`
 }
 
+type AuthHook struct {
```
Steven-Chan: This should be `Webhook` or `Hook` instead of `AuthHook`, because the same structure is used across all gear.

pkg/auth/hook_handler.go
```diff
@@ -0,0 +1,55 @@
...
+	ExecBeforeHooks(payload interface{}, user *response.User) error
+	HandleRequest(payload interface{}, user *response.User) (interface{}, error)
+	ExecAfterHooks(payload interface{}, resp interface{}, user response.User) error
+}
+
+type hookExecutor struct {
+	handler HookHandler
+}
+
+func HookHandlerToAPIHandler(h HookHandler, txContext db.TxContext) http.Handler {
```
Steven-Chan: Name should be `HookHandlerToHandler`? or it should return the skygear api handler.

kaojohnny:
@Steven-Chan Updated, thanks.

PR #929, opened 2019-04-23
Gateway auth info

louischan-oursky:
This PR implements this feature https://github.com/SkygearIO/features/pull/303

### DIFF
--- a/pkg/gateway/middleware/authinfo.go
+++ b/pkg/gateway/middleware/authinfo.go
@@ -7,13 +7,8 @@ import (
 	"github.com/skygeario/skygear-server/pkg/core/auth/authinfo"
 	"github.com/skygeario/skygear-server/pkg/core/auth/authtoken"
 	"github.com/skygeario/skygear-server/pkg/core/db"
+	coreHttp "github.com/skygeario/skygear-server/pkg/core/http"
 	"github.com/skygeario/skygear-server/pkg/core/model"
-)
-
-const (
-	httpHeaderAuthInfoID       = "x-skygear-auth-userid"
-	httpHeaderAuthInfoVerified = "x-skygear-auth-verified"
-	httpHeaderAuthInfoDisabled = "x-skygear-auth-disabled"
 )
 
 // AuthInfoMiddleware injects auth info headers into the request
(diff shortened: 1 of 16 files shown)


pkg/gateway/middleware/authinfo.go
```diff
@@ -0,0 +1,70 @@
...
+type AuthInfoMiddleware struct {
+	TokenStore    authtoken.Store `dependency:"TokenStore"`
+	AuthInfoStore authinfo.Store  `dependency:"AuthInfoStore"`
+	TxContext     db.TxContext    `dependency:"TxContext"`
+}
+
+// Handle implements InjectableMiddleware.
+func (m *AuthInfoMiddleware) Handle(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer next.ServeHTTP(w, r)
```
louischan-oursky: I wonder if we encounter error, should we still call the next handler?

Steven-Chan: Depends on how you handle the error.

If you write something to the response in the middleware, then the next handler won't be able to change status code and headers. Moreover, the next handler would write data on top of the data you have written in the middleware. This would probably lead unexpected behaviour.

See https://golang.org/pkg/net/http/#ResponseWriter.

louischan-oursky: OK. Let me change to not invoke the next handler if error is detected.

pkg/gateway/middleware/inject.go
```diff
@@ -0,0 +1,30 @@
...
+// Middleware at request time.
+type Injecter struct {
+	Middleware InjectableMiddleware
+	Dependency gateway.DependencyMap
+}
+
+// Handle implements gorilla middleware signature.
+func (m Injecter) Handle(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		inject.DefaultRequestInject(m.Middleware, m.Dependency, r)
```
carmenlau: Since gateway is multi-tenant, it is not safe to inject dependency to the same middleware instance and use it for all the requests. If we want to reuse the core inject function, we will need to do it like core. We should use the middleware factory, the middleware instance will be created in the request runtime and should not be reused.

louischan-oursky: Fixed.

Issue #932, opened 2019-04-26
Update gateway to provide CORS header for CF


carmenlau:
Gateway should get the CORS setting from tenant config and update the headers. OPTIONS request should be handled by gateway, and should not route to the cloud code backend.

tensiuyan:
ref #890

louischan-oursky:
Use this library https://github.com/iawaknahc/originmatcher @kiootic

louischan-oursky:
Done in https://github.com/SkygearIO/skygear-server/pull/941

PR #941, opened 2019-06-18
Respond with correct CORS headers in middleware #932

kiootic:


pkg/core/middleware/cors.go
```diff
@@ -25,22 +27,32 @@ type CORSMiddleware struct {
...
-
 		tConfig := config.GetTenantConfig(r)
-		w.Header().Set("Access-Control-Allow-Origin", tConfig.CORSHost)
-		w.Header().Set("Access-Control-Allow-Credentials", "true")
-
-		if corsMethod != "" {
-			w.Header().Set("Access-Control-Allow-Methods", corsMethod)
+		matcher, err := originmatcher.Parse(tConfig.CORSHost)
+		if err != nil {
+			panic(fmt.Errorf("Unable to parse CORS host config: %v", err.Error()))
```
louischan-oursky: Let's not panic if parsing failed. Treat this case as never matched. So

```golang
if err != nil && origin != "" && matcher.MatchOrigin(origin) {
```

kiootic: This would be a silent configuration failure, which app developer may have a hard time to figure out why it's not returning correct CORS headers.

If panic is undesirable, I suppose the CORS host config would be validated before committing into DB?

louischan-oursky: Currently we have no ways for the developer to config CORS host. The observed existing values are `""` and `"*"`. When we allow the developer to edit it, we definitely will use originmatcher to parse it once before storing in database.

louischan-oursky: I just discovered a bug in the library. I have fixed it

https://github.com/iawaknahc/originmatcher/commits/master

Please update.

PR #943, opened 2019-06-25
Revamp TenantConfiguration

louischan-oursky:


pkg/core/config/newtenant.go
```diff
@@ -0,0 +1,532 @@
...
+	return url
+}
+
+// UserConfiguration represents user-editable configuration
+type UserConfiguration struct {
+	APIKey           string                         `json:"api_key" yaml:"api_key" msg:"api_key"`
+	MasterKey        string                         `json:"master_key" yaml:"master_key" msg:"master_key"`
+	URLPrefix        string                         `json:"url_prefix" yaml:"url_prefix" msg:"url_prefix"`
+	CORS             CORSConfiguration              `json:"cors" yaml:"cors" msg:"cors"`
+	Auth             NewAuthConfiguration           `json:"auth" yaml:"auth" msg:"auth"`
```
kiootic: Do we remove the `New` prefix? The old configuration seems deleted.

louischan-oursky: Yes. we should. Let me remove all `New` prefixes.

louischan-oursky:
@kiootic Remove the `New` prefix both in filenames and types.

rickmak:
@louischan-oursky How can I enter Standalone mode as there is no envconfig directive in the config?

rickmak:
I found the envconfig `STANDALONE` is working as default the loader use env key that is capitalized of the variable name.

PR #946, opened 2019-06-26
Update README.md

remcoplasmeyer:


chpapa:
It doesn't seems the change are necessary. Thanks @remcoplasmeyer but will close it first.

Issue #933, opened 2019-05-02
update the loginIDs configuration so that we know when it is an email field


tensiuyan:
We need a way to identify when the loginID is an email so that all email related APIs can work properly.

Email related APIs:
1. Email verification
2. Welcome email
3. Forgot password

PR #947, opened 2019-06-27
Implement Implict login ID key & Login ID realm

kiootic:
refs #933


pkg/auth/dependency/provider/password/password.go
```diff
@@ -8,6 +8,6 @@ type Provider interface {
 	CreatePrincipal(principal Principal) error
 	GetPrincipalByLoginIDWithRealm(loginIDKey string, loginID string, realm string, principal *Principal) (err error)
 	GetPrincipalsByUserID(userID string) ([]*Principal, error)
-	GetPrincipalsByEmail(email string) ([]*Principal, error)
+	GetPrincipalsByLoginID(loginIDKey string, loginID string) ([]*Principal, error)
```
louischan-oursky: login id key is not part of the unique constraint `(login_id, realm)`. This method should not take `loginIDKey`. Similarly, GetPrincipalByLoginIDWithRealm should not take `loginIDKey`

pkg/auth/handler/login.go
```diff
@@ -132,9 +113,13 @@ func (h LoginHandler) Handle(req interface{}) (resp interface{}, err error) {
 		}
 	}()
 
-	if valid := h.PasswordAuthProvider.IsLoginIDValid(payload.RawLoginID); !valid {
-		err = skyerr.NewInvalidArgument("invalid login_id, check your LOGIN_IDS_KEY_WHITELIST setting", []string{"login_id"})
-		return
+	if payload.LoginIDKey != "" {
+		loginIDMap := make(map[string]string)
+		loginIDMap[payload.LoginIDKey] = payload.LoginID
+		if valid := h.PasswordAuthProvider.IsLoginIDValid(loginIDMap); !valid {
```
louischan-oursky: I think the validation does not need the login id value. Can we change `func IsLoginIDValid(map[string]string) bool` to `func IsLoginIDKeyValid(string) bool`?

pkg/auth/handler/userverify/verify_request.go
```diff
@@ -124,20 +130,30 @@ func (h VerifyRequestHandler) Handle(req interface{}) (resp interface{}, err err
 	}
 	user := userFactory.NewUser(*authInfo, userProfile)
 
-	var value string
-	var ok bool
-	if value, ok = user.LoginIDs[payload.RecordKey]; !ok {
-		err = skyerr.NewError(skyerr.UnexpectedError, "Value of "+payload.RecordKey+" doesn't exist.")
+	principals, err := h.PasswordAuthProvider.GetPrincipalsByLoginID(payload.LoginIDKey, payload.LoginID)
```
louischan-oursky: We do not consider realm here. Can we document the expected behavior here?

pkg/auth/handler/forgotpwd/reset_form.go
```diff
@@ -291,7 +291,21 @@ func (h ForgotPasswordResetFormHandler) ServeHTTP(rw http.ResponseWriter, r *htt
 	}
 
 	hashedPassword := principals[0].HashedPassword
-	email := templateCtx.user.LoginIDs["email"]
+
+	email := ""
+	for _, principal := range principals {
+		if principal.LoginIDKey == "email" {
```
louischan-oursky: We cannot assume login id key`email` is email

pkg/auth/handler/signup.go
```diff
@@ -288,18 +289,18 @@ func (h SignupHandler) createPrincipal(payload SignupRequestPayload, authInfo au
 	return
 }
 
-func (h SignupHandler) sendWelcomeEmail(user response.User) {
-	if email, ok := user.LoginIDs["email"]; ok {
+func (h SignupHandler) sendWelcomeEmail(user response.User, loginIDs map[string]string) {
+	if email, ok := loginIDs["email"]; ok {
```
louischan-oursky: no assumption

pkg/auth/handler/sso/sso.go
```diff
@@ -241,7 +241,7 @@ func (h respHandler) createEmptyPasswordPrincipal(userID string, oauthAuthInfo s
 	if valid := h.PasswordAuthProvider.IsLoginIDValid(oauthAuthInfo.ProviderAuthData); valid {
 		// if ProviderAuthData matches loginIDsKeyWhitelist, and it can't be link with current account,
 		// we also creates an empty password principal for later the user can set password to it
-		return h.PasswordAuthProvider.CreatePrincipalsByLoginID(userID, "", oauthAuthInfo.ProviderAuthData)
+		return h.PasswordAuthProvider.CreatePrincipalsByLoginID(userID, "", oauthAuthInfo.ProviderAuthData, "")
```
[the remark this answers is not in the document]
kiootic: I'll remove it for now; developer should instead add new password principal after OAuth manually if required. We need to consider this use case when designing `createLoginID` API.

pkg/auth/handler/forgotpwd/forgot_password.go
```diff
@@ -101,7 +101,7 @@ func (h ForgotPasswordHandler) Handle(req interface{}) (resp interface{}, err er
 		"email": payload.Email,
 	}
 
-	principals, principalErr := h.PasswordAuthProvider.GetPrincipalsByEmail(payload.Email)
+	principals, principalErr := h.PasswordAuthProvider.GetPrincipalsByLoginID("email", payload.Email)
```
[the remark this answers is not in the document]
kiootic: The `"email"` parameter is temporary. However, it'll need `GetPrincipalsByLoginID` in final solution; we want user able to request password recovery by providing only email. (i.e. login ID)

pkg/auth/dependency/provider/password/mock_password.go
```diff
@@ -62,7 +77,8 @@ func (m *MockProvider) CreatePrincipal(principal Principal) error {
 	}
 
 	for _, p := range m.PrincipalMap {
-		if reflect.DeepEqual(principal.LoginID, p.LoginID) {
+		if reflect.DeepEqual(principal.LoginID, p.LoginID) &&
+			reflect.DeepEqual(principal.Realm, p.Realm) {
```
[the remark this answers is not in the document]
kiootic: I think yes.

pkg/auth/dependency/provider/password/password.go
kiootic: Login ID may be normalize based on the login ID key. I don't want to do it in handler, so it should be in this place. Later may divide this interface into Provider and Store, so `GetPrincipalsByLoginID` has login ID key parameter in Provider, and no login ID key parameter in Store.

pkg/auth/handler/login.go
kiootic: `IsLoginIDValid` is also used in signing up, which may have validation on login ID in future. I suggest keeping it at the moment.

pkg/auth/handler/userverify/verify_request.go
kiootic: This behavior is documented in proposed spec. I'll add a comment to clarify the intention here.

pkg/auth/handler/forgotpwd/reset_form.go
kiootic: Yes, this will be changed in stage 2. I just keep the behavior same for now.

pkg/auth/handler/login.go
louischan-oursky: Fair enough.

louischan-oursky:
Can we add `TODO` comment to places where temporary solutions are used and final solutions will be done in subsequent PRs? Those comments help us to avoid mistakes and also give reader (e.g. me) the intention of future changes. Using a consistent format like `TODO(login-id): ` is highly appreciated because we can grep the text to see if we have missed anything.

kiootic:
@louischan-oursky updated, please check again.

louischan-oursky:
Looks good to me so far.

louischan-oursky:
Test plan:

- Signup with default realm
- Login with default realm
- Configure realms
- Signup with non-default realm
- Login with non-default realm

PR #956, opened 2019-07-02
Implement multiple login ID per key and login ID key config

kiootic:
refs #933
depends on #947


pkg/auth/dependency/sso/sso.go
```diff
@@ -97,10 +97,14 @@ type AuthInfo struct {
 	ProviderUserID          string
 	ProviderUserProfile     map[string]interface{}
 	ProviderAccessTokenResp interface{}
-	ProviderAuthData        map[string]string
+	ProviderAuthData        ProviderAuthKeys
 	State                   State
 }
 
+type ProviderAuthKeys struct {
```
louischan-oursky: IMO, `ProviderUserInfo` is a better name.

pkg/auth/handler/signup.go
```diff
@@ -147,6 +148,7 @@ func (h SignupHandler) DecodeRequest(request *http.Request) (handler.RequestPayl
 	if payload.Realm == "" {
 		payload.Realm = password.DefaultRealm
 	}
+	payload.LoginIDs = password.ParseLoginIDs(payload.RawLoginIDs)
```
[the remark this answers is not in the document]
kiootic: `ParseLoginIDs` only parses the raw list of maps into list of key value pairs; all possible input of `RawLoginIDs` is valid for this function. The validation is performed in the `Validate` function of the payload instead.

pkg/auth/dependency/sso/sso.go
kiootic: This struct is intended to store only auth info (e.g. email) for matching with login IDs according to current SSO code. I agree changing it to `ProviderUserInfo` would be better, but then we should refactor the `DecodeXXX` functions in `AuthInfoProcessor` interface into single `DecodeUserInfo` function. We can leave this for later?

louischan-oursky: It is good refactoring! When should we do this refactoring if not now?

pkg/auth/handler/signup.go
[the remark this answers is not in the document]
louischan-oursky: Let's adjust the HTTP API to accept LoginID directly. The client SDK should flexibly accept different kind of shorthands and convert for the server.

pkg/auth/dependency/sso/sso.go
kiootic: Will refactor SSO related code in #957

pkg/auth/handler/userverify/verify_request.go
```diff
@@ -56,16 +59,32 @@ func (f VerifyRequestHandlerFactory) ProvideAuthzPolicy() authz.Policy {
...
+	case loginIDTypeEmail:
+		return provider.CheckLoginIDKeyType(principal.LoginIDKey, metadata.Email)
+	default:
+		return false
 	}
+}
+
+type VerifyRequestPayload struct {
+	RawLoginIDType string      `json:"login_id_type"`
+	LoginIDType    loginIDType `json:"-"`
```
louischan-oursky: `RawLoginIDType` seems to redundant. We have `Validate` so we can validate `LoginIDType loginIDType` before `Handle` can ever be run

PR #957, opened 2019-07-02
Implement welcome email destination config

kiootic:
refs #933
depends on #956


kiootic:
@louischan-oursky refactored OAuth user info in this PR too.

Issue #960, opened 2019-07-08
Get current identity


kiootic:
Return current identity in `/me` API. Add identity ID to session token.

- Skygear Server Date/Version:
- Application Platform:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Steps to reproduce

1.
2.
3.

PR #963, opened 2019-07-10
Add current identity to session information

kiootic:
ref #960


pkg/auth/inject.go
```diff
@@ -185,6 +186,42 @@ func (m DependencyMap) Provide(
...
 			db.NewSafeTxContextWithContext(ctx, tConfig),
 		)
+	case "IdentityProvider":
+		// TODO:
+		// from tConfig
+		passwordHistoryEnabled := tConfig.UserConfig.UserAudit.Password.HistorySize > 0 || tConfig.UserConfig.UserAudit.Password.HistoryDays > 0
+		return principal.NewIdentityProvider(
+			db.NewSQLBuilder("auth", tConfig.AppName),
+			db.NewSQLExecutor(ctx, db.NewContextWithContext(ctx, tConfig)),
+			anonymous.NewSafeProvider(
```
louischan-oursky: Should we reuse the dependent providers instead of creating them again here?

pkg/auth/handler/login.go
```diff
@@ -192,8 +192,7 @@ func (h LoginHandler) Handle(req interface{}) (resp interface{}, err error) {
 	return
 }
 
-func (h LoginHandler) getUserID(pwd string, loginIDKey string, loginID string, realm string) (userID string, err error) {
-	principal := password.Principal{}
+func (h LoginHandler) getPrincipal(pwd string, loginIDKey string, loginID string, realm string) (principal password.Principal, err error) {
```
louischan-oursky: `(principal *password.Principal, err error)`

pkg/auth/handler/sso/customtoken.go
```diff
@@ -241,10 +245,11 @@ func (h CustomTokenLoginHandler) handleLogin(payload customTokenLoginPayload, in
 			return
 		}
 
-		principal := customtoken.NewPrincipal()
+		p := customtoken.NewPrincipal()
+		principal = &p
```
louischan-oursky: Simply `principal = customtoken.NewPrincipal()`

pkg/auth/model/auth_response.go
```diff
@@ -0,0 +1,27 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package model
+
+type AuthResponse struct {
+	User User `json:"user"`
```
louischan-oursky: I think we should keep the original shape by embedding User

pkg/auth/model/user.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"time"
+
+	"github.com/skygeario/skygear-server/pkg/auth/dependency/userprofile"
+	"github.com/skygeario/skygear-server/pkg/core/auth/authinfo"
+)
+
+// User is the unify way of returning a AuthInfo with LoginID to SDK
+type User struct {
+	ID          string           `json:"id,omitempty"`
```
louischan-oursky:
```golang
UserID string `json:"user_id,omitempty"`
```

Because User is going to be embedded. Also retains backward compatibility.

pkg/auth/model/auth_response.go
kiootic: I don't think it's a good idea to embed access token in user object.

pkg/auth/inject.go
kiootic: I see many of existing nested dependencies (PasswordHistoryStore, TemplateEngine) are created individually, so I choose to follow existing convention.

louischan-oursky: Let's change that convention. Repeating the creation is sometimes harmful because we may forget to change all of them.

pkg/auth/model/user_with_identity.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package model
+
+type UserWithIdentity struct {
+	User     User     `json:"user"`
+	Identity Identity `json:"identity"`
+}
```
louischan-oursky: I think we can have a single struct

```golang
type AuthResponse struct {
  User User `json:"user"`
  Identity *Identity `json:"identity,omitempty"`
  AccessToken string `json:"access_token,omitempty"`
}
```

louischan-oursky:
Thanks!

PR #965, opened 2019-07-12
Align the SSO implementation with the spec

louischan-oursky:
The missing piece is to support Azure AD v2. It seems that we have to implement OpenID Connect.


pkg/auth/dependency/principal/customtoken/claims.go
```diff
@@ -5,24 +5,51 @@ import (
 	"github.com/dgrijalva/jwt-go"
 )
 
-type SSOCustomTokenClaims struct {
-	RawProfile map[string]interface{} `json:"skyprofile"`
-	jwt.StandardClaims
+type SSOCustomTokenClaims jwt.MapClaims
```
kiootic: I think it's better to add comment, instead of just commit message, to describe the reason why `MapClaims` is used instead.

pkg/core/config/tenant.go
```diff
@@ -577,15 +671,31 @@ type WelcomeEmailConfiguration struct {
...
+	Secret                     string `json:"secret" yaml:"secret" msg:"secret"`
+	OnUserDuplicateAllowMerge  bool   `json:"on_user_duplicate_allow_merge" yaml:"on_user_duplicate_allow_merge" msg:"on_user_duplicate_allow_merge"`
+	OnUserDuplicateAllowCreate bool   `json:"on_user_duplicate_allow_create" yaml:"on_user_duplicate_allow_create" msg:"on_user_duplicate_allow_create"`
+}
+
+type OAuthConfiguration struct {
+	URLPrefix                      string                       `json:"url_prefix" yaml:"url_prefix" msg:"url_prefix"`
+	JSSDKCDNURL                    string                       `json:"js_sdk_cdn_url" yaml:"js_sdk_cdn_url" msg:"js_sdk_cdn_url"`
+	StateJWTSecret                 string                       `json:"state_jwt_secret" yaml:"state_jwt_secret" msg:"state_jwt_secret"`
+	AllowedCallbackURLs            []string                     `json:"allowed_callback_urls" yaml:"allowed_callback_urls" msg:"allowed_callback_urls"`
```
kiootic: I think `AllowedCallbackURLs` should be required; otherwise, `AuthHandler.validateCallbackURL` would allow any callback URL, which would be a major security problem.

pkg/core/config/tenant.go
```diff
@@ -577,15 +671,31 @@ type WelcomeEmailConfiguration struct {
...
+	Audience                   string `json:"audience" yaml:"audience" msg:"audience"`
+	Secret                     string `json:"secret" yaml:"secret" msg:"secret"`
+	OnUserDuplicateAllowMerge  bool   `json:"on_user_duplicate_allow_merge" yaml:"on_user_duplicate_allow_merge" msg:"on_user_duplicate_allow_merge"`
+	OnUserDuplicateAllowCreate bool   `json:"on_user_duplicate_allow_create" yaml:"on_user_duplicate_allow_create" msg:"on_user_duplicate_allow_create"`
+}
+
+type OAuthConfiguration struct {
+	URLPrefix                      string                       `json:"url_prefix" yaml:"url_prefix" msg:"url_prefix"`
+	JSSDKCDNURL                    string                       `json:"js_sdk_cdn_url" yaml:"js_sdk_cdn_url" msg:"js_sdk_cdn_url"`
+	StateJWTSecret                 string                       `json:"state_jwt_secret" yaml:"state_jwt_secret" msg:"state_jwt_secret"`
```
kiootic: I think `StateJWTSecret` should be required (or default to master key); otherwise, the OAuth state param would be signed with empty string.

pkg/auth/handler/sso/auth_handler_test.go
```diff
@@ -151,62 +166,35 @@ func TestAuthHandler(t *testing.T) {
...
-					"identity": {
-						"id": "%s",
-						"type": "oauth",
-						"provider_id": "mock",
-						"provider_user_id": "mock_user_id",
-						"raw_profile": {},
-						"claims": {}
-					},
-					"access_token": "%s"
+					"result": {
```
kiootic: Is the double nested results intentional?

louischan-oursky: Yes. The first level is `callback` and `result`. This `result` denotes the http response body. The second level is either `result` or `error` which is our conventional API response shape.

louischan-oursky:
@kiootic Comments addressed.

louischan-oursky:
Superseded by https://github.com/SkygearIO/skygear-server/pull/967

PR #967, opened 2019-07-16
Azure AD v2 OpenID Connect

louischan-oursky:


pkg/auth/handler/sso/auth_url.go
```diff
@@ -117,51 +122,125 @@ func (p AuthURLRequestPayload) Validate() error {
...
+	if cookieErr == http.ErrNoCookie {
+		// Set nonce
+		nonce := sso.GenerateOpenIDConnectNonce()
+		cookie = &http.Cookie{
+			Name:     coreHttp.CookieNameOpenIDConnectNonce,
+			Value:    nonce,
+			Path:     "/",
+			HttpOnly: true,
+		}
+		http.SetCookie(w, cookie)
```
kiootic: So we expect all SDK would have a cookie jar implemented to use OIDC providers?

pkg/auth/dependency/sso/sso.go
```diff
@@ -105,11 +105,48 @@ type ProviderUserInfo struct {
...
+	// Nonce is required when the underlying provider is OpenID connect compliant.
+	// The implementation is based on the suggestion in the spec.
+	// See https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes
+	//
+	// The nonce is a cryptographically random string.
+	// The nonce passed to the provider is SHA256 hash of it.
+	// The get-authorization URL endpoint ensures the nonce in session cookie.
+	// The callback endpoint expect the user agent to include the nonce in session cookie.
+	// The nonce in session cookie will be validated against the hashed nonce in the ID token.
+	Nonce string
```
kiootic: We should also validate nonce on pure OAuth2 flow. ref: https://tools.ietf.org/html/draft-ietf-oauth-v2-threatmodel-06#section-5.3.5

pkg/auth/handler/sso/customtoken.go
```diff
@@ -247,57 +311,105 @@ func (h CustomTokenLoginHandler) handleLogin(
...
+	// We need to consider password principal
+	passwordPrincipal, err := h.findExistingPasswordPrincipal(payload.Claims.Email(), payload.MergeRealm)
+	if err != nil {
 		return
 	}
 
+	// Case: Custom Token principal was not found and Password principal was not found
+	// => Simple create case
+	if passwordPrincipal == nil {
+		createFunc()
```
kiootic: We should also send welcome email when creating user through SSO.

pkg/auth/dependency/sso/google.go
```diff
@@ -1,58 +1,73 @@
...
+	OAuthConfig    config.OAuthConfiguration
+	ProviderConfig config.OAuthProviderConfiguration
 }
 
 func (f *GoogleImpl) GetAuthURL(params GetURLParams) (string, error) {
-	if f.Config.ClientID == "" {
-		skyErr := skyerr.NewError(skyerr.InvalidArgument, "ClientID is required")
-		return "", skyErr
-	}
 	params.Options["access_type"] = "offline"
```
kiootic: I think we should use `online` at the moment; we do not need refresh tokens.

pkg/auth/handler/sso/auth_url.go
louischan-oursky: A simple memory cookie jar is enough for this flow. Cookie is part of the HTTP spec and it does not require any special support from our SDK.

pkg/auth/dependency/sso/sso.go
louischan-oursky: It requires the provider support. For now, we just validate nonce for OIDC provider.

pkg/auth/dependency/sso/google.go
louischan-oursky: I didn't change that part of code. Though I agree we can use `online`.

louischan-oursky:
- [ ] Remove `sso.Options`
- [ ] Always generate nonce to ensure it is unpredictable.

louischan-oursky:
@kiootic Do not merge this yet. I realized there are some remaining tasks

- [ ] Populate Principal Claims

louischan-oursky:
@kiootic Done.

pkg/auth/dependency/sso/userinfo.go
```diff
@@ -0,0 +1,80 @@
...
+
+	info.ID, _ = data["id"].(string)
+	info.Email, _ = data["email"].(string)
+	return
+}
+
+type Azureadv2UserInfoDecoder struct{}
+
+func NewAzureadv2UserInfoDecoder() Azureadv2UserInfoDecoder {
+	return Azureadv2UserInfoDecoder{}
```
kiootic: I think it should be AzureADv2UserInfoDecoder?

louischan-oursky: You're right!

PR #982, opened 2019-08-01
Inline SSO javascript

louischan-oursky:


pkg/auth/dependency/sso/auth_handler_html.go
```diff
@@ -21,28 +19,98 @@ func (i *AuthHandlerHTMLProvider) HTML() (out string, err error) {
...
+
+function cookieJarDelete(name) {
+	document.cookie = name + "=;expires=Thu, 01 Jan 1970 00:00:01 GMT;";
+}
+
+function StringStartsWith(s, search) {
+	return s.substring(0, search.length) === search;
+}
+
+function validateCallbackURL(callbackURL, authorizedURLs) {
```
kiootic: This logic does not seems matches what server does?

pkg/auth/dependency/sso/auth_handler_html.go
```diff
@@ -21,28 +19,98 @@ func (i *AuthHandlerHTMLProvider) HTML() (out string, err error) {
...
+	if (error) {
+		windowObject.postMessage({
+			type: "error",
+			error: error
+		}, "*");
+	} else {
+		windowObject.postMessage({
+			type: "result",
+			result: result
+		}, callbackURL);
```
kiootic: I think postMessage takes allowed origin, not a URL?

pkg/auth/dependency/sso/auth_handler_html.go
louischan-oursky: This is the original code.

pkg/auth/dependency/sso/auth_handler_html.go
louischan-oursky: Also original code.

pkg/auth/dependency/sso/auth_handler_html.go
kiootic: Do we fix it to match server logic now or later?

louischan-oursky: I am going to make the two implementation the same.

pkg/auth/dependency/sso/auth_handler_html.go
louischan-oursky: https://gist.github.com/louischan-oursky/bcab4ce06c30a47f775194505f7f1aa9

Tested on Chrome.

louischan-oursky:
@kiootic 

`validateCallbackURL` was made the same across Golang and JavaScript.

`window.postMessage` accepts URL too.

PR #983, opened 2019-08-01
Use token in authorization header if present

kiootic:


pkg/core/model/app.go
```diff
@@ -59,6 +60,27 @@ func CheckAccessKeyType(config config.TenantConfiguration, apiKey string) KeyTyp
...
+
+	scheme := authorization[0]
+	if strings.ToLower(scheme) != httpAuthzBearerScheme {
+		return
+	}
+
+	return authorization[1]
+}
+
 func GetAccessToken(r *http.Request) string {
```
louischan-oursky: I wonder why this function is in this file and this package...

pkg/core/model/app_test.go
```diff
@@ -0,0 +1,33 @@
...
+			req.Header.Add(httpHeaderAuthorization, "")
+			So(GetAccessToken(req), ShouldEqual, "access-token")
+		})
+		Convey("should return value of authorization header", func() {
+			req, _ := http.NewRequest("", "", nil)
+			req.Header.Add(corehttp.HeaderAccessToken, "")
+			req.Header.Add(httpHeaderAuthorization, "Bearer bearer-token")
+			So(GetAccessToken(req), ShouldEqual, "bearer-token")
+		})
+		Convey("should prioritize authorization header", func() {
```
louischan-oursky: Not sure if we should prioritize `authorization` header over `x-skygear-access-token`

cc @carmenlau

Issue #989, opened 2019-08-06
Generate HTTP doc


tensiuyan:

louischan-oursky:
Please investigate build failure https://travis-ci.org/SkygearIO/skygear-server/builds/574152208?utm_source=slack&utm_medium=notification

@kiootic

PR #992, opened 2019-08-06
Add API doc

kiootic:
ref #989

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,6 +153,10 @@ docker-push-version:
 release-commit:
 	./scripts/release-commit.sh
 
+.PHONY: preview-doc-auth
+preview-doc-auth:
+	./scripts/preview-doc.sh auth
+
 .PHONY: generate-doc-auth
 generate-doc-auth:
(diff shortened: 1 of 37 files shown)


scripts/generate-doc.sh
```diff
@@ -0,0 +1,27 @@
...
+git reset --hard
+
+mkdir -p apis/auth
+openapi3-gen -dir "$PROJ_ROOT" -output "apis/auth/$VERSION.yaml" ./cmd/auth/... ./pkg/auth/...
+
+URL="https://generator.swagger.io/?url=https://skygeario.github.io/skygear-server/apis/auth/$VERSION.yaml"
+sed "s|{URL}|$URL|g" "$PROJ_ROOT/scripts/template/index.html" > index.html
+
+git add .
+git commit -m "Update documentation for $VERSION" || true
```
louischan-oursky: Why `|| true`?

kiootic: The commit would fail if nothing changed. On a second thought, I should just notify the user about it.

PR #999, opened 2019-08-14
Validate user config

louischan-oursky:


louischan-oursky:
@carmenlau 

Added error printing to `cmd/auth/main.go`

Issue #978, opened 2019-07-26
Fix host header handling


kiootic:
- Gateway should put real `Host` in `X-Forwarded-Host` header
    - Also other `X-Forwarded-` headers?
    - `x-forwarded-for` is [handled](https://golang.org/src/net/http/httputil/reverseproxy.go?s=488:2518#L251)
- Auth gear should check the `Host` / `X-Forwarded-Host` header
    - In standalone mode:
        - Add config field (expected host), reject request if mismatched.
    - For web-hook:
        - Use `X-Forwarded-Host` / `Host` header to generate web-hook delivery URL.

PR #1001, opened 2019-08-15
Handle relative web-hook URL

kiootic:
ref #978

### DIFF
--- a/cmd/auth/main.go
+++ b/cmd/auth/main.go
@@ -37,6 +37,7 @@ type configuration struct {
 	StandaloneTenantConfigurationFile string `envconfig:"STANDALONE_TENANT_CONFIG_FILE" default:"standalone-tenant-config.yaml"`
 	PathPrefix                        string `envconfig:"PATH_PREFIX"`
 	Host                              string `default:"localhost:3000"`
+	ValidHosts                        string `envconfig:"VALID_HOSTS"`
 }
 
 /*
@@ -75,6 +76,9 @@ func main() {
 
 	configuration := configuration{}
 	envconfig.Process("", &configuration)
+	if configuration.ValidHosts == "" {
+		configuration.ValidHosts = configuration.Host
+	}
 
 	// default template initialization
 	templateEngine := template.NewEngine()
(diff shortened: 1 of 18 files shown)


cmd/auth/main.go
```diff
@@ -35,6 +35,7 @@ type configuration struct {
 	StandaloneTenantConfigurationFile string `envconfig:"STANDALONE_TENANT_CONFIG_FILE" default:"standalone-tenant-config.yaml"`
 	PathPrefix                        string `envconfig:"PATH_PREFIX"`
 	Host                              string `default:"localhost:3000"`
+	ValidateHosts                     string `envconfig:"VALIDATE_HOSTS"`
```
louischan-oursky: Maybe ValidHosts is a better name. Also rename the environment variable.

pkg/core/middleware/validate_host.go
```diff
@@ -0,0 +1,37 @@
...
+	if len(validateHosts) == 1 && validateHosts[0] == "" {
+		// skip validation if no host is provided
+		return next
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		host := coreHttp.GetHost(r)
+		if !utils.StringSliceContains(validateHosts, host) {
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte{})
```
louischan-oursky: return

pkg/auth/dependency/hook/provider.go
```diff
@@ -1,12 +1,15 @@
...
 import (
+	"net/url"
+
 	"github.com/skygeario/skygear-server/pkg/auth/event"
 	"github.com/skygeario/skygear-server/pkg/auth/model"
 	"github.com/skygeario/skygear-server/pkg/core/handler"
 )
 
 type Provider interface {
+	SetBaseURL(*url.URL)
```
louischan-oursky: It seems that it is possible to avoid this method. In `pkg/core/inject/inject.go`, `RequestDependencyMap.Provide` has access to `*http.Request` but `DependencyMap.Provide` hides it. Maybe we can remove such hiding.

pkg/auth/dependency/hook/provider_impl.go
```diff
@@ -42,6 +49,10 @@ func NewProvider(
 	}
 }
 
+func (provider *providerImpl) SetBaseURL(url *url.URL) {
```
louischan-oursky: No longer needed?

PR #1003, opened 2019-08-19
Add gh-pages CI deploy script

kiootic:


scripts/deploy-gh-pages.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh -eu
+
+scripts/update-gh-pages.sh
+git push https://skygear-bot:$GITHUB_TOKEN@github.com/SkygearIO/skygear-server.git gh-pages > /dev/null 2>/dev/null
```
louischan-oursky: Why do we pipe stdout and stderr to /dev/null? I think keeping them make debugging easier.

kiootic: To prevent leaking the token accidentally.

PR #1008, opened 2019-08-19
Validate raw user config

louischan-oursky:


pkg/gateway/store/pq/app.go
```diff
@@ -79,8 +79,12 @@ func (s *Store) getConfigByID(id string, configValue *config.TenantConfiguration
 	if err == sql.ErrNoRows {
 		return ErrConfigNotFound
 	}
+	if err != nil {
+		return err
+	}
 
-	return err
+	configValue.AfterUnmarshal()
```
carmenlau: It is not ideal to call `AfterUnmarshal` outside the config model, we may miss it in future updates. After discussion, we will put this back to the `Scan` function in `TenantConfiguration` model. Only controller want to bypass the `AfterUnmarshal` to get the raw config for editing, controller should has its own struct that wrap the `TenantConfiguration` and implement `Scan` that bypass `AfterUnmarshal`.

louischan-oursky:
@carmenlau comment addressed.

Issue #1009, opened 2019-08-20
Reorganize multi-tenant database structure


kiootic:
To support connection pooling in for multi-tenant database, discriminator based multi-tenancy would be used. A single database/schema would be used for all tenants, and all tables would have a `tenant_id` column to identify owner tenant. All database access in auth gear should filter on `tenant_id`.

TODO:
- [ ] Update tenant config
- [ ] Add DB migrations
- [ ] De-dupe DB source in migrate cmd
- [ ] Add connection pool

PR #1013, opened 2019-08-21
Use discriminator column for multi tenancy

kiootic:
ref #1009


migrate/cmd/migrate/main.go
```diff
@@ -212,23 +212,18 @@ func runCmd(req commandRequest) (string, error) {
 		sourceURL = fmt.Sprintf("file://%s", sourceURL)
 	}
 
-	schema := req.Schema
```
louischan-oursky: I think this is still relevant

pkg/core/db/pool.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/jmoiron/sqlx"
+	"github.com/skygeario/skygear-server/pkg/core/config"
+)
+
+type Pool interface {
+	Open(tConfig config.TenantConfiguration) (*sqlx.DB, error)
+	Dispose() error
```
louischan-oursky: Why not just call this `Close() error`?

pkg/core/db/pool.go
```diff
@@ -69,3 +70,16 @@ func (p *poolImpl) Dispose() (err error) {
...
 	return
 }
+
+func openPostgresDB(url string) (db *sqlx.DB, err error) {
+	db, err = sqlx.Open("postgres", url)
+	if err != nil {
+		return
+	}
+
+	// TODO: configurable / profile for good value?
```
louischan-oursky: `TODO(pool)` for easier search.

migrations/core/1566290052_discrimator-multi-tenancy.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE _core_user ADD COLUMN app_id TEXT;
+UPDATE _core_user SET app_id = '';
+ALTER TABLE _core_user ALTER COLUMN app_id SET NOT NULL;
+ALTER TABLE _core_user ADD CONSTRAINT _core_user_id_idx UNIQUE(app_id, id)
```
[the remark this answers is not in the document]
kiootic: I think keep using `(id)` as primary key is preferable: this can ensure IDs would not be valid across apps, in case bugs occurred in queries.

I'll change the `(app_id, id)` unique constraint to a simple index.

pkg/core/db/pool.go
kiootic: No particular reason. If you prefer to name it `Close` I'll change it.

migrate/cmd/migrate/main.go
kiootic: This variable exists so that the following conditional can force schema to be `app_config` when gateway migration is running. It seems strange to do it, and I don't think it is needed?

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/franela/goreq v0.0.0-20170418064916-b5b0f5eb2d16
 	github.com/go-gomail/gomail v0.0.0-20150902115704-41f357289737
 	github.com/go-sql-driver/mysql v1.4.1 // indirect
+	github.com/golang-migrate/migrate v3.5.4+incompatible // indirect
```
carmenlau: This is dependency of the migrate program. Since the dependency of migrate program and the server programs are too different, we moved the migrate program into folder `migrate` and has it own go.mod in that folder. Please remove it here, and run `migrate.go` in the `migrate` folder.

kiootic:
@louischan-oursky @carmenlau updated, please review again.

migrations/core/1566290052_discrimator-multi-tenancy.up.sql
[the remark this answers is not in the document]
louischan-oursky: Maybe the index can just be `(app_id)`?

migrate/pkg/migrate/tenant/tenant.go
```diff
@@ -10,14 +10,15 @@ import (
...
+	Name           string
+	DatabaseURL    string
+	DatabaseSchema string
 }
 
 func GetMigrateApps(hostnameOverride string, db sq.BaseRunner, appFilterKey string, appFilterValue string) ([]App, error) {
 	// get apps for migration
 	configSchema := "app_config"
-	sqlizer := sq.Select("app.name", "config.config #> '{app_config,database_url}'").
+	sqlizer := sq.Select("app.name", "config.config #> '{app_config,database_url,database_schema}'").
```
louischan-oursky: This seems incorrect. `column #> '{a,b}'` roughly means the JSON expression `column["a"]["b"]`. `database_schema` is not a property of `database_url`. I guess what we want here is

```go
sq.Select("app.name",
"config.config #> '{app_config,database_url}'",
"config.config #> '{app_config,database_schema}'")
```

This will select 3 columns.

kiootic:

louischan-oursky:
@kiootic conflicted.

Issue #1010, opened 2019-08-20
Revamp session mechanism


kiootic:
- [ ] Use Redis as stateful session store
- [ ] Configure API Key per client
- [ ] Allow cookies as session transport
- [ ] Support refresh tokens
- [ ] Frequently accessed data handling
- [ ] Session management API

PR #1017, opened 2019-08-27
Use Redis as session store

kiootic:
ref #1010


pkg/core/redis/context.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+
+	"github.com/gomodule/redigo/redis"
+)
+
+type contextKeyType struct{}
+
+var contextKey = contextKeyType{}
+
+type Context interface {
```
louischan-oursky: Seems unused in this commit.

pkg/core/auth/session/redis/store.go
```diff
@@ -0,0 +1,70 @@
...
+func (s *store) Create(sess *session.Session) (err error) {
+	json, err := json.Marshal(sess)
+	if err != nil {
+		return
+	}
+	conn := redis.GetConn(s.ctx)
+	key := sessionKey(s.appID, sess.ID)
+	_, err = goredis.String(conn.Do("SET", key, json, "NX"))
+	if err == goredis.ErrNil {
+		err = fmt.Errorf("cannot create session")
```
louischan-oursky: Make this error constant?

pkg/core/auth/session/provider_impl_test.go
```diff
@@ -0,0 +1,152 @@
...
+				CreatedAt:            initialTime,
+				AccessedAt:           initialTime,
+				AccessToken:          "access-token",
+				AccessTokenCreatedAt: initialTime,
+			}
+
+			Convey("should be successful", func() {
+				err := provider.Invalidate("session-id")
+				So(err, ShouldBeNil)
+				So(store.Sessions, ShouldBeEmpty)
```
louischan-oursky: Get once before invalidation and assert the session is there.
Get once after invalidation and assert the session is gone.

pkg/core/auth/testing/context.go
```diff
@@ -0,0 +1,59 @@
+package testing
+
+import (
+	"github.com/skygeario/skygear-server/pkg/core/auth"
+	authinfo "github.com/skygeario/skygear-server/pkg/core/auth/authinfo"
+	authtoken "github.com/skygeario/skygear-server/pkg/core/auth/authtoken"
+	model "github.com/skygeario/skygear-server/pkg/core/model"
+)
+
+type MockContext struct {
```
louischan-oursky: This new API looks nice!

docker-compose.yml
```diff
@@ -13,6 +13,10 @@ services:
     - db_data:/var/lib/postgresql/data
     ports:
     - "5432:5432"
+  redis:
+    image: redis:5.0
+    ports:
+      - "6379:6379"
```
louischan-oursky: Make the data of this redis persistent.

pkg/core/redis/context.go
kiootic: It's some old prototyping code, will remove it.

pkg/core/auth/session/provider_impl_test.go
kiootic: I don't think it is needed? The parent block setup one session in the fake store, and each block assert the expected number of sessions in fake store after the operation.

PR #1018, opened 2019-08-28
Implement per client API key

kiootic:
ref #1010 
depends on #1017

### DIFF
--- a/pkg/core/auth/session/redis/store.go
+++ b/pkg/core/auth/session/redis/store.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	goredis "github.com/gomodule/redigo/redis"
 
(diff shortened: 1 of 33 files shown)


pkg/core/validation/schema.go
```diff
@@ -38,7 +43,26 @@ const (
...
+					"type": "string",
+					"enum": ["header"]
+				},
+				"access_token_lifetime": { "type": "integer", "minimum": 0 },
+				"session_idle_timeout_enabled": { "type": "boolean" },
+				"session_idle_timeout": { "type": "integer", "minimum": 0 },
+				"refresh_token_disabled": { "type": "boolean" },
+				"refresh_token_lifetime": { "type": "integer", "minimum": 0 }
+			},
+			"required": ["name", "disabled", "api_key", "session_transport"]
```
louischan-oursky: `disabled` should be optional

pkg/core/middleware/tenant.go
```diff
@@ -29,11 +30,15 @@ func (m TenantConfigurationMiddleware) Handle(next http.Handler) http.Handler {
 			panic(fmt.Errorf("Unable to retrieve configuration: %v", err.Error()))
 		}
 
+		r = auth.InitRequestAuthContext(r)
```
louischan-oursky: We also have this line in `server/server.go`. Can we call this twice?

pkg/core/auth/session/redis/store.go
```diff
@@ -70,3 +70,7 @@ func (s *store) Delete(id string) (err error) {
 	_, err = conn.Do("DEL", key)
 	return
 }
+
+func toMilliseconds(d time.Duration) int64 {
+	return d.Nanoseconds() / 1000000
```
louischan-oursky: Should we follow the implementation in stdlib? https://golang.org/src/time/time.go?s=25703:25738#L793

pkg/core/middleware/tenant.go
kiootic: In standalone mode, the init in `core/server` would override the init in `middleware/tenant`, i.e. the API key set in `middleware/tenant` would be ignored. However, `core/server` would resolve the API key in header, set by `middleware/tenant`, so it would still work correctly.

In a later PR (#1022), the authn part (i.e. resolving API key & session) would be refactored to be cleaner.

pkg/core/auth/session/redis/store.go
kiootic: I'll change it to `return int64(d / time.Millisecond)`. Fractional part would be ignored, since Redis can't handle it and the precision is enough.

kiootic:

pkg/core/middleware/tenant.go
louischan-oursky: Can we add the above note in the code?

kiootic: Added.

PR #1019, opened 2019-08-28
Implement session cookies

kiootic:
ref #1010 
depends on #1018


pkg/gateway/middleware/authinfo.go
```diff
@@ -39,6 +40,24 @@ func (m *AuthInfoMiddleware) Handle(next http.Handler) http.Handler {
...
+					Path:    "/",
+					Expires: time.Unix(0, 0),
+				}
+				http.SetCookie(w, cookie)
+
+				skyErr := skyerr.NewNotAuthenticatedErr()
+				httpStatus := skyerr.ErrorDefaultStatusCode(skyErr)
+				response := struct {
+					Err skyerr.Error `json:"error,omitempty"`
+				}{Err: skyErr}
```
louischan-oursky: Can we reuse the struct `APIResponse`?

kiootic: Yes, updated.

louischan-oursky:
In auth gear standalone mode, If the client is `header` and the incoming request has only cookie set, the server will return 401 without clearing the cookie. Is this intended? I see mismatch transport is handled by the gateway but if the auth gear is in standalone mode, we ignore the error.

kiootic:
> In auth gear standalone mode, If the client is header and the incoming request has only cookie set, the server will return 401 without clearing the cookie. Is this intended?

Yes it is intended. Gateway / standalone mode gears should clear cookie only if token is present in both header & cookie. I'll change it so that it'll also clear cookie when logging in/out if unexpected cookie is present.

> I see mismatch transport is handled by the gateway but if the auth gear is in standalone mode, we ignore the error.

I'll make the access token handling consistent by refactoring it into a middleware.

pkg/core/middleware/authn.go
```diff
@@ -0,0 +1,131 @@
...
+			// invalid session token -> must not proceed
+
+			if err == model.ErrTokenConflict {
+				// clear session cookie if session token is conflicted
+				cookie := &http.Cookie{
+					Name:    coreHttp.CookieNameSession,
+					Path:    "/",
+					Expires: time.Unix(0, 0),
+				}
+				http.SetCookie(w, cookie)
```
louischan-oursky: This part seems redundant because in the `if err != nil` block, we never assign `err` to `nil` so the above defer block will clear the cookie for us.

pkg/core/middleware/authn.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		if err = m.TxContext.BeginTx(); err != nil {
+			panic(err)
+		}
+		defer m.TxContext.RollbackTx()
+
+		s, err := m.SessionProvider.GetByToken(accessToken, auth.SessionTokenKindAccessToken)
+		if err != nil {
+			// session not found -> treat as no access token is provided
```
louischan-oursky: Example of such case: the access token expires, so session is not found. The client should be notified of this fact and act accordingly. The server should clear the cookie and the client should clear local authn state.

pkg/core/middleware/authn.go
```diff
@@ -0,0 +1,131 @@
...
+			// inconsistent session token transport -> must not proceed
+			err = session.ErrSessionNotFound
+			return
+		}
+
+		authInfo := authinfo.AuthInfo{}
+		err = m.AuthInfoStore.GetAuth(s.UserID, &authInfo)
+		if err != nil {
+			if err == skydb.ErrUserNotFound {
+				// user not found -> treat as no access token is provided
```
louischan-oursky: Maybe we should treat user not found as session not found so that the cookie is cleared and the client local authn state is also cleared.

pkg/core/server/server.go
```diff
@@ -100,12 +98,31 @@ func NewServerWithOption(
...
-		configuration := config.GetTenantConfig(r)
-
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		// TODO: improve the logger
 		log := logging.CreateLoggerWithContext(r.Context(), "server")
 
+		policy := hf.ProvideAuthzPolicy()
+		if err := policy.IsAllowed(r, auth.NewContextGetterWithContext(r.Context())); err != nil {
+			// TODO: log
+			log.WithError(err).Error("authz not allowed")
```
louischan-oursky: Reduce the log level to `Info` because it is not an unexpected error.

pkg/core/middleware/authn.go
kiootic: Considering calling a public API (i.e. authentication not required), I don't think an expired access token should fail the request?

louischan-oursky: But for endpoint requiring authentication, the client should be notified the session is expired and clean up state. The problem here is that this middleware does not know the whether the underlying handler requires authentication or not.

One solution is to let authn middleware to save the resolution result in request context. The result is `NoSession`, `InvalidSession`, `ValidSession(user)`. Then the authz middleware can make use of this piece information and act accordingly. However, the approach does not work for microservice because we do not have authz for microservice currently. So application must call `/me` on application launch to ensure its authentication state is in sync with the auth gear.

kiootic:
As discussed offline:
- Authn middleware should never return errors; errors should be emitted by gears/microservices
- There is no need to differentiate between invalid session and no session; invalid session will be treated as no session.
- When unexpected cookie found (invalid token/unexpected transport/...), the session cookie would be cleared by gateway (for microservice) or gear.

pkg/core/auth/session/writer_impl.go
```diff
@@ -56,20 +57,38 @@ func (w *writerImpl) WriteSession(rw http.ResponseWriter, accessToken *string) {
...
+		Path:     "/",
+		HttpOnly: true,
+		Secure:   !w.useInsecureCookie,
+		Expires:  time.Unix(0, 0),
+	}
+	updateCookie(rw, cookie)
+}
+
+func updateCookie(rw http.ResponseWriter, cookie *http.Cookie) {
+	headers := rw.Header()
```
louischan-oursky: Can we just ask the stdlib to parse for us? https://golang.org/pkg/net/http/#Response.Cookies

The idea is to call Response.Cookies to get parsed cookies. Update the target. Remove all `Set-Cookie` headers from the map. Set the updated cookies with `http.SetCookie` again. The advantage is we do not need to parse the serialized cookie header value ourselves.

kiootic: I had considered this before, but it feels quite hacky to me: we'll need to make a dummy `http.Response` with the header set to use its `Cookies` method. If you feel that it is better than a manual parsing (which is extracted from stdlib), I'll change it to use that way.

louischan-oursky: Let's create a dummy Response and call its Cookies method

louischan-oursky:
In standalone mode, the authn middleware get the tenant config from request. But the authn middleware comes before the tenant config middleware so it cannot get tenant config and crash.

kiootic:

PR #1032, opened 2019-09-10
MFA

louischan-oursky:


pkg/auth/dependency/mfa/provider_impl.go
```diff
@@ -35,7 +35,11 @@ func (p *providerImpl) GenerateRecoveryCode(userID string) ([]string, error) {
 }
 
 func (p *providerImpl) ListAuthenticators(userID string) ([]interface{}, error) {
-	return p.store.ListAuthenticators(userID)
+	authenticators, err := p.store.ListAuthenticators(userID)
+	if err != nil {
+		return nil, err
+	}
+	return MaskAuthenticators(authenticators), nil
```
kiootic: Not sure if it's good to mask authenticators at provider layer: in future, there may be some admin APIs would return the authenticator list without masking.

pkg/auth/dependency/mfa/authenticator.go
```diff
@@ -113,3 +114,60 @@ func MaskAuthenticators(authenticators []interface{}) []interface{} {
...
+			default:
+				panic("unknown OOB authenticator channel")
+			}
+		default:
+			panic("unknown authenticator")
+		}
+	}
+
+	// Simulate the count if new one is added.
+	totalCount++
```
kiootic: Can we extract this switch as a local function? The dupe code in the loop doesn't look good...

pkg/auth/handler/mfa/create_totp.go
```diff
@@ -0,0 +1,137 @@
...
+				"secret": { "type": "string" }
+			}
+		}
+	}
+}
+`
+
+/*
+	@Operation POST /mfa/totp/new - Create TOTP authenticator.
+		Create TOTP authenticator. It must be activated.
```
kiootic: Not sure what you mean by `It must be activated`?

pkg/auth/dependency/mfa/pq/store.go
```diff
@@ -308,6 +317,51 @@ func (s *storeImpl) CreateTOTP(a *mfa.TOTPAuthenticator) error {
...
+	if err != nil {
+		if err == sql.ErrNoRows {
+			err = mfa.ErrAuthenticatorNotFound
+		}
+		return nil, err
+	}
+	return &a, nil
+}
+
+func (s *storeImpl) UpdateTOTP(a *mfa.TOTPAuthenticator) error {
```
kiootic: Personally I'll make it takes 3 arguments (id, activated, activatedAt), since it does not update other fields in the authenticator.

pkg/auth/dependency/mfa/authenticator.go
```diff
@@ -171,3 +171,42 @@ func CanAddAuthenticator(authenticators []interface{}, newA interface{}, mfaConf
 
 	return true
 }
+
+func IsDeletingLastActivatedAuthenticator(authenticators []interface{}, a interface{}) bool {
```
kiootic: I think naming it `IsDeletingOnlyActivatedAuthenticator` would be better. My first impression is that it means the authenticator activated the latest ('last activated')

pkg/auth/dependency/mfa/pq/store.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var ids []string
+	for _, a := range old {
+		ids = append(ids, a.ID)
+	}
+
+	if len(ids) > 0 {
+		q1 := s.sqlBuilder.Tenant().
```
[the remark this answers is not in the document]
louischan-oursky: Did not think of `DELETE CASCADE` when I wrote this. But it is quite dangerous operation so I prefer we keep the origin approach.

pkg/core/auth/session.go
```diff
@@ -18,6 +18,10 @@ type Session struct {
 	UserID      string `json:"user_id"`
 	PrincipalID string `json:"principal_id"`
 
+	AuthenticatorID         string                  `json:"authenticator_id,omitempty"`
```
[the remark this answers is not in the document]
louischan-oursky: In a later change, Principal has multiple fields. I guess your comment also applies there.

I am worried that if they are put in a single struct, we will have two places having `identity` key but the shape is different.

pkg/core/phone/phone.go
```diff
@@ -0,0 +1,52 @@
...
+	"strings"
+
+	"github.com/nyaruka/phonenumbers"
+)
+
+// ErrNotInE164Format means the given phone number is not in E.164 format.
+var ErrNotInE164Format = errors.New("not in E.164 format")
+
+// Parse ensures the given phone is in E.164 format.
+func Parse(phone string) error {
```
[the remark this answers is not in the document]
louischan-oursky: What about `EnsureE164`?

pkg/auth/dependency/mfa/provider_impl.go
louischan-oursky: In that case, I will remove `ListAuthenticator` and add `ListRawAuthenticator` and `ListMaskedAuthenticator`. For now, there is no use case of listing raw authenticator so I prefer we keep it.

pkg/auth/dependency/mfa/pq/store.go
louischan-oursky: Taking the whole struct seems making the API more pretty. We have Create to return a pointer to the struct, Update to take a pointer and Delete to take a pointer.

pkg/core/mail/mask.go
```diff
@@ -0,0 +1,42 @@
...
+
+	var buf strings.Builder
+	for i := 0; i < length; i++ {
+		if i < halfLength {
+			buf.WriteRune(runes[i])
+		} else {
+			buf.WriteRune('*')
+		}
+	}
+	buf.WriteRune('@')
```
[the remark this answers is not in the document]
louischan-oursky: https://golang.org/src/net/mail/message.go?s=5217:5250#L216

It writes `@` unconditionally. Which part are you referring to?

pkg/core/auth/session.go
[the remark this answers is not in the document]
kiootic: This session struct is for internal use, so it should not matters. If we want to expose it in API, we can still flatten the nested structs.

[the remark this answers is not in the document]
louischan-oursky: I do expose the fields in `model.Session` the the fields are now flatten. I prefer keep them similar

louischan-oursky:
@kiootic All indisputable comments were addressed.

pkg/core/auth/session.go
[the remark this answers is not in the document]
kiootic: It's okay then. It's just a personal preference that group fields that must be changed together into single struct.

pkg/auth/dependency/mfa/pq/store.go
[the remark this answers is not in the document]
kiootic: I think it's better to use `ON DELETE CASCADE` than deleting the related records manually. It would be safer to maintain referential integrity through relational mechanism than maintain it manually.

pkg/core/mail/mask.go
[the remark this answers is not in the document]
kiootic: I was thinking if this function receive `"test"`, without the '@' character, it should returns `"te**"`, even if it may not be a valid email address. However after looking into the details, seems this function rejects invalid email address by returning empty string, and the passed in email should be valid email already (read from database). So just ignore my comment.

pkg/auth/dependency/mfa/pq/store.go
kiootic: I'm worried that later on people may see `UpdateTOTP` and use it, thinking that it would update other fields (e.g. display name). Especially when godoc just states `UpdateTOTP updates TOTP authenticator`, it is misleading.

louischan-oursky: Let me update the documentation to better describe the behavior.

pkg/auth/dependency/mfa/pq/store.go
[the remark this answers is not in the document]
louischan-oursky: I have an opposite opinion. Deleting manually states the relation in the code. If we add additional relation in future and cascading delete does not apply, then cascading delete is dangerous. Manual delete will save us because the foreign key constraint will trigger error.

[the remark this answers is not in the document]
kiootic: I agree that cascade should not be used if the foreign key is used for ownership (e.g. app created_by -> user id). However, this scenario is inheritance/composition (i.e. user contains some recovery code authenticator; recovery code authenticator is composed of single recovery code), where the referencing row is logically the same entity as referenced row. I think using cascade delete is appropriate in this scenario.

pkg/auth/dependency/authnsession/mock_provider.go
```diff
@@ -0,0 +1,43 @@
...
+) Provider {
+	authContext := authTesting.NewMockContext()
+	mfaConfiguration := config.MFAConfiguration{
+		Enforcement: config.MFAEnforcementOff,
+	}
+	authenticationSessionConfiguration :=
+		config.AuthenticationSessionConfiguration{
+			Secret: "authnsessionsecret",
+		}
+	return NewProvider(
```
kiootic: I don't think this is a mock provider? It still perform the real provider logic.

pkg/auth/dependency/mfa/pq/store.go
```diff
@@ -0,0 +1,162 @@
...
+			"arc.created_at",
+			"arc.consumed",
+		).
+		From(s.sqlBuilder.FullTableName("authenticator"), "a").
+		Join(
+			s.sqlBuilder.FullTableName("authenticator_recovery_code"),
+			"arc",
+			"a.id = arc.id",
+		).
+		Where("a.user_id = ?", userID)
```
kiootic: Also, the order of this list of not deterministic. Maybe sort by code?

pkg/auth/dependency/authnsession/provider.go
```diff
@@ -0,0 +1,24 @@
...
+	"net/http"
+
+	"github.com/skygeario/skygear-server/pkg/auth/event"
+	"github.com/skygeario/skygear-server/pkg/core/auth"
+)
+
+// Provider manipulates authentication session
+type Provider interface {
+	// NewWithToken decodes an authentication session from a token.
+	NewWithToken(token string) (*auth.AuthnSession, error)
```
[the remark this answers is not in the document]
louischan-oursky: Let me change it to `NewFromToken`

pkg/auth/dependency/authnsession/provider.go
```diff
@@ -0,0 +1,24 @@
...
+	// NewFromScratch creates a new authentication session.
+	NewFromScratch(userID string, principalID string, reason event.SessionCreateReason) auth.AuthnSession
+	// GenerateResponse generates authentication response.
+	GenerateResponse(session auth.AuthnSession) (interface{}, error)
+	// WriteResponse alters the response, write Cookies and write HTTP Body. It should be used in a defer block.
+	// It should be used in most cases.
+	WriteResponse(w http.ResponseWriter, resp interface{}, err error)
```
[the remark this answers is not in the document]
louischan-oursky: Actually I have been thinking the other way round. Is it a good idea to merge `SessionProvider` and `SessionWriter`?

pkg/core/auth/authnsession.go
```diff
@@ -11,11 +11,12 @@ const (
...
-	UserID        string             `json:"user_id"`
-	PrincipalID   string             `json:"principal_id"`
-	RequiredSteps []AuthnSessionStep `json:"required_steps"`
-	FinishedSteps []AuthnSessionStep `json:"finished_steps"`
+	ClientID            string             `json:"client_id"`
+	UserID              string             `json:"user_id"`
+	PrincipalID         string             `json:"principal_id"`
+	RequiredSteps       []AuthnSessionStep `json:"required_steps"`
+	FinishedSteps       []AuthnSessionStep `json:"finished_steps"`
+	SessionCreateReason string             `json:"session_create_reason"`
```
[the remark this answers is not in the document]
louischan-oursky: We do have such enum in `event/session_create.go`. `auth/authnsession.go` depends on it would be strange. Move `SessionCreateReason` to `core/auth`?

pkg/core/auth/session/provider_impl.go
```diff
@@ -48,24 +48,18 @@ func NewProvider(req *http.Request, store Store, eventStore EventStore, authCont
...
-		UserID:      userID,
-		PrincipalID: principalID,
-
-		InitialAccess: accessEvent,
-		LastAccess:    accessEvent,
-
-		CreatedAt:  now,
-		AccessedAt: now,
-	}
+	sess := authnSess.Session()
```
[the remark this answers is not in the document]
louischan-oursky: Perhaps change it to functional builder pattern?

```golang
sess := authnSess.BuildSession(func (sess *Session) {
  sess.ID := id
  // ...
})
```

I think it is now hard to be misused.

pkg/auth/dependency/authnsession/mock_provider.go
louischan-oursky: Yes. It is not a mock provider. But the mock version of it is nearly identical to its real version because this provider depends on other providers which are all mocked. I skipped writing a mock.

pkg/core/auth/authnsession.go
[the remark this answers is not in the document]
kiootic: I agree, I'm thinking about doing `type SessionCreateReason auth.SessionCreateReason` in event package.

pkg/auth/dependency/authnsession/provider.go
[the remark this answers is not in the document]
kiootic: When I was implementing it, I wanted to separate application logic and presentation logic (i.e. HTTP I/O), so I put them in different interface (i.e. Provider & Writer). Also for this specific case, I think other APIs returning an AuthResponse with access token (e.g. refreshing token) can benefit from this helper function.

pkg/core/auth/session/provider_impl.go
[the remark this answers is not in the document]
kiootic: The builder would look rather complicated. My main points are:
- it is unlikely to be used correctly by other callers; and
- it is unlikely that other callers needed to use it.
Therefore I suggested just remove `Session()` and inline it here.

pkg/auth/dependency/authnsession/provider_impl.go
```diff
@@ -155,7 +155,7 @@ func (p *providerImpl) NewFromScratch(userID string, principalID string, reason
 	}, nil
 }
 
-func (p *providerImpl) GenerateResponse(authnSess auth.AuthnSession) (interface{}, error) {
+func (p *providerImpl) GenerateResponseAndUpdateLastLoginAt(authnSess auth.AuthnSession) (interface{}, error) {
 	step, ok := authnSess.NextStep()
 	if !ok {
 		var authInfo authinfo.AuthInfo
```
kiootic: I think this entire block should be refactored into `SessionProvider`: it is logically creating a session with needed extra step. Maybe fold it into `SessionProvider.Create`? Then, we can keep the same name, since creating a session should always update `last_login_at`.

pkg/auth/dependency/authnsession/provider_impl.go
```diff
@@ -236,7 +236,12 @@ func (p *providerImpl) WriteResponse(w http.ResponseWriter, resp interface{}, er
 	if err == nil {
 		switch v := resp.(type) {
 		case model.AuthResponse:
-			p.sessionWriter.WriteSession(w, &v.AccessToken, &v.MFABearerToken)
+			// Do not touch the cookie if it is not in the response.
+			if v.MFABearerToken == "" {
```
kiootic: It seems we can't distinguish the cases of token is absent and token should be deleted. Do we have a way to delete a bearer token in cookie?

pkg/auth/dependency/mfa/pq/store.go
```diff
@@ -322,6 +322,45 @@ func (s *storeImpl) CreateBearerToken(a *mfa.BearerTokenAuthenticator) error {
...
+			"abt.created_at",
+			"abt.expire_at",
+		).
+		From(s.sqlBuilder.FullTableName("authenticator"), "a").
+		Join(
+			s.sqlBuilder.FullTableName("authenticator_bearer_token"),
+			"abt",
+			"a.id = abt.id",
+		).
+		Where("a.user_id = ? AND abt.token = ?", userID, token)
```
kiootic: I think we need to compare the token in constant time.

pkg/auth/handler/mfa/authenticate_bearer_token.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (h *AuthenticateBearerTokenHandler) DecodeRequest(request *http.Request) (handler.RequestPayload, error) {
+	payload := AuthenticateBearerTokenRequest{}
+	err := handler.DecodeJSONBody(request, &payload)
+	if err != nil {
+		return nil, err
+	}
+
+	if payload.BearerToken == "" {
+		cookie, err := request.Cookie(coreHttp.CookieNameMFABearerToken)
```
kiootic: We should only use the bearer token in cookie if cookie is enabled.

pkg/auth/handler/mfa/create_oob.go
```diff
@@ -0,0 +1,156 @@
...
+	AuthenticatorID   string `json:"authenticator_id"`
+	AuthenticatorType string `json:"authenticator_type"`
+	Channel           string `json:"channel"`
+}
+
+// @JSONSchema
+const CreateOOBRequestSchema = `
+{
+	"$id": "#CreateOOBRequest",
+	"oneOf": [
```
kiootic: Just want to note that the presence of `phone` or `email` must depend on the `channel`. i.e. `additionalProperties` should be false.

pkg/auth/dependency/authnsession/provider.go
```diff
@@ -7,6 +7,17 @@ import (
...
 
+type ResolveUserIDOptions struct {
+	MFACase ResolveUserIDMFACase
+}
+
+type ResolveUserIDMFACase int
+
+const (
+	ResolveUserIDMFACaseAlwaysAccept ResolveUserIDMFACase = iota
+	ResolveUserIDMfaCaseOnlyWhenNoAuthenticators
```
kiootic: Let's use consistent casing (`Mfa` vs `MFA`)

pkg/auth/dependency/mfa/provider_impl.go
```diff
@@ -126,6 +126,32 @@ func (p *providerImpl) ActivateTOTP(userID string, id string, code string) ([]st
...
+		return nil, err
+	}
+
+	now := p.timeProvider.NowUTC()
+	for _, iface := range authenticators {
+		switch a := iface.(type) {
+		case TOTPAuthenticator:
+			ok, err := ValidateTOTP(a.Secret, code, now)
+			if err != nil {
+				return nil, err
```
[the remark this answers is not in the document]
louischan-oursky: I will change the function signature of ValidateTOTP to return just bool.

pkg/auth/dependency/authnsession/provider_impl.go
louischan-oursky: That would greatly increase the dependency of SessionProvider.

pkg/auth/dependency/mfa/pq/store.go
louischan-oursky: It requires us to fetch all tokens from database. The number can be unbound.

kiootic: I see, we do not limit amount of bearer tokens associated with a single authenticator. I think we should add a note here stating the limitation, since it may be a security risk.

pkg/auth/dependency/authnsession/provider_impl.go
kiootic: Okay let's just keep it as is for now.

pkg/auth/dependency/mfa/pq/store.go
louischan-oursky: Sure.

pkg/auth/dependency/mfa/pq/store.go
```diff
@@ -0,0 +1,811 @@
...
+			"at.secret",
+			"at.display_name",
+		).
+		From(s.sqlBuilder.FullTableName("authenticator"), "a").
+		Join(
+			s.sqlBuilder.FullTableName("authenticator_totp"),
+			"at",
+			"a.id = at.id",
+		).
+		Where("a.user_id = ? AND at.activated = TRUE", userID)
```
kiootic: I think we need stable order here too.

kiootic:
- The API documentation has invalid syntax, and some has inaccurate description ('List recovery codes if allowed.' for listing authenticators)
- Seems authentication session can be reused after finishing (i.e. multiple sessions can be created), do we want to allow it?
- Do we have mechanism to clear stale data in database? e.g. not yet activated authenticators (cannot be listed), expired bearer tokens.

pkg/auth/dependency/mfa/authenticator.go
```diff
@@ -0,0 +1,211 @@
...
+	incrFunc := func(a interface{}) {
+		switch aa := a.(type) {
+		case TOTPAuthenticator:
+			totpCount++
+		case OOBAuthenticator:
+			switch aa.Channel {
+			case coreAuth.AuthenticatorOOBChannelSMS:
+				oobSMSCount++
+			case coreAuth.AuthenticatorOOBChannelEmail:
+				oobSMSCount++
```
kiootic: Typo

louischan-oursky:
> Seems authentication session can be reused after finishing (i.e. multiple sessions can be created), do we want to allow it?

Somehow we need to remember the token was consumed. The consumption of the token is the creation of a new session. A trivial way to remember this is to generate the session ID when the token is generated. When the token is turned into a session, the ID is inherited. Later on when the same token is turned into a session again, we check if the session exists. The downside is that the session ID is no longer unpredictable and is fixed once the authentication session was created.

> Do we have mechanism to clear stale data in database? e.g. not yet activated authenticators (cannot be listed), expired bearer tokens.

We can housekeep non yet activated authenticators in `/mfa/*/activate`. It is based on the assumption that the user do not register authenticator concurrently.

We can housekeep expired bearer token in `mfa/bearer_token/authenticate`.

louischan-oursky:
I am now rethinking how can we minimize housekeeping inactive authenticators.

For TOTP, `/mfa/totp/new` always ensure there is one and only one inactive fresh authenticator. `/mfa/totp/activate` no longer takes authenticator ID and activate the only authenticator.

For OOB, `/mfa/oob/new` always ensure there is one and only one inactive authenticator. If the phone or email is the same, the existing inactive authenticator and the code are reused. If the code is expired, a new code is generated and delivered again. If the code is still valid, no delivery. `/mfa/oob/activate` no longer takes authenticator ID and activate the only authenticator.

Is this good enough?

kiootic:
- I think it's okay to allow reusing the authentication session: it's as if the authentication session token represents the entered info (e.g. password, MFA code). As long as it is short-lived enough it should be acceptable. We can note down this behavior in code comments/spec.
- I think only allowing one inactive authenticator is an acceptable approach to avoid too many stale records. I suppose this should not affect UX for adding MFA authenticator.

louischan-oursky:
Let me update the spec to include our conclusion.

pkg/auth/dependency/mfa/pq/store.go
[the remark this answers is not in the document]
louischan-oursky: I now hide the delete cascading delete logic in the store implementation so provider is unaware of it.

pkg/auth/dependency/authnsession/provider_impl.go
louischan-oursky: Addressed this issue in commit `Return InvalidMFABearerToken and clear the token when it is invalid`

Issue #1020, opened 2019-08-29
Refactor middlewares


kiootic:
For `core/server`, it has two more responsibilities besides managing the server:
- setup database context
- check authz policy 

These logics should be refactor into middlewares. Also, the setup of Redis context (#1017) should be also refactored into middlewares.

carmenlau:
Currently when auth gear fail to connect db or redis, the error is the clear to show the problems. 

For example, if it fails to connect to redis, `AuthnMiddleware` will call session writer to clear the session and api will return error `require authenticated user`.

PR #1033, opened 2019-09-10
Make session error handling more fine-grained

kiootic:
ref #1020

### DIFF
--- a/pkg/core/middleware/authn.go
+++ b/pkg/core/middleware/authn.go
@@ -3,6 +3,8 @@ package middleware
 import (
 	"net/http"
 
+	"github.com/skygeario/skygear-server/pkg/core/logging"
+
 	"github.com/skygeario/skygear-server/pkg/core/auth"
 	"github.com/skygeario/skygear-server/pkg/core/auth/authinfo"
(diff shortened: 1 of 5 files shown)


pkg/core/middleware/authn.go
```diff
@@ -33,9 +34,12 @@ func (m *AuthnMiddleware) Handle(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		var err error
 		defer func() {
-			if err != nil {
-				// clear session if error occurred
+			if err == session.ErrSessionNotFound {
+				// clear session if session is not found
 				m.SessionWriter.ClearSession(w)
+			} else if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
```
carmenlau: Some questions, 
1. I think we should provide the error message for debugging. If we don't want to expose the error message to client, maybe we should log the error? Otherwise, it is difficult to debug.
2. Can the SDK handle this kind of error? Without the response body.

kiootic: 1. Yes, I've updated to log the error.
2. I've opened an issue to handle it in JS SDK.

PR #1039, opened 2019-09-24
Multi tenant azure AD

louischan-oursky:
The code needed to change is actually very little but this PR fixes a couple of things.

1. Fix DenyDisabledUser
2. Store API Key in Oauth state so that it can be retrieved in the callback
3. Make email optional
4. Include `tenant` in Identity model
5. Return AuthResponse in link
6. Actually support two special tenant values

The feature PR: https://github.com/SkygearIO/features/pull/359


pkg/auth/dependency/sso/sso.go
```diff
@@ -38,7 +38,8 @@ type State struct {
 	LoginState
 	LinkState
 	OAuthAuthorizationCodeFlowState
-	Nonce string `json:"nonce,omitempty"`
+	Nonce  string `json:"nonce,omitempty"`
+	APIKey string `json:"api_key"`
```
kiootic: I think it's better to store just the client ID in OAuth state, since the state is already authenticated using JWT. API key can be considered as a sensitive information so we should minimize where it would be stored.

pkg/auth/dependency/sso/azureadv2.go
```diff
@@ -32,7 +28,46 @@ type azureadv2OpenIDConfiguration struct {
 
 func (f *Azureadv2Impl) getOpenIDConfiguration() (c azureadv2OpenIDConfiguration, err error) {
 	// TODO(sso): Cache OpenID configuration
-	endpoint := fmt.Sprintf(azureadv2ConfigurationURLFormat, f.ProviderConfig.Tenant)
+
+	tenant := f.ProviderConfig.Tenant
+	var endpoint string
+	// GUIDE(sso): Azure special tenant
```
kiootic: I don't think we should include guides in comment. It's better to put it in documentation/spec.

pkg/auth/dependency/sso/azureadv2.go
```diff
@@ -206,11 +206,10 @@ func (f *Azureadv2Impl) OpenIDConnectGetAuthInfo(r OAuthAuthorizationResponse) (
...
 	}
-	email, ok := mapClaims["email"].(string)
-	if !ok {
-		err = fmt.Errorf("cannot find email")
-		return
-	}
+	// For "Microsoft Account", email usually exists.
+	// For "AD guest user", email usually exists because to invite an user, the inviter must provide email.
+	// For "AD user", email never exists even one is provided in "Authentication Methods".
+	email, _ := mapClaims["email"].(string)
```
kiootic: If I understand correctly, most use cases for AD would fall into the "AD user" case, which email could not be obtained, so auto-linking would not work in this case.

pkg/auth/handler/sso/sso.go
```diff
@@ -160,7 +160,7 @@ func (h respHandler) linkActionResp(oauthAuthInfo sso.AuthInfo, linkState sso.Li
 		return
 	}
 
-	resp = map[string]string{}
+	resp = model.NewAuthResponseWithUser(user)
```
kiootic: It seems quite useful to include the just created identity in auth response, but it would be inconsistent with other auth response, where the current session identity is returned. What do you think about it?

louischan-oursky: I have the exact concern as yours so I didn't include it.

pkg/auth/dependency/sso/azureadv2.go
louischan-oursky: Yes. But we can do nothing about it.

pkg/auth/dependency/sso/azureadv2.go
louischan-oursky: Later on when we have guide, I will move this comment there.

pkg/auth/dependency/sso/sso.go
louischan-oursky: I had discussed this with Carmen and we thought API Key is public and Client ID is private (at least it only appears in tenant configuration) so I included API key instead.

kiootic: For me, I don't think API key is public, otherwise we don't need this concept: API key is supposed to be accessible only to authorized clients (e.g. if somebody somehow get the server endpoint, they can't call API without also finding a valid API key from some client). Therefore API key should be considered a private information.

Personally I consider API key a more private (and sensitive) information than client ID: it is kind of like an access token for API.

louischan-oursky: @carmenlau Any thoughts?

louischan-oursky: Right now Client ID does not have any special meanings. We do not accept Client ID in any endpoints. So even it is known by a third party, they should not be able to do anything with it. As long as we can uphold this assumption. We can change (should change) to Client ID.

kiootic: In my opinion ID should not be a kind of sensitive info, so exposing client ID should not have major security implication.

louischan-oursky:
@kiootic Switched to use ClientID

PR #1045, opened 2019-09-30
Fix server configuration

kiootic:


pkg/gateway/handler/timeout.go
```diff
@@ -0,0 +1,10 @@
+package handler
+
+import (
+	"net/http"
+	"time"
+)
+
+func withTimeout(h http.Handler) http.Handler {
+	return http.TimeoutHandler(h, 60*time.Second, "Request Timeout")
```
louischan-oursky: Using this will break any microservice websocket endpoint because it does not support Hijacker interface.

https://golang.org/pkg/net/http/#TimeoutHandler

kiootic: There seems no easy way to set timeout at reverse proxy level, so for now I set it at `http.Server`.

PR #1049, opened 2019-10-03
Fix template rendering

kiootic:


pkg/core/template/template.go
```diff
@@ -97,18 +100,51 @@ func ParseHTMLTemplate(templateString string, context map[string]interface{}) (o
...
+var errLimitReached = skyerr.NewError(skyerr.UnexpectedError, "rendered template is too large")
+
+type limitedWriter struct {
+	w io.Writer
+	n int64
+}
+
+func (l *limitedWriter) Write(p []byte) (n int, err error) {
+	if l.n-int64(len(p)) <= 0 {
+		panic(errLimitReached)
```
louischan-oursky: Can we document why we must use panic?

Issue #1040, opened 2019-09-24
Incorrect error handling in some handlers


kiootic:
e.g. In sign up:
the response is written in a deferred block. However, if a panic occurred, the deferred block can run with both nil result and err, which breaks response writing and hook logic.

PR #1051, opened 2019-10-04
Fix handler error handling

kiootic:
ref #1040

### DIFF
--- a/pkg/auth/handler/change_password.go
+++ b/pkg/auth/handler/change_password.go
@@ -135,34 +135,33 @@ func (h ChangePasswordHandler) DecodeRequest(request *http.Request, resp http.Re
 
 func (h ChangePasswordHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 	var err error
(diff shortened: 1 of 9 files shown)


kiootic:
For example, when database is down, `Transactional` would panic with a 'connection refused' error. The deferred block would then run and, while `result` and `err` is nil, panic again in `WriteResponse`.

louischan-oursky:
Do you think we should reduce the usage of `panic` in our code? `panic` can still be used in cases like exhaustive switch statement or unreachable code.

kiootic:
I agree. However, we still need to handle when there is unexpected panic.

louischan-oursky:
I think we have a recover middleware for that?

kiootic:
The problem is we're doing response writing in a deferred block, which run regardless of panic. The `WriteResponse` function would panic itself if the panic is not handling correctly.

louischan-oursky:
Per our offline discussion, we are going to refactor the code so that the defer block will be gone.

Issue #1055, opened 2019-10-08
Use golang template library


kiootic:
pongo2 is difficult to harden its security. We'll use `html/template` & `text/template` module in golang instead.

- blacklist all control structures, except `if`
- context aware auto escaping (html/text)

PR #1058, opened 2019-10-09
Use golang templates

### LINKED ISSUE #1056: Log database connection issues as ERROR

kiootic:
ref #1055 

also fix #1056 

note: my limited testing shows that in the worst case, a 1MB malicious template could consume ~17MB memory during rendering due to escaping function. do we want to disable escaping functions, and enforce auto-escaping instead?

### DIFF
--- a/pkg/core/template/template_test.go
+++ b/pkg/core/template/template_test.go
@@ -11,12 +11,12 @@ import (
 func TestTemplateRender(t *testing.T) {
 	Convey("template rendering", t, func() {
 		Convey("should not render large templates", func() {
-			longStr := strings.Repeat("\\", 1024*512)
-			template := fmt.Sprintf(`{{if $v := "%s" | js}}{{$v|js}}{{$v|js}}{{$v|js}}{{$v|js}}{{end}}`, longStr)
+			longStr := strings.Repeat("&", MaxTemplateSize - 25)
+			template := fmt.Sprintf(`{{"%s"|html|html|html}}`, longStr)
 
 			var err error
 
-			_, err = ParseHTMLTemplate("test", template, nil)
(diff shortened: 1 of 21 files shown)


pkg/core/template/validation.go
```diff
@@ -0,0 +1,149 @@
...
+
+		return err == nil
+	}
+
+	traverseTree(tree, validateFn)
+	return
+}
+
+var badIdentifiers = []string{
+	"printf",
```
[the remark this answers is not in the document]
kiootic: `print` family of functions are banned. `call` should be okay to be allowed; it require we somehow put a function in the template context, which is unlikely.

louischan-oursky:
> do we want to disable escaping functions, and enforce auto-escaping instead?

So do we have auto escaping now? If yes, can we add a test case to demonstrate that?

kiootic:
@louischan-oursky Added some test case demonstrating context-aware rendering of HTML/JS/CSS.

pkg/core/template/template_test.go
```diff
@@ -19,5 +19,87 @@ func TestTemplateRender(t *testing.T) {
...
+				</script>
+			</body>
+			
+			</html>
+			`
+
+			out, err := ParseHTMLTemplate("test", template, map[string]interface{}{
+				"URL":             "https://www.example.com",
+				"Title":           "Welcome to <b>My App</b>.",
+				"BackgroundColor": "#101020; /* for contrast */",
```
louischan-oursky: Why is BackgroundColor considered unsafe?

kiootic: The template placeholder is in CSS value context, so only plain CSS value is allowed: comment is disallowed.

Issue #1054, opened 2019-10-04
Asset Gear


louischan-oursky:
- [x] Azure Storage
- [x] AWS S3
- [x] Google Cloud Storage
- [x] `POST /_asset/presign_upload`
- [x] `POST /_asset/sign` with direct backend URL
- [x] `GET /_asset/get/<asset-name>`
- [x] `POST /_asset/sign` with Asset Gear URL
- [x] `GET /_asset/assets`
- [x] `DELETE /_asset/delete/<asset-name>`
- [x] Install `github.com/davidbyttow/govips`
  - [x] Locally
  - [x] In Docker image
- [x] Implement image processing via libvips
  - [x] Parse query
  - [x] Apply parsed query with libvips
- [x] Support image processing in `/_asset/get/<asset-name>`

PR #1070, opened 2019-10-18
Asset gear

louischan-oursky:
connect https://github.com/SkygearIO/skygear-server/issues/1054


pkg/asset/handler/presign_upload.go
```diff
@@ -0,0 +1,167 @@
...
+// @JSONSchema
+const PresignUploadRequestSchema = `
+{
+	"$id": "#PresignUploadRequest",
+	"type": "object",
+	"additionalProperties": false,
+	"properties": {
+		"prefix": {
+			"type": "string",
+			"pattern": "^[^\\x00\\\\/:*'<>|]*$"
```
kiootic: Do you have any reference on this regex?

pkg/core/imageprocessing/operation.go
```diff
@@ -0,0 +1,16 @@
+package imageprocessing
```
kiootic: I think this package should belongs to the asset gear, not the core library.

scripts/template/index.html
```diff
@@ -2,13 +2,12 @@
...
-        <script type="text/javascript">
-            window.location.href = "{URL}"
-        </script>
-        <title>Page Redirection</title>
+        <title>Skygear HTTP API Reference</title>
     </head>
     <body>
-        If you are not redirected automatically, follow this <a href='{URL}'>link to documentation</a>.
+        <ul>
+            <li><a href="{AUTH_URL}">Auth Gear</a></li>
```
kiootic: I think we can try using swagger-ui configuration to show both in same page.

ref: https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md

go.mod
```diff
@@ -48,16 +52,14 @@ require (
...
+	google.golang.org/api v0.9.0
 	gopkg.in/airbrake/gobrake.v2 v2.0.9 // indirect
 	gopkg.in/alexcesaro/quotedprintable.v3 v3.0.0-20150716171945-2caba252f4dc // indirect
-	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
 	gopkg.in/gemnasium/logrus-airbrake-hook.v2 v2.1.2 // indirect
 	gopkg.in/gomail.v2 v2.0.0-20160411212932-81ebce5c23df // indirect
 	gopkg.in/yaml.v2 v2.2.2
 )
+
+replace github.com/davidbyttow/govips => github.com/skygeario/govips v0.0.0-20191017114550-b7b5b9596467
```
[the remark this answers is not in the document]
louischan-oursky: I guess go.mod supports comment. Let me clarify why a fork is needed.

pkg/asset/handler/get.go
```diff
@@ -0,0 +1,138 @@
...
+			resp.Header.Del("Accept-Ranges")
+		}
+
+		// Check access
+		accessType := h.CloudStorageProvider.AccessType(resp.Header)
+		if accessType == cloudstorage.AccessTypePrivate && !originallySigned {
+			return ErrBadAccess
+		}
+
+		valid := imageprocessing.IsApplicableToHTTPResponse(resp)
```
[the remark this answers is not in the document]
louischan-oursky: Actually HEAD is only partially supported. The URL returned by `/_asset/get_signed_url` is a GET request. For some provider (S3), a signed request of GET and HEAD has distinct signature while other providers allow HEAD even if the signed request is GET. Therefore, if the provider is S3, the signed URL can only be used as GET.

pkg/asset/handler/presign_upload.go
louischan-oursky: I forgot. Let me revise it to match https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words

pkg/core/imageprocessing/operation.go
louischan-oursky: Just in case anyone else wants to have image processing.

pkg/core/imageprocessing/resize.go
```diff
@@ -0,0 +1,281 @@
...
+		targetWidth = originalWidth
+		targetHeight = originalHeight
+	}
+
+	// w is not specified
+	if targetWidth == 0 {
+		if o.ScalingMode.ShouldInferAnotherSideWithAspectRatio() {
+			targetWidth = roundFloat(float64(targetHeight) * aspectRatio)
+		} else {
+			targetWidth = targetHeight
```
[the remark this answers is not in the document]
louischan-oursky: It is by definition.

For some scaling mode, it makes more sense to have `w` default to `h`.

For example, for resizing mode `fill`, if only width is specified, it makes more sense to have height equal to width. If height is equal to the width divided by the aspect ratio, you cannot see the `fill` effect because the aspect ratio is the same.

scripts/template/index.html
louischan-oursky: It seems that a separate doc for each gear is better for the developer. If the developer wants to use Auth Gear, they do not need to see the doc of Asset Gear. It is distraction for them.

pkg/core/imageprocessing/resize.go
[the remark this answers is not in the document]
kiootic: Okay. I think we need to note this down in the spec, that the default value of `w` and `h` depends on the scaling mode and aspect ratio.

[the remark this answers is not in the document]
louischan-oursky: https://github.com/SkygearIO/features/pull/360/commits/bb95bfffa2f0c13c91906acc869bb0f52619e517

pkg/asset/handler/presign_upload.go
louischan-oursky: Let me revise it to `a-z` `A-Z` `0-9` `_-.`

cmd/asset/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+FROM alpine:3.10
+
+RUN apk add \
+    --repository https://dl-3.alpinelinux.org/alpine/edge/testing \
+    --repository https://dl-3.alpinelinux.org/alpine/edge/main \
+    --repository https://dl-3.alpinelinux.org/alpine/edge/community \
+    --update \
+    --no-cache \
+    libc6-compat ca-certificates vips vips-dev && \
```
kiootic: It seems `libc6-compat` and `vips-dev` are not needed.

pkg/core/cloudstorage/presign_upload_request.go
```diff
@@ -0,0 +1,81 @@
...
+		r.Headers["content-type"] = "application/octet-stream"
+	}
+}
+
+func (r *PresignUploadRequest) DeriveAssetName() (assetName string, err error) {
+	// Derive file extension
+	contentType, ok := r.Headers["content-type"].(string)
+	var ext string
+	if ok {
+		exts, err := mime.ExtensionsByType(contentType)
```
kiootic: Seems `mime` only has a small set of built-in types, we need to install `mailcap` on alpine to get better extension list.

cmd/asset/Dockerfile
louischan-oursky: libc6-compat is not added by me.
I guess vips-dev is needed because this package provides header files.

kiootic: `vips-dev` is needed at build time, but it shouldn't be needed at runtime?

louischan-oursky: From [the description of libc6-compat](https://pkgs.alpinelinux.org/package/edge/main/x86/libc6-compat), it seems that it is needed if the program uses glibc. And vips does use glibc.

louischan-oursky: Anyway let me experiment if we can remove both

kiootic: `ldd` seems to show it's using musl:
```
/ # ldd skygear-asset 
        /lib/ld-musl-x86_64.so.1 (0x7f68c629b000)
        libvips.so.42 => /usr/lib/libvips.so.42 (0x7f68c5e1a000)
        libgobject-2.0.so.0 => /usr/lib/libgobject-2.0.so.0 (0x7f68c5dd0000)
        libglib-2.0.so.0 => /usr/lib/libglib-2.0.so.0 (0x7f68c5cd1000)
        libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7f68c629b000)
...
```

pkg/core/cloudstorage/provider_impl.go
```diff
@@ -81,14 +88,17 @@ func (p *providerImpl) checkDuplicate(assetID string) error {
...
-			return nil, err
+		u := &url.URL{
+			Scheme: scheme,
+			Host:   host,
+			Path:   fmt.Sprintf("/_asset/get/%s", assetItem.AssetName),
 		}
-		r.Assets[i].URL = u.String()
+		httpRequest, _ := http.NewRequest("GET", u.String(), nil)
+		httpsigning.Sign(p.secret, httpRequest, now, int(PresignGetExpires.Seconds()))
+		r.Assets[i].URL = httpRequest.URL.String()
```
kiootic: The signature of this function seems returns a new sign request, but it actually modify the input request in-place. It would be better to either really return a new request, or remove the request from return value.

pkg/asset/handler/get.go
```diff
@@ -108,7 +122,7 @@ func (h *GetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 
 		valid := imageprocessing.IsApplicableToHTTPResponse(resp)
-		if !valid || !hasPipeline {
+		if isHead || !valid || !hasPipeline {
```
kiootic: If it's `HEAD`, we should try not sending the body to client.

pkg/core/cloudstorage/provider_impl.go
```diff
@@ -103,6 +103,31 @@ func (p *providerImpl) Sign(scheme string, host string, r *SignRequest) (*SignRe
 	return r, nil
 }
 
+func (p *providerImpl) Verify(r *http.Request) error {
+	now := p.timeProvider.NowUTC()
+	copiedReq := &http.Request{
```
kiootic: I think we can just do `copiedReq = *r` and use `&copiedReq` instead.

pkg/asset/handler/get.go
louischan-oursky: I tried to do this but a weird error got printed to console saying that somebody wrote to body after header is flushed.

kiootic: I think replacing `resp.Body` with `ioutil.NopCloser(bytes.NewReader([]byte{}))` should work?

cmd/asset/main.go
```diff
@@ -111,21 +109,13 @@ func main() {
...
-				}
-			}
 			logger.WithError(err).Fatal("Cannot parse standalone config")
 		}
 
 		serverOption := server.DefaultOption()
 		serverOption.GearPathPrefix = "/_asset"
 		srv = server.NewServerWithOption(configuration.ServerHost, dependencyMap, serverOption)
-		srv.Use(middleware.TenantConfigurationMiddleware{
+		srv.Use(middleware.WriteTenantConfigMiddleware{
```
kiootic: We'll need `ReadTenantConfigMiddleware` otherwise, in the else block.

pkg/asset/handler/upload_form.go
```diff
@@ -61,23 +62,22 @@ func (h *UploadFormHandler) Handle(w http.ResponseWriter, r *http.Request) (err
...
 	if err != nil {
-		err = skyerr.NewError(skyerr.BadRequest, err.Error())
 		return
 	}
 
 	contentType := r.Header.Get("Content-Type")
 	mediaType, params, err := mime.ParseMediaType(contentType)
 	if err != nil {
-		err = skyerr.NewError(skyerr.BadRequest, "invalid content-type")
+		err = skyerr.Invalid.WithReason("InvalidContentType").New("invalid content-type")
```
kiootic: These errors should be `BadRequest` instead of `Invalid`, since the server does not understand the request. I should clarify that `BadRequest` and `Invalid` is like syntactical and semantic error.

Also, there is not need to provide this level of details to client. I think it is better to use a generic reason `InvalidUploadBody` for these errors, since these errors are programming errors that would not be inspected at runtime.

louischan-oursky: > I should clarify that BadRequest and Invalid is like syntactical and semantic error.

Thanks. Better to put this guideline in the doc.

PR #1075, opened 2019-10-23
Update error handling

### LINKED ISSUE #1034: Refactor errors
We don't have standard way to define errors in code and cause following problems:

- Some of the api return incorrect status code
- Exposing unexpected detail of error in api response, e.g. For production environment, should not show sql error in response. We should log the details and response should only show unexpected error.

Discussions:

- Which layer should be responsible to the status code (error name), store, provider or handler?
- Should we wrap the errors so that we can trace how the error occur? How to wrap? And how to generate the error message for response, should not show the whole trace?

Confirmed todos:

- Remove error code, should use error name instead

kiootic:
NOTE:
```
- error types:
    - panic:
        - unreachable code (e.g. enum exhaustive check)
        - server programming error (e.g. unexpected parameter type/query result)
        - should try to be string of format "<module>: <message>", or errors produced elsewhere
    - plain errors: errors from library/`errors.Newf()`, etc.
        - unexpected errors/user config errors
        - to be caught by top-level code (i.e. recover/APIHandler) and logged
        - e.g. DB/network errors, library errors
    - skyerrs: `skyerr.Kind.New()`, etc.
        - commonly occurs, returned in API response
        - message is used for debugging, should be mostly literals
    - generic skyerrs: `skyerr.NewInteralError()`/`skyerr.NewBadRequest()`
        - client programming errors (not config error, i.e. should not happen if code is correct)
        - unrecoverable errors (i.e. cannot handle reasonably by app/server)
        - e.g. panic, invalid JSON encoding, body size too large
    - error sentinels: `var ErrNotFound = errors.New()`, etc.
        - to be interpreted by upper layer code
        - upper layer must convert to skyerrs/another error sentinels
        - e.g. ErrUserNotFound
- handler/task -> provider/store
    - store: IO stuff (i.e. network/DB)
        - returns plain errors/error sentinels
        - no skyerrs
    - provider: application logic, depends on store/provider
        - returns skyerrs/plain errors/error sentinels
        - errors from store must be wrapped with Handled/HandledWithMessage
        - errors from provider should be preserved, optionally wrapped
        - use `skyerr.IsKind` to check for skyerr sentinels
        - authinfo & userprofile are treated as provider
    - handlers/task: HTTP transport, depends on store/provider
        - must return skyerrs, or pass through provider errors
- use `errors.Is` or `skyerr.IsKind` instead of equality check
- return NotFound error sentinels for single entity, empty slice for multiple entities
- skyerr:
    - name 1..1 reason 1..1 info shape
    - kind (i.e. reason): should be defined in own module, instead globally in core
    - info:
        - e.g. `{ "cause": "ExpiredCode" }`
        - e.g. `{ "causes": [{"kind": "PasswordTooShort"}, {"kind": "PasswordContainsKeyword"}] }`
```

ref #1034


pkg/auth/task/send_verify_code.go
```diff
@@ -69,20 +68,16 @@ func (v *VerifyCodeSendTask) Run(param interface{}) (err error) {
 	loginID := taskParam.LoginID
 	userID := taskParam.UserID
 
-	v.Logger.WithFields(logrus.Fields{
-		"user_id": userID,
-	}).Debug("Start sending user verify message")
+	v.Logger.WithFields(logrus.Fields{"user_id": taskParam.UserID}).Debug("Sending verification email")
```
louischan-oursky: The "email" could be an SMS message.

cmd/auth/main.go
```diff
@@ -120,14 +118,6 @@ func main() {
 		}
 		tenantConfig, err := config.NewTenantConfigurationFromYAML(reader)
 		if err != nil {
-			if skyError, ok := err.(skyerr.Error); ok {
```
louischan-oursky: This is actually needed (for ourselves). The intention here is to pretty-print the validation error to stderr so that we (the skygear team member) knows what went wrong. Can we leave a TODO here to restore this functionality when we have implemented validation error?

pkg/core/auth/authinfo/mock_store.go
```diff
@@ -33,7 +33,7 @@ func NewMockStoreWithAuthInfoMap(authInfoMap map[string]AuthInfo) *MockStore {
 // CreateAuth creates AuthInfo in AuthInfoMap.
 func (s *MockStore) CreateAuth(authinfo *AuthInfo) error {
 	if _, existed := s.AuthInfoMap[authinfo.ID]; existed {
-		return skydb.ErrUserDuplicated
+		return errors.New("dupliated auth info")
```
louischan-oursky: typo

pkg/core/handler/api.go
```diff
@@ -79,17 +86,27 @@ func APIHandlerToHandler(apiHandler APIHandler, txContext db.TxContext) http.Han
...
+	if err != nil {
+		httpStatus = err.Code
 	}
 
 	rw.Header().Set("Content-Type", "application/json")
 	rw.WriteHeader(httpStatus)
 	encoder.Encode(response)
+
+	if err != nil && err.Code >= 500 && err.Code < 600 {
+		// delegate logging to panic recovery
```
louischan-oursky: Can we document the struct `HandledError` instead? For example, mention it is meant to be used as sentinel value.

pkg/core/xskyerr/kinds.go
```diff
@@ -0,0 +1,49 @@
...
+
+type Name string
+
+const (
+	BadRequest         Name = "BadRequest"
+	Invalid            Name = "Invalid"
+	Unauthorized       Name = "Unauthorized"
+	Forbidden          Name = "Forbidden"
+	NotFound           Name = "NotFound"
+	AlreadyExists      Name = "AlreadyExists"
```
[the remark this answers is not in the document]
kiootic: `Conflict` is rather generic, referring to all kinds of conflicting conditions. We only produce conflicts due to entity already exists, so I used a simpler and clearer name here.

Note that I referenced [k8s's error reasons](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#StatusReason) for the error names.

pkg/core/auth/session/provider_impl.go
```diff
@@ -101,6 +102,9 @@ func (p *providerImpl) GetByToken(token string, kind auth.SessionTokenKind) (*au
 
 	s, err := p.store.Get(id)
 	if err != nil {
+		if err != ErrSessionNotFound {
+			err = errors.HandledWithMessage(err, "failed to get session")
+		}
```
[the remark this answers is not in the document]
kiootic: To be precise, it is a sentinel error value, which providers are allowed to return. The upper layer can choose to handle it if it understand it (e.g. session not found error in refresh session handler), or propagate it if unexpected (e.g. user profile not found error in me handler).

pkg/core/auth/session/redis/store.go
```diff
@@ -156,14 +175,18 @@ func (s *store) DeleteAll(userID string, sessionID string) error {
...
+	if err == nil {
+		args := append([]interface{}{listKey}, sessionKeysDel...)
+		_, err = conn.Do("HDEL", args...)
+		if err != nil {
+			// ignore non-critical errors
+			s.logger.
+				WithError(err).
+				WithField("key", listKey).
+				Error("failed to update session list")
+		}
```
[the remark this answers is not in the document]
kiootic: It'll get non-effective assignment lint error, since err is not used as return value.

pkg/core/auth/authz/policy/compound_test.go
```diff
@@ -1,21 +1,35 @@
...
 import (
+	"errors"
 	"net/http"
 	"testing"
 
+	"github.com/skygeario/skygear-server/pkg/core/auth"
 	. "github.com/smartystreets/goconvey/convey"
 )
 
+type everybody struct {
```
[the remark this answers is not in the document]
kiootic: It is not used anywhere else, so I make it private to be used in this test.

pkg/auth/dependency/principal/password/provider.go
```diff
@@ -373,17 +372,16 @@ func (p *providerImpl) UpdatePassword(principal *Principal, password string) (er
...
-
+		err = errors.HandledWithMessage(err, "failed to update password")
 		return
 	}
 
-	if p.passwordHistoryEnabled && isPasswordChanged {
-		err = p.passwordHistoryStore.CreatePasswordHistory(
-			principal.UserID, principal.HashedPassword, timeNow(),
-		)
+	if isPasswordChanged {
```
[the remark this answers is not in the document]
kiootic: The condition is refactored into `savePasswordHistory`, since it is used by multiple places.

cmd/auth/main.go
kiootic: I intended that the validation errors would be stored in error details, and the logger would print it out automatically. I'll add a TODO to `TenantConfiguration.Validate` to return correct validation error details.

pkg/auth/handler/change_password_test.go
```diff
@@ -149,13 +149,14 @@ func TestChangePasswordHandler(t *testing.T) {
...
+						"reason": "PasswordPolicyViolated",
+						"message": "password policy violated",
+						"code": 400,
 						"info": {
-							"reason": "PasswordTooShort",
-							"min_length": 6,
-							"pw_length": 4
+							"causes": [
+								{ "kind": "PasswordTooShort", "min_length": 6, "pw_length": 4 }
+							]
```
carmenlau: Currently module defines its own shape of info, I think it is good that we have more flexible for future features. But seems now we already have some conventions?

For errors that may has multiple causes:
`{ "causes": [{"kind": "PasswordTooShort"}] }`
For error that only has single cause: 
`{ "cause": "ExpiredCode" }`

When new developers develop gears, they may not know this convention when developing new feature. To avoid having inconsistent in future maybe we can 1. Have helper functions to create cause and causes info. or  2. We have documentation about the error conventions. I prefer 1, thought?

Also for the shape of single cause, would it be better to use object instead of string? So we can add more data about the error if needed?

e.g.
```
{ "cause": { "kind": "ExpiredCode" } }
```

kiootic: Yes, the `causes` & `cause` is explained in the issue's note. I intended to organize the notes into the repo doc after reviewing.

I don't think a helper function would be much useful, since not all error would use this 'cause' convention:
- Some errors are not precisely an error, such as authentication session. The token and step is put in info directly.
- Some error would only have single kind of cause, such as user disabled error. In this case, the extra info is put directly in info.

For single cause error, I would prefer to flatten it directly at the info (i.e. `cause` field is same as `kind`). For example:
```json
{ "cause": "ExpiredCode", "user_id": "..." }
```

louischan-oursky: Just my gut feeling, `causes` and `cause` would make people think their difference is just cardinality. If the type of `cause` is `T`, then `causes` should be of type `T[]`. Currently it is not the case.

kiootic: That's a good point. I'll add helper functions `NewWithCause`/`NewWithCauses`, which helps to populate the info in proper format. However, I'll keep the `NewWithDetails` function for the non-conforming errors mentioned above.

Issue #1077, opened 2019-10-25
Support templates


louischan-oursky:
- [x] Remove inline template configuration from UserConfiguration and its descendant types
- [x] Add TemplateItemConfiguration
```golang
type TemplateItemConfiguration struct {
  Type TemplateItemType
  LanguageTag string
  Key string
  // URI is either
  // file:///<local-path>
  // asset-gear:///<asset-name>
  URI string
  ContentMD5 string
}
```
- [x] Add `TenantConfiguration.Templates []TemplateItemConfiguration`
- [x] Update the template resolution mechanism to use URI
  - [x] `file:`
  - [x] `asset-gear:`
- [x] Update the template resolution mechanism to respect LanguageTag
- [x] Make redirect have full access to context

PR #1087, opened 2019-10-31
Template

louischan-oursky:
connect #1077


pkg/core/template/engine.go
```diff
@@ -1,24 +1,32 @@
...
+	PreferredLanguageTags []string
 }
 
 type RenderOptions struct {
 	Required bool
 	Key      string
 }
 
-func NewEngine(enabledFileLoader bool) *Engine {
+func NewEngine(enabledFileLoader bool, templateItems []config.TemplateItem, tags []string) *Engine {
```
carmenlau: When we develop user preferred languages in the future, most likely we will need to get the language tags from the auth context.
Maybe we should add some comment to describe why we want to put the preferred languages tags here, we may need to update the middleware in future and see how can we get the preferred language in the DI function.

pkg/core/template/asset_gear_loader.go
```diff
@@ -0,0 +1,65 @@
...
+type AssetGearLoader struct {
+	AssetGearEndpoint  string
+	AssetGearMasterKey string
+}
+
+func (l *AssetGearLoader) Load(u *url.URL) (templateContent string, err error) {
+	// The url is asset-gear:///assetname
+	path := u.Path
+	assetName := strings.TrimPrefix(path, "/")
+	reqBody := map[string]interface{}{
```
kiootic: Can we define a local struct instead for request & response? It is rather hard to read.

pkg/auth/template/types.go
```diff
@@ -0,0 +1,24 @@
+package template
+
+import (
+	"github.com/skygeario/skygear-server/pkg/core/config"
+)
+
+const (
+	TemplateItemTypeForgotPasswordEmailTXT           config.TemplateItemType = "forgot_password_email.txt"
```
kiootic: I think these template types should be defined in corresponding feature package instead. This package should be a generic template library and should not have knowledge of individual features.

For example, we can have a `type Template struct { Key string; IsHTML bool; Default string }` and a global function `RegisterTemplate(t Template)`.

pkg/core/template/uri_loader.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (l *URILoader) Load(uri string) (templateContent string, err error) {
+	u, err := url.Parse(uri)
+	if err != nil {
+		err = errors.HandledWithMessage(err, "failed to parse template URI")
+		return
+	}
+
+	switch u.Scheme {
```
kiootic: It seems that:
- `EnableFileLoader` is configurable through environment variable
- `EnableDataLoader` is always disabled currently
- `AssetGearLoader` is provided if appropriate environment variable is set (endpoint & master key)

Is that correct?

pkg/core/template/template.go
```diff
@@ -2,41 +2,92 @@ package template
...
+	return
+}
+
+func SetContextToURLQuery(u *url.URL, context map[string]interface{}) error {
+	encoded, err := EncodeContextToURLQueryParamValue(context)
+	if err != nil {
+		return err
+	}
+	query := u.Query()
+	query.Set("x-skygear-redirect-data", encoded)
```
kiootic: Seems this is used to pass the template rendering context values to the redirection target. We may need to document it in specification.

pkg/core/template/uri_loader.go
louischan-oursky: Yes. DataLoader was introduced to implement test sender. But test sender was removed later. FileLoader is not intended for use on cloud deployment. It is helpful during testing. AssetGearLoader is for cloud deployment but is optional for local development.

pkg/core/template/template.go
louischan-oursky: Yes. But the entire user verification is currently unspecified. [The feature issue was closed](https://github.com/SkygearIO/features/issues/336)

pkg/core/template/uri_loader.go
kiootic: I suppose we'll have no API for sending test message, and if needed can use `DataLoader` to implement it in other project.

pkg/core/template/template.go
kiootic: I think this should be specified in the SkygearIO/features#251 ? It is a common functionality to features that would make use of template redirection.

pkg/core/template/uri_loader.go
louischan-oursky: Correct.

pkg/core/template/template.go
louischan-oursky: Left notes on that issue.

pkg/auth/template/types.go
louischan-oursky: See if https://github.com/SkygearIO/skygear-server/pull/1087/commits/725fb7495a7ec3dfa98feb1addaa897741a1705e looks good to you

kiootic: I'd prefer unifying `RenderTextTemplate` and `RenderHTMLTemplate` into `RenderTemplate(t T)`, and it will render it as text/HTML according to a flag `IsHTML` in `T`. I've fixed a bug that a HTML template is rendered as text before, so I think we may need this safe guarding here.

louischan-oursky: For some template types, we do not have the corresponding T so I didn't make RenderXXXTemplate to accept T. Also, T has Default, which I think is unnecessary information for RenderXXXTemplate.

kiootic: Okay. So for now we'll accept that we need to be careful when reviewing code so that correct function would be used to render the template.

louischan-oursky:
@kiootic OK to merge?

PR #1089, opened 2019-10-31
Use JSON schema for validation

kiootic:
ref #1034


pkg/core/handler/bind.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/skygeario/skygear-server/pkg/core/validation"
+)
+
+type bodyDefaulter interface {
+	SetDefaultValue()
+}
+
+type bodyValidator interface {
+	Validate() []validation.ErrorCause
+}
```
louischan-oursky: Why don't we make these two interfaces public?

pkg/asset/handler/presign_upload.go
```diff
@@ -140,18 +139,13 @@ func (h *PresignUploadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
 
-func (h *PresignUploadHandler) ParsePresignUploadRequest(r io.Reader, p interface{}) error {
-	return h.Validator.ParseReader("#PresignUploadRequest", r, p)
-}
-
 func (h *PresignUploadHandler) Handle(w http.ResponseWriter, r *http.Request) (result interface{}, err error) {
 	// Parse request
 	var payload cloudstorage.PresignUploadRequest
-	err = handler.ParseJSONBody(r, w, h.ParsePresignUploadRequest, &payload)
+	err = handler.BindJSONBody(r, w, h.Validator, "#PresignUploadRequest", &payload)
```
louischan-oursky: I think we can remove ParseJSONBody since it is replaced by the better BindJSONBody?

pkg/auth/handler/forgotpwd/reset_form.go
```diff
@@ -114,11 +115,11 @@ func (h ForgotPasswordResetFormHandler) prepareResultTemplateContext(r *http.Req
 	var payload ForgotPasswordResetFormPayload
 	payload, err = decodeForgotPasswordResetFormRequest(r)
 	if err != nil {
-		err = skyerr.NewBadRequest("invalid form data")
+		err = skyerr.NewBadRequest("invalid request form")
 		return
 	}
 
-	if err = payload.Validate(); err != nil {
+	if err = h.Validator.WithMessage("invalid request form").ValidateGoValue("#ForgotPasswordResetForm", payload); err != nil {
```
louischan-oursky: Is it possible to remove ValidateGoValue? I know we are validating a url.Values here. Can we first encode it to JSON to preserve the original form and then use ValidateReader? It allows us to specify `additionalProperties: false` to catch typo. We should specify `additionalProperties: false` whenever possible.

pkg/auth/handler/mfa/create_oob.go
```diff
@@ -80,17 +66,17 @@ const CreateOOBRequestSchema = `
...
 			},
 			"required": ["channel", "phone"]
 		},
 		{
 			"additionalProperties": false,
 			"properties": {
 				"channel": { "enum": ["email"] },
-				"email": { "type": "string" },
-				"authn_session_token": { "type": "string" }
+				"email": { "type": "string", "format": "email" },
```
louischan-oursky: I guess the implementation of `format: email` allow name in the address. Can you verify if it is true? If yes, can we introduce our own format that only allow email address without name?  We have `EnsureAddressOnly` implemented already.

pkg/auth/handler/session/get.go
```diff
@@ -47,15 +48,11 @@ const GetRequestSchema = `
 	"$id": "#SessionGetRequest",
 	"type": "object",
 	"properties": {
-		"session_id": { "type": "string" }
+		"session_id": { "type": "string", "minLength": 1 }
```
louischan-oursky: In most places, we should also include `"required": ["session_id"]` to ensure, for example, `session_id` is specified and is not empty.

pkg/auth/handler/session/revoke_all.go
```diff
@@ -68,68 +68,77 @@ func (h RevokeAllHandler) ProvideAuthzPolicy() authz.Policy {
...
-}
-
-func (h RevokeAllHandler) DecodeRequest(request *http.Request, resp http.ResponseWriter) (handler.RequestPayload, error) {
-	payload := handler.EmptyRequestPayload{}
-	err := handler.DecodeJSONBody(request, resp, &payload)
-	return payload, err
+func (h RevokeAllHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var response handler.APIResponse
+	var payload struct{}
+	if err := handler.DecodeJSONBody(r, w, &payload); err != nil {
```
louischan-oursky: Maybe we can use JSON Schema to ensure the input is an empty object.

```
{
  "type": "object",
  "additionalProperties": false
}
```

pkg/core/handler/bind.go
kiootic: No particular reason. I'll make it public then.

pkg/asset/handler/presign_upload.go
kiootic: `ParseJSONBody` is used by `DecodeJSONBody`, which is used by handlers accepting empty JSON body. If we'd like to remove them, then these handlers need to validate against the `emptyResponseSchema` schema.

pkg/auth/handler/forgotpwd/reset_form.go
kiootic: I don't think using `additionalProperties: false` would be forward compatible: a newer client that add extra fields in request would not be able to call old version of API. Do you think we should have this level of compatibility?

louischan-oursky: This issue is related to explicit API versioning. Let's not to include additionalProperties for now.

pkg/auth/dependency/hook/deliverer_impl.go
```diff
@@ -156,7 +155,7 @@ func noFollowRedirectPolicy(*gohttp.Request, []*gohttp.Request) error {
 	return gohttp.ErrUseLastResponse
 }
 
-func performRequest(client gohttp.Client, request *gohttp.Request, withResponse bool) (hookResp *event.HookResponse, err error) {
+func performRequest(client gohttp.Client, request *gohttp.Request, withResponse bool) (hookResp event.HookResponse, err error) {
```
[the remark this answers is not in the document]
kiootic: To simplify the parse function body at `hook_response.go`.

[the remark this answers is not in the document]
kiootic: As discussed, we will follow Go's convention of using pointer to struct on optional return values.

PR #1093, opened 2019-11-01
Override the expire of signed URL

### LINKED ISSUE #1092: Support overriding the expire of signed URL
- https://github.com/SkygearIO/features/pull/360/commits/aa54bf27b0db607e129a3fb2fd94d08aa6a2bed7

louischan-oursky:
fixes #1092

### DIFF
--- a/pkg/core/cloudstorage/provider_impl.go
+++ b/pkg/core/cloudstorage/provider_impl.go
@@ -88,7 +88,10 @@ func (p *providerImpl) checkDuplicate(assetID string) error {
 	return ErrDuplicateAsset
 }
 
-func (p *providerImpl) Sign(scheme string, host string, r *SignRequest) error {
+func (p *providerImpl) Sign(scheme string, host string, r *SignRequest) (*SignResponse, error) {
+	output := &SignResponse{
+		Assets: make([]SignedAssetItem, len(r.Assets)),
+	}
 	now := p.timeProvider.NowUTC()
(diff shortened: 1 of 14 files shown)


pkg/core/cloudstorage/provider_impl.go
```diff
@@ -97,7 +97,11 @@ func (p *providerImpl) Sign(scheme string, host string, r *SignRequest) error {
 			Path:   fmt.Sprintf("/_asset/get/%s", assetItem.AssetName),
 		}
 		httpRequest, _ := http.NewRequest("GET", u.String(), nil)
-		httpsigning.Sign(p.secret, httpRequest, now, int(PresignGetExpires.Seconds()))
+		expire := int(PresignGetExpires.Seconds())
+		if assetItem.Expire != 0 {
+			expire = assetItem.Expire
+		}
+		httpsigning.Sign(p.secret, httpRequest, now, expire)
```
carmenlau: Should we assign the default `PresignGetExpires.Seconds()` back to ` r.Assets[i].Expire`. if it is empty?

louischan-oursky: I intentionally did not do that because I wanted to keep it as zero value. If the input does not specify `expire`, then `expire` does not appear in the response.

louischan-oursky: Removed `expire` from response

PR #1097, opened 2019-11-04
Integrate Sentry

### LINKED ISSUE #900: Integrate sentry
Add sentry integration to gateway and auth gear

kiootic:
fix #900 

NOTE: use `SENTRY_DSN` environment variable to configure Sentry.


pkg/core/sentry/client.go
```diff
@@ -0,0 +1,41 @@
...
+var DefaultClient *Client
+
+func init() {
+	dsn := os.Getenv("SENTRY_DSN")
+	if dsn == "" {
+		DefaultClient = &Client{
+			Hub:   sentry.CurrentHub(),
+			Scope: sentry.CurrentHub().Scope(),
+		}
+		return
```
louischan-oursky: If we did not provide SENTRY_DSN, how would the integration behave?

I saw we add the sentry hook unconditionally. And we always add the sentry middleware.

kiootic: If sentry is not configured, everything would still work, except the call to `CaptureEvent` would be a no-op.

PR #1099, opened 2019-11-04
Validate login ID format

### LINKED ISSUE #1095: enforce login IDs' type constraints
(Describe the issue here)

- Skygear Server Date/Version:
- Application Platform:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Steps to reproduce

1.
2.
3.

kiootic:
fix #1095

### DIFF
--- a/pkg/auth/dependency/principal/password/login_id.go
+++ b/pkg/auth/dependency/principal/password/login_id.go
@@ -1,7 +1,11 @@
 package password
 
 import (
-	"errors"
+	"fmt"
+
+	"github.com/skygeario/skygear-server/pkg/core/skyerr"
+
+	"github.com/skygeario/skygear-server/pkg/core/validation"
 
 	"github.com/skygeario/skygear-server/pkg/core/auth/metadata"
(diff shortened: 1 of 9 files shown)


pkg/auth/dependency/principal/password/login_id.go
```diff
@@ -38,6 +40,10 @@ func (c defaultLoginIDChecker) validate(loginIDs []LoginID) error {
 			return errors.New("login ID is empty")
 		}
 		amounts[loginID.Key]++
+
+		if err := c.validateLoginIDFormat(loginID); err != nil {
```
louischan-oursky: In this validate function, some errors are programming errors, such as login ID is empty while some are user errors, such as invalid format. I think we want to tell the user and the developer that

1. Login ID exceeds maximum
2. Login ID is in invalid format

Other errors should be programming errors.

kiootic: I think except when login ID key is not allowed in configuration, other errors should be reported as validation errors to be consistent with schema validation of other APIs.

PR #1124, opened 2019-11-21
Add templates information API

### LINKED ISSUE #1123: APIs to return template information
Gears should have APIs to return supported template types and default template. The APIs should have no access control, and return static content (i.e. independent of tenant config).

kiootic:
fix #1123


pkg/auth/handler/gear/templates_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+			var handler = TemplatesHandler{
+				TemplateEngine: engine,
+			}
+
+			r, _ := http.NewRequest("GET", "", nil)
+			rw := httptest.NewRecorder()
+			handler.ServeHTTP(rw, r)
+
+			So(rw.Body.Bytes(), ShouldEqualJSON, `[
```
louischan-oursky: Not sure if [this](https://stackoverflow.com/questions/3503102/what-are-top-level-json-arrays-and-why-are-they-a-security-risk) is still valid?

kiootic: This is not sensitive information, so there's no need to be concerned about it.

pkg/auth/handler/gear/templates.go
```diff
@@ -0,0 +1,50 @@
...
+
+type TemplatesHandler struct {
+	TemplateEngine *template.Engine `dependency:"TemplateEngine"`
+}
+
+func (h TemplatesHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	encoder := json.NewEncoder(rw)
+	rw.Header().Set("Content-Type", "application/json")
+
+	specs := make([]template.Spec, len(h.TemplateEngine.TemplateSpecs))
```
carmenlau: The TemplateEngine here has the app tenant config that has ability to query app's templates. Currently it is okay that we are just displaying the template specs which is not app specific information. To avoid we update this handler incorrectly in future, maybe we can choose either:

1. Add comment here to make sure we aware this handler is public, and should not return any templates that related to the app. Or
1. Inject a template engine that has empty tenant config for this handler?

kiootic: In future, the information returned by this handler may indeed depends on the tenant config. For example, we may want to hide templates for disabled features. Therefore, I suggest we keep it as is. Also, at the moment the public interface of `TemplateEngine` cannot be used to access raw tenant config, so there's not much chance to be misused.

PR #1133, opened 2019-11-27
Universal oauth flow

### LINKED ISSUE #1130: OAuth flow in native apps
In the OAuth flow, we use cookie to store OIDC nonce. Therefore, the flow must be performed by the same user agent. However, in native apps, we have to open external browser to perform the flow. The external browser does not share the same cookie jar with the SDK.

We can first generate an URL which is served by the Auth Gear. The SDK open the URL with the external browser. The Auth Gear handles the request by a 302 to redirect the user to the provider authentication screen. The endpoint is GET so we have to put all information in the query string.

Notably, the information we have to include are

- API Key. API Key is required to generate session with correct APIClientConfiguration.
- Access Token. Access Token is required for `link`ing provider identity with authenticated user. It is not required in `login`

Are there any potential (security) problems with the above approach?

@kiootic @carmenlau

louischan-oursky:
fix #1130


pkg/auth/dependency/sso/utils.go
```diff
@@ -41,3 +35,31 @@ func DecodeState(secret string, encoded string) (*State, error) {
...
+
+type CodeClaims struct {
+	SkygearAuthorizationCode
+	jwt.StandardClaims
+}
+
+func EncodeSkygearAuthorizationCode(secret string, code SkygearAuthorizationCode) (string, error) {
+	claims := CodeClaims{
+		code,
+		jwt.StandardClaims{},
```
kiootic: We should at least set `exp` and `aud` in the JWT.

pkg/auth/handler/sso/auth_handler.go
```diff
@@ -74,19 +73,16 @@ func (p AuthRequestPayload) Validate() error {
...
 // For ux_mode is 'mobile_app',
-// it creates a 302 response, and Location points to:
-// myapp://user.skygear.io/sso/{provider}/auth_handler?result=
+// it returns a 302 response with Location set to
+// <callback_url>?x-skgyear-result=...
 //
 // Fox ux_mode is 'web_redirect',
-// it creates a 302 response, and Location points to: sso_callback_url
-// and set cookie in the response.
+// it returns a 302 response with Location set to
```
kiootic: Seems there is no difference between two flow. Should we merge them together, so that we'll have `redirect` and `popup` UX mode?

pkg/auth/handler/sso/auth_redirect.go
```diff
@@ -0,0 +1,105 @@
...
+		return
+	}
+
+	encodedState := r.Form.Get("state")
+	state, err := h.SSOProvider.DecodeState(encodedState)
+	if err != nil {
+		return
+	}
+
+	// Always generate a new nonce to ensure it is unpredictable.
```
kiootic: I think this comment need some updates; auth_url will not touches cookies anymore?

pkg/auth/dependency/sso/callback_url.go
```diff
@@ -0,0 +1,24 @@
...
+
+func ValidateCallbackURL(allowedCallbackURLs []string, callbackURL string) (err error) {
+	// The logic of this function must be in sync with the inline javascript implementation.
+	if callbackURL == "" {
+		err = errors.New("missing callback URL")
+		return
+	}
+
+	for _, v := range allowedCallbackURLs {
+		if strings.HasPrefix(callbackURL, v) {
```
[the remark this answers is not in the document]
louischan-oursky: Let me research if common OAuth providers use prefix match or exact match. We should follow them.

pkg/auth/handler/sso/auth_handler.go
louischan-oursky: I guess keeping them separate is not harmful.

pkg/auth/handler/sso/auth_redirect.go
louischan-oursky: auth_url still writes nonce cookie.

kiootic: I mean the paragraph below, calling `auth_url` multiple times should have no effect.

pkg/auth/dependency/sso/callback_url.go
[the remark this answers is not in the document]
louischan-oursky: ## Google

If the value is "http://localhost:8000", then any path is matched.
if the value contains path, then it is exact match.

## Azure Active Directory

Exact match.

---

Let us use exact match.

louischan-oursky:
@kiootic comments addressed.

pkg/auth/dependency/sso/auth_handler_html.go
```diff
@@ -16,55 +16,28 @@ func NewAuthHandlerHTMLProvider(APIEndPoint *url.URL) AuthHandlerHTMLProvider {
...
 	return false;
 }
 
 function postSSOResultMessageToWindow(windowObject, authorizedURLs) {
-	var resultStr = cookieJarGet("sso_data");
-	cookieJarDelete("sso_data");
-	var data = resultStr && JSON.parse(atob(resultStr));
-	var callbackURL = data && data.callback_url;
-	var result = data && data.result;
+	var callbackURL = {{ .callback_url }};
```
kiootic: I think this should be quoted?

pkg/auth/dependency/sso/utils.go
```diff
@@ -1,35 +1,52 @@
...
-type CustomClaims struct {
+type StateClaims struct {
 	State
 	jwt.StandardClaims
 }
 
-// NewState constructs a new state
-func NewState(params GetURLParams) State {
-	return params.State
+const jwtAudience = "skygear-server"
```
kiootic: The audience of the JWT should be the specific app, not all apps. We should use app ID/name instead.

pkg/auth/dependency/sso/auth_handler_html.go
louischan-oursky: https://play.golang.org/p/dOP141YAnYB

louischan-oursky:
@kiootic comments addressed.

PR #1134, opened 2019-11-27
Update asset gear config & behavior

kiootic:
For local development purpose, it is easier to setup a minio server and use S3 backend of asset gear.

Also, when CORS headers is returned by storage backend, we should pass-through these headers directly in the reverse proxy. For asset gear in standalone mode, removing headers added by middleware would be enough. However for asset gear behind gateway, we'll need some way to instruct gateway to pass-through the CORS headers from asset gear (TODO).

### DIFF
--- a/pkg/asset/handler/get.go
+++ b/pkg/asset/handler/get.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
+	"strings"
 
 	"github.com/gorilla/mux"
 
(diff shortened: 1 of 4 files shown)


pkg/asset/handler/get.go
```diff
@@ -114,6 +115,13 @@ func (h *GetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			return nil
 		}
 
+		// Remove existing CORS header in response
```
louischan-oursky: Leave a TODO here to support custom Access-Control-* headers

kiootic: I think we already support them? The proprietary headers mapping contains the CORS header.

louischan-oursky: I meant your TODO in this PR description. To support custom CORS header in gateway.

kiootic: I think I'll open an issue instead, since I'm not sure where should it be implemented.

kiootic: ref #1135

PR #1138, opened 2019-11-28
Use SHA256 instead for template item hash

### LINKED ISSUE #1137: Use SHA256 instead of MD5 for template item hash
Web platform does not support MD5 in native API. We can use SHA256 instead.

kiootic:
fix #1137


louischan-oursky:
https://docs.docker.com/registry/spec/api/#content-digests

If we ever want to change the hashing algorithm, we can follow their encoding. `<scheme>:<value>`

PR #1143, opened 2019-12-02
Improve existing CORS header handling

### LINKED ISSUE #1135: Customize CORS header handling at gateway
Currently, the gateway would eat all CORS header from upstream (gears, microservices, etc.), and replace them as specified in tenant config. In some situations, this may not be desirable. For example, we'd like to pass through CORS header from asset gears. We'll need some mechanism to customize CORS header handling at gateway.

kiootic:
fix #1135

### DIFF
--- a/pkg/asset/handler/get.go
+++ b/pkg/asset/handler/get.go
@@ -115,17 +115,17 @@ func (h *GetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			return nil
 		}
 
-		// Remove existing CORS header in response
-		for name := range w.Header() {
-			if strings.HasPrefix(name, "Access-Control-") {
-				w.Header().Del(name)
-			}
-		}
-
 		resp.Header = h.CloudStorageProvider.ProprietaryToStandard(resp.Header)
 		// Do not support range request if image processing query is present.
 		if hasPipeline {
 			resp.Header.Del("Accept-Ranges")
+		}
+
+		// Delete existing CORS header if exists in response
+		for name := range resp.Header {
+			if strings.HasPrefix(name, "Access-Control-") {
+				w.Header().Del(name)
+			}
 		}
 
 		// Check access
--- a/pkg/asset/handler/get.go
+++ b/pkg/asset/handler/get.go
@@ -6,7 +6,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
-	"strings"
 
 	"github.com/gorilla/mux"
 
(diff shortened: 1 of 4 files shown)


pkg/asset/handler/get.go
```diff
@@ -115,19 +115,19 @@ func (h *GetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		}
-
 		resp.Header = h.CloudStorageProvider.ProprietaryToStandard(resp.Header)
 		// Do not support range request if image processing query is present.
 		if hasPipeline {
 			resp.Header.Del("Accept-Ranges")
 		}
 
+		// Delete existing CORS header if exists in response
+		for name := range resp.Header {
```
louischan-oursky: If `Access-Control-` is ever found in upstream response, we must remove all of our CORS headers. The rationale is that if the upstream has any CORS headers, then we treat them as complete and do not merge with ours.

pkg/gateway/handler/deployment_route.go
```diff
@@ -39,8 +38,18 @@ func newDeploymentRouteReverseProxy() *httputil.ReverseProxy {
 		// Remove tenant config from header.
 		coreConfig.WriteTenantConfig(req, nil)
 	}
+	modifyResponse := func(resp *http.Response) error {
+		// Remove CORS headers if upstream provides them
+		for name := range resp.Header {
```
louischan-oursky: Same here.

pkg/gateway/handler/gear.go
```diff
@@ -33,16 +33,16 @@ func newGearReverseProxy() *httputil.ReverseProxy {
...
-	modifyResponse := func(r *http.Response) error {
-		// Remove CORS headers because they are managed by this gateway.
-		// Auth gear in standalone mode mounts CORS middleware.
-		// If we do not remove CORS headers, then the headers will duplicate.
-		r.Header.Del("Access-Control-Allow-Origin")
-		r.Header.Del("Access-Control-Allow-Credentials")
-		r.Header.Del("Access-Control-Allow-Methods")
-		r.Header.Del("Access-Control-Allow-Headers")
+	modifyResponse := func(resp *http.Response) error {
+		// Remove CORS headers if upstream provides them
```
louischan-oursky: Same here.

pkg/core/http/cors.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"net/http"
+	"strings"
+)
+
+func FixupCORSHeaders(downstream http.ResponseWriter, upstream *http.Response) {
+	hasCORSHeaders := false
+	for name := range upstream.Header {
+		if strings.HasPrefix(name, "Access-Control-") {
+			hasCORSHeaders = true
```
louischan-oursky: We can `break` here to speed up.

PR #1157, opened 2019-12-09
Add default HTML template

### LINKED ISSUE #1153: Add default HTML email templates
Currently we only provide default TXT email templates.

louischan-oursky:
fix #1153

### DIFF
--- a/pkg/auth/dependency/forgotpwdemail/template.go
+++ b/pkg/auth/dependency/forgotpwdemail/template.go
@@ -29,6 +29,17 @@ Thanks.`,
 var TemplatePasswordEmailHTML = template.Spec{
 	Type:   TemplateItemTypeForgotPasswordEmailHTML,
 	IsHTML: true,
+	Default: `<!DOCTYPE html>
+<html>
+<body>
+<p>Dear {{ .email }},</p>
+<p>You received this email because someone tries to reset your account password on {{ .appname }}. To reset your account password, click this link:</p>
+<p>{{ .link }}</p>
+<p>If you did not request to reset your account password, Please ignore this email.</p>
+<p>Thanks.</p>
+</body>
+</html>
+`,
 }
 
 var TemplateForgotPasswordResetHTML = template.Spec{
--- a/pkg/auth/dependency/mfa/template.go
+++ b/pkg/auth/dependency/mfa/template.go
@@ -24,4 +24,11 @@ var TemplateMFAOOBCodeEmailTXT = template.Spec{
 var TemplateMFAOOBCodeEmailHTML = template.Spec{
 	Type:   TemplateItemTypeMFAOOBCodeEmailHTML,
 	IsHTML: true,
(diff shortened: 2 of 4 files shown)


pkg/auth/dependency/forgotpwdemail/template.go
```diff
@@ -29,6 +29,17 @@ Thanks.`,
 var TemplatePasswordEmailHTML = template.Spec{
 	Type:   TemplateItemTypeForgotPasswordEmailHTML,
 	IsHTML: true,
+	Default: `<!DOCTYPE html>
+<html>
+<body>
+<p>Dear {{ .email }},</p>
```
chpapa: I think this is actually a pretty good example / idea of why if we put the first login ID as display name in metadata could be quite useful.

e.g. I would change it to:

<p> Dear {{ .metadata.displayID }}, </p> 

(assume `.metadata.displayID` is the way we put a metadata var here...) It would works really well...

pkg/auth/dependency/forgotpwdemail/template.go
```diff
@@ -29,6 +29,17 @@ Thanks.`,
...
+<p>You received this email because someone tries to reset your account password on {{ .appname }}. To reset your account password, click this link:</p>
+<p>{{ .link }}</p>
```
chpapa: {{ .link }} will put a HTMl link here? Or just the URL?

pkg/auth/dependency/mfa/template.go
```diff
@@ -24,4 +24,11 @@ var TemplateMFAOOBCodeEmailTXT = template.Spec{
 var TemplateMFAOOBCodeEmailHTML = template.Spec{
 	Type:   TemplateItemTypeMFAOOBCodeEmailHTML,
 	IsHTML: true,
+	Default: `<!DOCTYPE html>
+<html>
+<body>
+<p>Your MFA code is: {{ .code }}</p>
```
chpapa: Your {{ .appname }} Two Factor Auth Verification code is: {{ .code }}

Please enter the Verification Code on the Sign-in screen. 

Please ignore this code if this Sign-In was not initated by you.

pkg/auth/dependency/forgotpwdemail/template.go
louischan-oursky: The related feature issue is here https://github.com/SkygearIO/features/issues/388

I guess we don't have anything to do at this stage.

pkg/auth/dependency/forgotpwdemail/template.go
louischan-oursky: https://github.com/SkygearIO/skygear-server/issues/1161

pkg/auth/dependency/mfa/template.go
louischan-oursky: https://github.com/SkygearIO/skygear-server/issues/1161

PR #1159, opened 2019-12-11
Sign in with Apple

### LINKED ISSUE #1117: Support apple id login

louischan-oursky:
fix #1117 

The other platforms integration is just OIDC.

The iOS integration deviates from the authorization code flow so I added two endpoints.
In particular, we pass both the nonce and the hashed nonce to the client while in the authorization code flow, we pass the nonce via HTTP only cookie.

The corresponding iOS integration is here. https://gist.github.com/louischan-oursky/6ca2f70eee957343be7ff3d2eb7d4107


pkg/auth/dependency/sso/azureadv2.go
```diff
@@ -135,12 +135,12 @@ func (f *Azureadv2Impl) OpenIDConnectGetAuthInfo(r OAuthAuthorizationResponse, s
...
 	c, err := f.getOpenIDConfiguration()
 	if err != nil {
-		err = errors.Newf("failed to get OIDC discovery document: %w", err)
+		err = NewSSOFailed(NetworkFailed, "failed to get OIDC discovery document")
 		return
 	}
 	keySet, err := f.getKeys(c.JWKSUri)
 	if err != nil {
-		err = errors.Newf("failed to get OIDC JWKS: %w", err)
+		err = NewSSOFailed(NetworkFailed, "failed to get OIDC JWTs")
```
kiootic: It should be JWKs

pkg/auth/dependency/sso/apple.go
```diff
@@ -0,0 +1,207 @@
...
+		err = NewSSOFailed(InvalidParams, "no nonce")
+		return
+	}
+	if subtle.ConstantTimeCompare([]byte(hashedNonce), []byte(crypto.SHA256String(r.Nonce))) != 1 {
+		err = NewSSOFailed(SSOUnauthorized, "invalid nonce")
+		return
+	}
+
+	// Verify the issuer
+	if !mapClaims.VerifyIssuer("https://appleid.apple.com", true) {
```
kiootic: The referenced document says the issuer should 'contains' the constant string. I'm not sure if it'll be some other strings like `https://appleid.apple.com/abc` in future.

pkg/auth/handler/sso/apple_request.go
```diff
@@ -0,0 +1,223 @@
...
+		p.MergeRealm = password.DefaultRealm
+	}
+
+	if p.OnUserDuplicate == "" {
+		p.OnUserDuplicate = model.OnUserDuplicateDefault
+	}
+}
+
+func (p *AppleRequestPayload) Validate() []validation.ErrorCause {
+	if !p.PasswordAuthProvider.IsRealmValid(p.MergeRealm) {
```
kiootic: We should validate the provider is Apple ID.

pkg/auth/handler/sso/apple_response.go
```diff
@@ -0,0 +1,178 @@
...
+
+	err = hook.WithTx(h.HookProvider, h.TxContext, func() (err error) {
+		result, err = h.Handle(payload)
+		return
+	})
+	h.AuthnSessionProvider.WriteResponse(w, result, err)
+}
+
+func (h *AppleResponseHandler) Handle(payload *AppleResponseRequest) (result interface{}, err error) {
+	if h.OAuthProvider == nil {
```
kiootic: We should validate the provider is Apple ID.

pkg/auth/dependency/sso/apple.go
louischan-oursky: The wording is a bit confusing. In my real testing, the value of issuer is exactly `https://appleid.apple.com`. If I change to use `strings.HasPrefix`, do you have any concerns?

kiootic: My concern is that the document explicitly used the word 'contains' for `iss` field, and the word 'is' is used for exact matching the `aud` field. It seems to me that they anticipated the need to change it in future. I think using `strings.Contains` would be safer.

louischan-oursky:
@kiootic Removed the two endpoints and added UXModeManual.

PR #1173, opened 2019-12-18
Make HookProvider a scoped dependency

### LINKED ISSUE #1172: after_session_create hook event is not dispatched
`after_session_create` event is dispatched in `authnsession.providerImpl` through injected `hook.Provider`. However, `hook.Provider` is stateful but not shared between the dependencies. Therefore the handler and `authnsession.providerImpl` got different `hook.Provider`, and the `after_session_create` event is gone.

We'll need some mechanisms to ensure stateful dependencies is instantiated only once within a request (i.e. scoped dependencies in DI terms).

kiootic:
fix #1172 

NOTE: we may want to revisit our DI system design in future to make these problems easier to notice.

### DIFF
--- a/pkg/core/inject/context.go
+++ b/pkg/core/inject/context.go
@@ -1,0 +1,72 @@
+package inject
+
+import (
+	"context"
+)
+
+type contextKeyType struct{}
+
+var contextKey = contextKeyType{}
+
+type injectContext struct {
+	Root         *injectContext
+	Parent       *injectContext
+	Dependencies map[string]interface{}
+}
+
+func WithInject(ctx context.Context) context.Context {
+	injectCtx := &injectContext{Dependencies: map[string]interface{}{}}
+
+	if baseCtx, ok := ctx.Value(contextKey).(*injectContext); ok {
+		injectCtx.Parent = baseCtx
+		injectCtx.Root = baseCtx.Root
+	} else {
+		injectCtx.Parent = nil
+		injectCtx.Root = injectCtx
+	}
+
+	return context.WithValue(ctx, contextKey, injectCtx)
+}
+
+func getContext(ctx context.Context) *injectContext {
+	injectCtx, _ := ctx.Value(contextKey).(*injectContext)
+	return injectCtx
+}
+
+func Singleton(ctx context.Context, name string, factory func() interface{}) func() interface{} {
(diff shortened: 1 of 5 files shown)


pkg/core/inject/context.go
```diff
@@ -0,0 +1,72 @@
...
+
+func WithInject(ctx context.Context) context.Context {
+	injectCtx := &injectContext{Dependencies: map[string]interface{}{}}
+
+	if baseCtx, ok := ctx.Value(contextKey).(*injectContext); ok {
+		injectCtx.Parent = baseCtx
+		injectCtx.Root = baseCtx.Root
+	} else {
+		injectCtx.Parent = nil
+		injectCtx.Root = injectCtx
```
carmenlau: We design the inject context to support getting the same dependency from Root by using `Singleton` function, but we haven't used it, right? Since we called `WithInject ` in `DefaultRequestInject`, but the context is used within the `DefaultRequestInject` only. If this is the case, maybe leave some comment in `DefaultRequestInject` function?

kiootic: In our current usage:
- `Transient` and `Scoped` would work correctly.
- `Singleton` would act same as `Scoped`, since the request does not have a parent inject context. We shouldn't use it for now.
- Context of request is not updated with the inject context. Handlers should not use these function with request context.

I think I'll add comments to the functions to indicate the current limitations.

louischan-oursky: I guess parent is not being used now?

kiootic: Yes, I was thinking whether scoped dependencies should be resolved from all contexts in the chain, or just the current context. For simplicity, it is resolved in current context at the moment.

pkg/core/inject/context.go
```diff
@@ -50,10 +54,16 @@ func Singleton(ctx context.Context, name string, factory func() interface{}) fun
 	}
 }
 
+// Transient returns a newly created named dependency, existing singleton and
+// scoped dependencies are ignored.
 func Transient(ctx context.Context, name string, factory func() interface{}) func() interface{} {
 	return func() interface{} { return factory() }
 }
 
+// Singleton returns the named dependency from current context, and construct
```
louischan-oursky: Singleton -> Scoped

PR #1178, opened 2019-12-30
Implement login ID APIs

### LINKED ISSUE #974: Implement Identity APIs
- [ ] addLoginID
- [ ] removeLoginID
- [ ] updateLoginID
- [ ] listIdentities
- [x] unlinkSSO

kiootic:
fix #974

### DIFF
--- a/pkg/auth/dependency/userverify/provider.go
+++ b/pkg/auth/dependency/userverify/provider.go
@@ -20,6 +20,11 @@ type Provider interface {
 		authInfo *authinfo.AuthInfo,
 		code string,
 	) (*VerifyCode, error)
+	UpdateVerificationState(
+		authInfo *authinfo.AuthInfo,
+		authStore authinfo.Store,
+		principals []*password.Principal,
+	) error
 }
 
 type providerImpl struct {
@@ -144,16 +149,32 @@ func (provider *providerImpl) markUserVerified(
 
 	// Update user
 	authInfo.VerifyInfo[verifyCode.LoginID] = true
-	authInfo.Verified = IsUserVerified(
+	if err = provider.UpdateVerificationState(authInfo, authStore, principals); err != nil {
+		return
+	}
+
+	return
+}
+
+func (provider *providerImpl) UpdateVerificationState(
+	authInfo *authinfo.AuthInfo,
+	authStore authinfo.Store,
+	principals []*password.Principal,
+) error {
+	isVerified := IsUserVerified(
 		authInfo.VerifyInfo,
 		principals,
 		provider.config.Criteria,
 		provider.config.LoginIDKeys,
 	)
-
-	if err = authStore.UpdateAuth(authInfo); err != nil {
-		return
+	if isVerified == authInfo.Verified {
+		return nil
 	}
 
-	return
+	authInfo.Verified = isVerified
+	if err := authStore.UpdateAuth(authInfo); err != nil {
+		return err
+	}
+
+	return nil
 }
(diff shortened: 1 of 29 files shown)


pkg/auth/handler/loginid/add.go
```diff
@@ -0,0 +1,171 @@
...
+	password.LoginID
+}
+
+// @JSONSchema
+const AddLoginIDRequestSchema = `
+{
+	"$id": "#AddLoginIDRequest",
+	"type": "object",
+	"properties": {
+		"key": { "type": "string", "minLength": 1 },
```
carmenlau: Should we use `login_id_key` and `login_id` for consistency?

pkg/auth/handler/loginid/add.go
```diff
@@ -0,0 +1,171 @@
...
+		err = validateLoginIDs(h.PasswordAuthProvider, loginIDs, newLoginIDIndex)
+		if err != nil {
+			return err
+		}
+
+		newPrincipal, err := h.PasswordAuthProvider.MakePrincipal(userID, "", newLoginID, password.DefaultRealm)
+		if err != nil {
+			return err
+		}
+		newPrincipal.HashedPassword = principals[0].HashedPassword
```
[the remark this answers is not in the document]
kiootic: You're right. I can identify two problems for users without existing password principals at the moment:
1. We cannot fill in a correct hashed password, and user cannot change password.
2. Misconfiguration to required login ID amount may cause inability to add login ID.

For problem 1, the previously discussed solution is to create a password principals without a valid password hash (i.e. user cannot login using it), and password can be changed by re-authenticating using SSO. I'll implement the first part first in this PR.

For problem 2, we have error message showing the login ID amount constraint violation (e.g. not enough 'email' login ID), so developer can understand what's going on and fix the config (i.e. reduce the minimum amount of 'email' login ID to 0). Note that this would not occur in default configuration.

pkg/auth/handler/loginid/add.go
kiootic: This shape (`key` and `value` for login IDs) is previously used for signup API, so I don't think we need to change it.

`login_id_key` and `login_id` is used in login API, since using `key` and `value` would be unclear. It is clear in this context that `key` and `value` referred to login ID.

pkg/auth/dependency/userverify/provider.go
```diff
@@ -144,16 +149,32 @@ func (provider *providerImpl) markUserVerified(
...
+
+	return
+}
+
+func (provider *providerImpl) UpdateVerificationState(
+	authInfo *authinfo.AuthInfo,
+	authStore authinfo.Store,
+	principals []*password.Principal,
+) error {
+	isVerified := IsUserVerified(
```
louischan-oursky: I think we need to remove the login ID from authInfo.VerifyInfo if the login ID is removed or updated.

kiootic: I suppose VerifyInfo would contains the email addresses that the user possessed (i.e. verified emails), therefore the user would be verified if the verified email is removed and added as login ID again.

louischan-oursky: This is unspecified in the feature spec. We should discuss this with the team. Maybe we open a PR to update the related spec and continue the discussion there?

kiootic: The user verification feature is not specified (I opened feature#336 for tracking but it is closed). I'll add some concerns to discuss there.

PR #1187, opened 2020-01-03
Use more friendly error message

### LINKED ISSUE #1147: Design a way to provide helpful error message of invalid format
When the kind of the cause is `StringFormat`, the message is generic and looks like `Does not match format 'AppName'`. We want to explain what `AppName` is.

### LINKED ISSUE #1186: Allow names to be used for email sender/reply to address
We switched to use JSON schema format validation for emails in tenant config in #1185. However, our format checker does not allows names in email, which developer may want to have for email sender/reply to address. We need special handling for them.

kiootic:
fix #1147
fix #1186


pkg/core/validation/format.go
```diff
@@ -50,37 +51,41 @@ type URL struct {
...
-		return false
+		return errors.New("input URL must not have query/fragment")
 	}
 
 	p := ""
 	switch f.URLVariant {
 	case URLVariantFullOnly:
 		if u.Scheme == "" || u.Host == "" {
-			return false
+			return errors.New("input URL must be complete")
```
louischan-oursky: Maybe `must be absolute` is a more accurate term?

kiootic: Sure. I'll then use `input URL must be normalized` for the `filepath.Clean` check below.

louischan-oursky: I mean in [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.3) absolute URI is defined and its meaning matches the case here. The error message should be `input URL must be absolute`

kiootic:
@louischan-oursky updated the format checking logic, please review.

pkg/auth/handler/forgotpwd/forgot_password.go
```diff
@@ -55,7 +55,7 @@ const ForgotPasswordRequestSchema = `
 	"$id": "#ForgotPasswordRequest",
 	"type": "object",
 	"properties": {
-		"email": { "type": "string", "format": "email", "minLength": 1 }
+		"email": { "type": "string", "format": "email" }
```
louischan-oursky: It is better keep `minLength` because empty string is valid Go zero value.

Given a struct

```golang
type Params struct {
  OptionalEmail string `json:"email"`
}
```

It is difficult to write a schema for this struct while following Go convention to use string zero value to present absence of value.

kiootic: I think we're passing the raw JSON to the validator, so the Go convention should not matter?

louischan-oursky: Oh, you're right. We have updated all validation code to validate against the raw JSON.

PR #1189, opened 2020-01-06
Remove minimum login ID amount config

kiootic:


louischan-oursky:
Landed in https://github.com/SkygearIO/skygear-server/commit/aef181ac3f45af73f6de930086af0cf22eafae97

PR #1191, opened 2020-01-06
Remove login ID from VerifyInfo when it is removed

kiootic:


pkg/auth/handler/loginid/remove.go
```diff
@@ -160,6 +160,9 @@ func (h RemoveLoginIDHandler) Handle(w http.ResponseWriter, r *http.Request) err
 		}
 		user := model.NewUser(*authInfo, userProfile)
 
+		if h.UserVerificationProvider.IsVerifiable(p.LoginIDKey) {
```
carmenlau: Maybe we don't need to check `IsVerifiable`, and always delete the login id if it exists in the `VerifyInfo`?

PR #1192, opened 2020-01-06
Allow adding multiple login IDs

kiootic:


pkg/auth/handler/loginid/add_test.go
```diff
@@ -110,7 +112,7 @@ func TestAddLoginIDHandler(t *testing.T) {
 					"code": 400,
 					"info": {
 						"causes": [
-							{ "kind": "General", "message": "login ID key is not allowed", "pointer": "/key" }
+							{ "kind": "General", "message": "login ID key is not allowed", "pointer": "/0/key" }
```
louischan-oursky: Shouldn't the pointer be `/login_ids/0/key`?

kiootic:
@carmenlau Are you testing using SDK example in Chrome?

carmenlau:
I tried calling the api with http client directly, let me try in SDK too.

carmenlau:
I found I configured my http client incorrectly, the apis work! Thanks @kiootic

PR #1193, opened 2020-01-06
Rename UserConfiguration to AppConfiguration

louischan-oursky:
fix https://github.com/SkygearIO/skygear-server/issues/1190

### DIFF
--- a/pkg/core/config/schema.go
+++ b/pkg/core/config/schema.go
@@ -28,7 +28,7 @@ const (
 			"app_id": { "$ref": "#NonEmptyString" },
 			"app_name": { "$ref": "#NonEmptyString" },
 			"database_config": { "$ref": "#DatabaseConfiguration" },
-			"hook": { "$ref": "#HookTimeoutConfiguration" },
+			"hook": { "$ref": "#HookTenantConfiguration" },
 			"app_config": { "$ref": "#AppConfiguration" }
 		},
 		"required": ["version", "app_id", "app_name", "database_config", "app_config"]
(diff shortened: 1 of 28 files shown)


carmenlau:
@kiootic See if the new TenantConfiguration looks good to you too? Since we will not want to change it in future. Database and hook config are changed in this PR.

pkg/core/config/schema.go
```diff
@@ -24,22 +24,33 @@ const (
...
+	},
+	"DatabaseConfiguration": {
+		"$id": "#DatabaseConfiguration",
+		"type": "object",
+		"properties": {
+			"database_url": { "$ref": "#NonEmptyString" },
+			"database_schema": { "$ref": "#NonEmptyString" }
+		}
+	},
+	"HookTimeoutConfiguration": {
```
kiootic: I think it should not be named `HookTimeoutConfiguration`, since it may store other hook related configuration in future.

pkg/core/config/tenant.go
```diff
@@ -896,7 +896,7 @@ func (c NexmoConfiguration) IsValid() bool {
 	return c.APIKey != "" && c.APISecret != ""
 }
 
-type HookAppConfiguration struct {
+type HookTimeoutConfiguration struct {
```
kiootic: I think we may store other hook related configuration here, so maybe use another name?

carmenlau: Originally we have `HookAppConfiguration` and `HookUserConfiguration`, how about change it as followed?
`HookAppConfiguration` -> `HookTenantConfiguration`
`HookUserConfiguration` -> `HookAppConfiguration`

Any better suggestion?

pkg/core/config/schema.go
louischan-oursky: Can you suggest a good name? HookConfiguration is already taken :(

pkg/core/config/tenant.go
louischan-oursky: OK. Let's use Carmen's suggestion.

louischan-oursky:
@carmenlau Updated the hook configuration type name.

PR #1197, opened 2020-01-08
Handle concurrent write conflict

### LINKED ISSUE #1195: Set repeatable read isolation level
To mitigate issues caused by concurrent requests, we should set isolation level = repeatable read for transactions for now. If update failed, we can respond with HTTP Conflict status code and let client decide whether to retry request.

kiootic:
fix #1195

### DIFF
--- a/pkg/core/db/sql_executor.go
+++ b/pkg/core/db/sql_executor.go
@@ -93,6 +93,8 @@ func (e *SQLExecutor) QueryRowWith(sqlizeri sq.Sqlizer) (*sqlx.Row, error) {
 func isWriteConflict(err error) bool {
 	var pqErr *pq.Error
 	if errors.As(err, &pqErr) {
+		// 40001: serialization_failure
+		// 40P01: deadlock_detected
 		return pqErr.Code == "40001" || pqErr.Code == "40P01"
 	}
 	return false
--- a/pkg/core/db/sql_executor.go
+++ b/pkg/core/db/sql_executor.go
@@ -20,6 +20,7 @@ import (
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/jmoiron/sqlx"
+	"github.com/lib/pq"
 
 	"github.com/skygeario/skygear-server/pkg/core/errors"
 )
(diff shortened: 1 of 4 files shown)


pkg/core/db/sql_executor.go
```diff
@@ -75,7 +82,18 @@ func (e *SQLExecutor) QueryRowWith(sqlizeri sq.Sqlizer) (*sqlx.Row, error) {
...
+		}
 		return nil, errors.WithDetails(err, errors.Details{"sql": errors.SafeDetail.Value(sql)})
 	}
 	return db.QueryRowxContext(e.context, sql, args...), nil
 }
+
+func isWriteConflict(err error) bool {
+	var pqErr *pq.Error
+	if errors.As(err, &pqErr) {
+		return pqErr.Code == "40001" || pqErr.Code == "40P01"
```
louischan-oursky: Just curious. How did you find out these are the error codes we want to detect?

pkg/core/middleware/recover.go
```diff
@@ -33,10 +34,15 @@ func (m RecoverMiddleware) Handle(next http.Handler) http.Handler {
...
-				handled := false
+				const errorTypeUnexpected = 0
+				const errorTypeHandled = 1
+				const errorTypeConflict = 2
+				errorType := errorTypeUnexpected
 				if herr, ok := err.(handler.HandledError); ok {
-					handled = true
+					errorType = errorTypeHandled
 					err = herr.Error
+				} else if err == db.ErrWriteConflict {
```
louischan-oursky: errors.Is

pkg/core/db/sql_executor.go
kiootic: These are the error codes corresponded to the two errors mentioned in [documentation page on isolation level](https://www.postgresql.org/docs/9.5/transaction-iso.html).

louischan-oursky: 40P01 is not mentioned in that page. Where can I find it?

kiootic: You can find the error codes in https://www.postgresql.org/docs/9.1/errcodes-appendix.html. 40P01 is deadlock error, which is may indicate conflicts between running transactions.

louischan-oursky: Can you include these in the code as comments?

PR #1198, opened 2020-01-08
Update user verification feature

### LINKED ISSUE #1194: Update user verification feature
ref SkygearIO/features#336

kiootic:
fix #1194

### DIFF
--- a/pkg/auth/dependency/hook/mutator_impl.go
+++ b/pkg/auth/dependency/hook/mutator_impl.go
@@ -45,8 +45,8 @@ func (mutator *mutatorImpl) New(ev *event.Event, user *model.User) Mutator {
 }
 
 func (mutator *mutatorImpl) Add(mutations event.Mutations) error {
-	// update raw verified status if needed
-	if mutations.VerifyInfo != nil {
+	// update computed verified status if needed
+	if mutations.VerifyInfo != nil || mutations.IsManuallyVerified != nil {
 		// update IsVerified
 		if mutator.UserPasswordPrincipals == nil {
 			principals, err := mutator.PasswordAuthProvider.GetPrincipalsByUserID(mutator.User.ID)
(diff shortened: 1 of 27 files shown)


pkg/auth/dependency/hook/mutator_impl.go
```diff
@@ -104,12 +108,23 @@ func (mutator *mutatorImpl) Apply() error {
...
-	if mutations.IsVerified != nil {
-		authInfo.Verified = *mutations.IsVerified
+	if mutations.IsManuallyVerified != nil {
+		authInfo.ManuallyVerified = *mutations.IsManuallyVerified
 	}
 	if mutations.VerifyInfo != nil {
 		authInfo.VerifyInfo = *mutations.VerifyInfo
 	}
+
+	if mutations.IsManuallyVerified != nil || mutations.VerifyInfo != nil {
```
carmenlau: Seems we don't need to recompute the verified if user change the `IsManuallyVerified ` only? Maybe we should choose either,
1. We don't need this check and always recompute
1. Recompute only if `mutations.VerifyInfo != nil` ?

kiootic: Yes you're right, `Verified` flag in auth info does not depends on `ManuallyVerified`. I'll make it so that it is recomputed only if `VerifyInfo` is mutated.

pkg/auth/event/user_update.go
```diff
@@ -104,10 +104,11 @@ func (event UserUpdateEvent) WithMutationsApplied(mutations Mutations) UserAware
 	}
 	if mutations.VerifyInfo != nil {
 		newEvent.VerifyInfo = mutations.VerifyInfo
-		// IsVerified will be updated by mutator, if it is not also mutated
+		// IsComputedVerified will be updated by mutator
 	}
-	if mutations.IsVerified != nil {
-		newEvent.IsVerified = mutations.IsVerified
+	if mutations.IsComputedVerified != nil && mutations.IsManuallyVerified != nil {
```
carmenlau: `mutations.IsComputedVerified` will be non nil, if `VerifyInfo` or `IsManuallyVerified` change. So If user change `VerifyInfo` info only and `IsManuallyVerified ` will be nil. And this checking is incorrect that we should have `IsVerified ` in the new event for this case, right?

PR #1200, opened 2020-01-13
Rename version to api_version

### LINKED ISSUE #1199: Add api_version to TenantConfiguration

louischan-oursky:
fix #1199

### DIFF
--- a/pkg/core/apiversion/apiversion.go
+++ b/pkg/core/apiversion/apiversion.go
@@ -1,6 +1,7 @@
 package apiversion
 
 import (
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"strconv"
@@ -10,12 +11,27 @@ import (
 const MajorVersion = 2
 
 // MinorVersion is the current minor API Version.
-const MinorVersion = 0
+const MinorVersion = 1
 
 // APIVersion is the current API Version.
 var APIVersion = Format(MajorVersion, MinorVersion)
 
+// SupportedVersionsJSON is an JSON array of supported versions.
+var SupportedVersionsJSON string
+
 var regexpAPIVersion = regexp.MustCompile(`^v(\d+)\.(\d+)$`)
+
+func init() {
+	var supportedVersions []string
+	for i := 0; i <= MinorVersion; i++ {
+		supportedVersions = append(supportedVersions, Format(MajorVersion, i))
+	}
+	bytes, err := json.Marshal(supportedVersions)
+	if err != nil {
+		panic(err)
+	}
+	SupportedVersionsJSON = string(bytes)
+}
 
 // Format formats major and minor into `v<major>.<minor>`.
 func Format(major, minor int) string {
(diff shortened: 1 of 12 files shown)


pkg/core/middleware/apiversion_test.go
```diff
@@ -0,0 +1,54 @@
...
+					"code": 404,
+					"message": "incompatible API version",
+					"name": "NotFound",
+					"reason": "IncompatibleAPIVersion"
+				}
+			}
+			`)
+		}
+
+		r, _ := http.NewRequest("GET", "/v3.1/foobar", nil)
```
carmenlau: Please add one more test for given minor version less than the server minor version

pkg/core/server/server.go
```diff
@@ -41,14 +41,18 @@ func NewServerWithOption(
...
+		router = router.PathPrefix(option.GearPathPrefix).Subrouter()
 	}
+
+	if option.IsAPIVersioned {
+		router = router.PathPrefix("/{api_version}").Subrouter()
+	}
+
 	srv := Server{
-		router: subRouter,
+		router: router,
```
carmenlau: The `/healthz` check endpoint become 404 after this change

pkg/core/apiversion/apiversion.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+	"regexp"
+	"strconv"
+)
+
+// MajorVersion is the current major API Version.
+const MajorVersion = 2
+
+// MinorVersion is the current minor API Version.
+const MinorVersion = 0
```
kiootic: So we start from 2.0? I was thinking of following the milestone version (i.e. currently 2.1) for easier tracking.

pkg/core/config/schema.go
```diff
@@ -24,14 +24,14 @@ const (
 		"$id": "#TenantConfiguration",
 		"type": "object",
 		"properties": {
-			"version": { "const": "2" },
+			"api_version": { "enum": ["v2.0"] },
```
kiootic: Can we use the version string defined in `apiversion`? It seems quite easy to miss when updating the version.

pkg/core/middleware/apiversion.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (m *APIVersionMiddleware) Handle(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		apiVersion := mux.Vars(r)[m.APIVersionName]
+		major, minor, ok := apiversion.Parse(apiVersion)
+
+		if !ok || major != m.MajorVersion || minor > m.MinorVersion {
+			handler.WriteResponse(w, handler.APIResponse{
+				Error: ErrIncompatibleAPIVersion,
```
kiootic: We may want to include the expected version, so the clients know which version they should use.

louischan-oursky:
comments addressed.

PR #1204, opened 2020-01-15
Adjust static asset gateway proxying

kiootic:
Static assets goes through many services:
```
Client -> Gateway (static asset) -> Gateway (asset gear) -> Asset Gear
```

Before this fix, some duplicated headers can be observed:
- Request ID (x2)
- Vary: Origin (x3)

### DIFF
--- a/pkg/core/http/cors.go
+++ b/pkg/core/http/cors.go
@@ -41,6 +41,8 @@ func FixupCORSHeaders(downstream http.ResponseWriter, upstream *http.Response) {
 
 	// Upstream has provided CORS header; upstream will manage all CORS headers
 	// Remove existing CORS headers from response to downstream
+	// `Vary: Origin` is also used to control CORS policy, so upstream
+	// should manage it themselves.
 	headers := downstream.Header()
 	for name, values := range headers {
 		if strings.HasPrefix(name, "Access-Control-") {
--- a/pkg/core/http/cors.go
+++ b/pkg/core/http/cors.go
@@ -5,12 +5,33 @@ import (
 	"strings"
 )
 
+func parseVaryHeaders(values []string) []string {
+	var headers []string
+	for _, v := range values {
+		for _, h := range strings.Split(v, ",") {
+			h = strings.TrimSpace(h)
+			if h != "" {
+				headers = append(headers, h)
+			}
+		}
+	}
+	return headers
+}
+
 func FixupCORSHeaders(downstream http.ResponseWriter, upstream *http.Response) {
 	hasCORSHeaders := false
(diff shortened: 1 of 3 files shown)


pkg/core/http/cors.go
```diff
@@ -21,9 +42,26 @@ func FixupCORSHeaders(downstream http.ResponseWriter, upstream *http.Response) {
...
+					n++
+				}
+			}
+			varyHeaders = varyHeaders[:n]
+			if len(varyHeaders) > 0 {
+				headers[name] = []string{strings.Join(varyHeaders, ",")}
+			} else {
+				delete(headers, name)
+			}
+		}
```
louischan-oursky: Could you add some tests for the new behavior?

pkg/core/http/cors.go
```diff
@@ -21,9 +42,26 @@ func FixupCORSHeaders(downstream http.ResponseWriter, upstream *http.Response) {
 	// Upstream has provided CORS header; upstream will manage all CORS headers
 	// Remove existing CORS headers from response to downstream
 	headers := downstream.Header()
-	for name := range headers {
+	for name, values := range headers {
 		if strings.HasPrefix(name, "Access-Control-") {
 			headers.Del(name)
 		}
+		// Delete 'Vary: Origin' header
+		if name == "Vary" {
```
carmenlau: Would like to in this change, are we want to avoid duplicate value in `Vary` header? Why we need to target `Origin`? Or we should get the union of upstream  and downstream vary list, and remove them from the downstream vary list?

kiootic: Since this function targets headers injected by CORS middleware, so its main purpose is to remove duplicated CORS related headers (i.e. `Vary: Origin`). We'll need to revisit header handling in future if gateway would inject other headers in the response.

carmenlau: Since gateway will proxy to microservice, assume microservice response headers has `Access-Control-` only but not `Vary`, `hasCORSHeaders` will become true. Then we will remove `Origin` from down stream. Am I understand correctly and is it a problem?

kiootic: `Vary: Origin` indicates some response header may change according to request Origin. In our case, we may emit different `Access-Control-Allowed-Origin` header. If microservices has chosen to manage the CORS header themselves, they are responsible for emitted correct CORS related headers, including `Vary: Origin` if necessary.

carmenlau: Would you mind to add some comment here we have this assumption?
LGTM, thanks:)

kiootic: Comment added.

PR #1215, opened 2020-02-19
Auth UI

louischan-oursky:


louischan-oursky:
This PR is too big. I am going to split this into small PRs.

PR #1224, opened 2020-02-28
Simplify gateway error messages

### LINKED ISSUE #1203: Should show appropriate error when tenant config failed to validate
Currently it shows `App not found` which is difficult to debug, maybe we can change it to `Invalid app config`?

kiootic:
fix #1203

### DIFF
--- a/pkg/gateway/middleware/app.go
+++ b/pkg/gateway/middleware/app.go
@@ -27,17 +27,19 @@ func (f FindAppMiddleware) Handle(next http.Handler) http.Handler {
 		host := r.Host
 		app := gatewayModel.App{}
 		if err := f.Store.GetAppByDomain(host, &app); err != nil {
-			if !store.IsNotFound(err) {
+			if store.IsNotFound(err) {
+				http.Error(w, "Not found", http.StatusNotFound)
+			} else {
 				logger.WithError(err).Error("failed to find app")
+				http.Error(w, "Internal server error", http.StatusInternalServerError)
 			}
-			http.Error(w, "App not found", http.StatusBadRequest)
 			return
 		}
 
 		routes, err := f.Store.GetLastDeploymentRoutes(app)
 		if err != nil {
 			logger.WithError(err).Error("failed to get deployment routes")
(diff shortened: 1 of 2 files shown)


pkg/gateway/middleware/deployment_route.go
```diff
@@ -23,7 +23,7 @@ func (f FindDeploymentRouteMiddleware) Handle(next http.Handler) http.Handler {
 
 		match := model.MatchRoute(path, app.Config.DeploymentRoutes)
 		if match == nil {
-			http.Error(w, "Fail to match deployment route", http.StatusNotFound)
+			http.Error(w, "Not found", http.StatusNotFound)
```
carmenlau: I think we should keep the detail like `Deployment route not found`? If developer see the message `Not found`, there will be chance that he call a wrong app ("App not found") or wrong path ("Deployment route not found"). More detail may be helpful for debugging?

pkg/gateway/middleware/app.go
```diff
@@ -27,17 +27,19 @@ func (f FindAppMiddleware) Handle(next http.Handler) http.Handler {
 		host := r.Host
 		app := gatewayModel.App{}
 		if err := f.Store.GetAppByDomain(host, &app); err != nil {
-			if !store.IsNotFound(err) {
+			if store.IsNotFound(err) {
+				http.Error(w, "Not found", http.StatusNotFound)
```
carmenlau: Should we use `App not found`?

pkg/gateway/middleware/deployment_route.go
kiootic: I'm not sure if using richer message would cause potential information disclosure problem. For example in nginx, generic 404 error would be returned regardless of whether the host/path is not matched.

In my opinion:
- Errors like `Deployment route not found` are not user-friendly, since these terms is unfamiliar to developers, and may mislead developer.
- These errors can be simply resolved by double-checking the URL, so developer should be able to troubleshoot it easily. For advanced cases (e.g. path matching problems), additional information would not help.

PR #1228, opened 2020-02-28
Deploy static asset to S3 bucket

louischan-oursky:


.travis.yml
```diff
@@ -62,6 +67,19 @@ deploy:
       repo: SkygearIO/skygear-server
       branch: master
       go: 1.13.3
+  - provider: s3
+    access_key_id: "$AWS_ACCESS_KEY"
+    secret_access_key: "$AWS_SECRET_KEY"
```
carmenlau: About the setup,
1. Would like to know is those key ready in travis?
2. When I visit `code.skygear.dev`, seems it is not ready yet. Open another issue for completing the setup?

.travis.yml
```diff
@@ -62,6 +67,19 @@ deploy:
...
+    # This bucket is created in region us-east-1, same as code.skygear.io
+    # aws --region=us-east-1 s3api create-bucket --acl public-read --bucket "code.skygear.dev" --no-object-lock-enabled-for-bucket
+    bucket: "code.skygear.dev"
+    skip_cleanup: true
+    local_dir: ./scripts/deploy-asset/dist
+    upload-dir: dist
```
carmenlau: Better name for the folder(e.g. `authui`, `skygear-server` or another name...?) ?  We may have another components need to deploy static assets under domain `code.skygear.dev` in future?

louischan-oursky: My intention is to use this script to upload any assets. So the actual path should look like `dist/authui/css/main.css`, `dist/authui/js/main.js` and so on.

.travis.yml
louischan-oursky: > 1

Yes

> 2

Will set up this today.

louischan-oursky:
@carmenlau The cloudfront setup is finished.

.travis.yml
carmenlau: For every build, we will have `dist/authui/...` and `dist/git-[commithash]/authui/....`. And we will update the docker build to use the asset with the same commit hash in future. Am i correct? Looks good to me!

Will we use this bucket to host static asset from another repo? e.g. js sdk?

louischan-oursky: > Am i correct?

Yes. I guess we will add the static asset url prefix to images.yaml so we have one place to update all git commit hashes.

> Will we use this bucket to host static asset from another repo?

Good question. if it is the case, then `upload-dir` should be different.

```
/dist/git-<skygear-server-hash>/authui/css/main.css
/sdk-js/git-<sdk-js-hash>/...
```

Maybe we should rename `/dist` to `/skygear-server` now?

carmenlau: `skygear-server` looks good to me

louischan-oursky:
@carmenlau Updated the `upload-dir` to `skygear-server`

PR #1231, opened 2020-02-28
Refactor template engine

louischan-oursky:


pkg/core/template/engine.go
```diff
@@ -50,18 +50,30 @@ func (e *Engine) Register(spec Spec) {
...
+	templateBody, spec, err := e.resolveTemplate(templateType, resolveOptions)
 	if err != nil {
 		return
 	}
 	if spec.IsHTML {
-		return RenderHTMLTemplate(string(templateType), templateBody, context)
+		return RenderHTMLTemplate(RenderOptions{
+			Name:          string(templateType),
+			TemplateBody:  templateBody,
+			Defines:       spec.Defines,
```
kiootic: When would `Defines` be used?

louischan-oursky: To break the elements of Auth UI into several parts. And reuse them in the main page.

kiootic: Can user influence the content of `Defines` list? If yes, a billion laugh attack is possible.

louischan-oursky: > Can user influence the content of Defines list?

No. The Defines list is defined by Spec which is a compile time constant.

PR #1232, opened 2020-02-28
Setup wire

louischan-oursky:


pkg/auth/handler/wire.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/google/wire"
+
+	"github.com/skygeario/skygear-server/pkg/core/config"
+)
+
+func ProvideTenantConfigPtr(r *http.Request) *config.TenantConfiguration {
+	return config.GetTenantConfig(r.Context())
+}
+
+func ProvideTenantConfig(r *http.Request) config.TenantConfiguration {
```
kiootic: If we're refactoring/changing the DI mechanism, we may as well as unify to use either pointer/value, but not both. Is that possible?

louischan-oursky: Yes

louischan-oursky: Removed the provider of struct and added notes.

PR #1234, opened 2020-03-05
Remove server

louischan-oursky:


louischan-oursky:
> But I found another problem that we only used IsAPIVersioned in controller and haven't abled IsAPIVersioned in gears? When we discuss about migrating the existing app, maybe we should bring this up.

Seems it is better to move this to controller itself. Let me remove it too.

---

Actually the gear path prefix is also going to go away because auth and asset will have their own domain. The `/_auth` and `/_asset` rewrite should be a impl details of gateway.

carmenlau:
> Seems it is better to move this to controller itself. Let me remove it too. 

I am thinking do we need to have the `api_version` thing in gears too when we update the route? We may want similar protection in gear, like server should reject newer client (SDK), some endpoint maybe missing if the server is too old? This will happens when user deploy standalone auth gear. I think it needs discussion. 

Removing `IsAPIVersioned` option is fine, as long as the api version middleware. So the gear setup can be defined in its `main.go`, use api version or not.

carmenlau:
Sine the comment is not related to this PR, I will merge it first

Issue #1241, opened 2020-03-10
Update session mechanism


kiootic:
- Reorganize session mechanism into auth gear package
    - Update session model
    - Save session token in cookie
- Add Session Resolver endpoint
    - Resolve session to AuthnInfo

PR #1244, opened 2020-03-10
Refactor session mechanism

kiootic:
ref #1241


pkg/auth/dependency/session/context.go
```diff
@@ -1,5 +1,38 @@
...
 
+import (
+	"context"
+
+	"github.com/skygeario/skygear-server/pkg/core/auth/authinfo"
+)
+
 type Context interface {
 	Session() *Session
+	User() *authinfo.AuthInfo
```
louischan-oursky: Are we going to replace AuthContextGetter with this interface?

pkg/auth/dependency/session/cookie.go
```diff
@@ -0,0 +1,39 @@
...
+
+	corehttp.UpdateCookie(rw, cookieSession)
+}
+
+func ClearCookie(rw http.ResponseWriter, useInsecureCookie bool) {
+	corehttp.UpdateCookie(rw, &http.Cookie{
+		Name:     CookieName,
+		Path:     "/",
+		HttpOnly: true,
+		Secure:   !useInsecureCookie,
```
louischan-oursky: Though it has no effect, we should better keep `SameSite: http.SameSiteLaxMode` here.

pkg/auth/dependency/session/context.go
kiootic: Yes, I planned to have three interface for auth contexts:
- `session.Context`: contains user & session, used in auth gear only (in auth)
- `auth.AccessKeyContext`: contains API key (i.e. client ID), used in needed places (in core)
- `authn.Context`: contains `authn.Info`, used in everywhere (in core)

pkg/auth/handler/session/resolve.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/skygeario/skygear-server/pkg/auth"
+	"github.com/skygeario/skygear-server/pkg/auth/dependency/session"
+)
+
+func AttachResolveHandler(
+	router *mux.Router,
+	authDependency auth.DependencyMap,
+) {
+	router.NewRoute().
+		Path("/session/get").
```
carmenlau: Should be `/session/resolve`?

pkg/auth/dependency/session/provider_impl.go
```diff
@@ -59,7 +63,6 @@ func (p *ProviderImpl) MakeSession(authnSess *auth.AuthnSession) (*Session, stri
 	// if session has new fields.
 	session := &Session{
 		ID:                      uuid.New(),
-		ClientID:                authnSess.ClientID,
```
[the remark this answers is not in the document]
kiootic: This is the IdP session, which is unrelated to specific client. We'll have a separate Access Grant struct to associate access token with specific client.

PR #1248, opened 2020-03-11
Remove audit trail

louischan-oursky:


louischan-oursky:
Removed `UserAuditConfiguration` in the schema. Thanks!

PR #1259, opened 2020-03-13
Refactor API key mechanism

### LINKED ISSUE #1256: Refactor API Key mechanism
- new context for api key
- store resolved client config in context
- only include api key resolve middleware if needed

kiootic:
fix #1256 
depends on #1244


louischan-oursky:
@kiootic Can you rebase?

pkg/auth/handler/sso/auth_handler.go
```diff
@@ -153,8 +154,8 @@ func (h AuthHandler) Handle(w http.ResponseWriter, r *http.Request) (success boo
 	}
 
 	// Extract API Key from state
-	key := h.APIClientConfigurationProvider.GetAccessKeyByClientID(state.APIClientID)
-	h.AuthContextSetter.SetAccessKey(key)
+	client, _ := model.GetClientConfig(h.TenantConfiguration.AppConfig.Clients, state.APIClientID)
+	r = r.WithContext(coreAuth.WithAccessKey(r.Context(), coreAuth.AccessKey{Client: client}))
```
louischan-oursky: I guess this is problematic because the `ctx` received by the dependencies are the context WITHOUT access key. There is no way to update the `ctx` of the dependencies.

pkg/core/auth/authz/policy/accesskey_test.go
```diff
@@ -4,72 +4,61 @@ import (
...
-				mAccessKey: model.AccessKey{Type: model.MasterAccessKeyType},
-			}
-
+		Convey("should return error if auth context has master key ", func() {
+			req = req.WithContext(auth.WithAccessKey(req.Context(), auth.AccessKey{
+				IsMasterKey: true,
+			}))
 			err := DenyNoAccessKey(req, ctx)
-			So(err, ShouldBeEmpty)
+			So(err, ShouldNotBeEmpty)
```
louischan-oursky: So DenyNoAccessKey does not accept master key?

pkg/core/auth/access_key.go
```diff
@@ -0,0 +1,65 @@
...
+func WithAccessKey(ctx context.Context, ak AccessKey) context.Context {
+	if key, ok := ctx.Value(accessKeyContextKey).(*AccessKey); ok {
+		*key = ak
+		return ctx
+	} else {
+		return context.WithValue(ctx, accessKeyContextKey, &ak)
+	}
+}
+
+func GetAccessKey(ctx context.Context) AccessKey {
```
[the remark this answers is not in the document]
kiootic: Client is a type alias to map in current implementation, so it can be nil.

pkg/auth/handler/sso/auth_handler.go
kiootic: Yes, this is problematic. Originally I store a plain `AccessKey` in context. To workaround this problem for now, a pointer is stored in context to act as a value container, so that `WithAccessKey` can update the existing context.

I planned to refactor it (after authentication session), so that access key in context is only written by middlewares, and read by handlers. Providers should accept the access key (i.e. OAuthClientConfiguration) explicitly through parameters.

pkg/core/auth/authz/policy/accesskey_test.go
kiootic: Yes. Most use of access key is to obtain a client configuration (e.g. session configuration in authentication). Master key does not provide these information so it cannot be used.

pkg/auth/handler/sso/auth_handler.go
kiootic: Also, the access key container middleware is put in all handler (same as authn middleware), and I plan to make it per-handler as well.

pkg/core/auth/authz/policy/accesskey_test.go
louischan-oursky: What about changing the name from DenyNoAccessKey to RequireClient?

kiootic: Updated.

PR #1262, opened 2020-03-16
Refactor authentication mechanism

### LINKED ISSUE #1253: Refactor authentication session mechanism
- use new session mechanism
- decouple session writing & authn session token handling
- update some API for testing

kiootic:
fix #1253


pkg/auth/model/identity.go
```diff
@@ -28,6 +28,7 @@ type Identity struct {
 	Claims principal.Claims
 }
 
+// TODO(model): remove unused param
```
louischan-oursky: Thanks!!!

pkg/auth/handler/forgotpwd/reset_form.go
```diff
@@ -257,15 +257,12 @@ func (h ForgotPasswordResetFormHandler) HandleResetSuccess(rw http.ResponseWrite
...
 
 func (h ForgotPasswordResetFormHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	_, err := handler.Transactional(h.TxContext, func() (_ interface{}, err error) {
-		err = h.Handle(w, r)
-		if err == nil {
-			err = h.HookProvider.WillCommitTx()
-		}
-		return
+	h.TxContext.UseHook(h.HookProvider)
+	_, err := handler.Transactional(h.TxContext, func() (interface{}, error) {
```
louischan-oursky: I have an idea to have a further cleanup here.

- Move `db.WithTx` to `TxContext.Do`
- Move `handler.Transactional` to `TxContext.DoInterface`

What do you think?

pkg/core/async/queue.go
```diff
@@ -4,27 +4,35 @@ import (
...
 
 	"github.com/skygeario/skygear-server/pkg/core/config"
+	"github.com/skygeario/skygear-server/pkg/core/db"
 )
 
 type Queue interface {
-	Enqueue(name string, param interface{}, response chan error)
+	Enqueue(spec TaskSpec)
+	WillCommitTx() error
+	DidCommitTx()
```
louischan-oursky: Embed the `TransactionHook` interface instead of duplicating its methods?

pkg/auth/dependency/authn/deps.go
```diff
@@ -0,0 +1,101 @@
...
+		UserProfileStore:   us,
+		IdentityProvider:   ip,
+		HookProvider:       hp,
+	}
+}
+
+var DependencySet = wire.NewSet(
+	ProvideSignupProcess,
+	ProvideAuthenticateProcess,
+	wire.Struct(new(OAuthCoordinator), "*"),
```
louischan-oursky: Wow, this saves us from writing many boilerplate ProvideXXX functions!

pkg/auth/handler/wire.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/skygeario/skygear-server/pkg/core/handler"
+	"github.com/skygeario/skygear-server/pkg/core/validation"
+)
+
+func provideLoginHandler(
+	requireAuthz handler.RequireAuthz,
+	v *validation.Validator,
+	ap LoginAuthnProvider,
+	tx db.TxContext,
+) http.Handler {
```
louischan-oursky: I guess the return type of this function and `auth.MakeHandler` are problematic. Wire injects by type but in `/pkg/auth/handler` we have more than one `http.Handler`.

pkg/auth/handler/forgotpwd/reset_form.go
kiootic: I think there's no need to move these. These functions are implemented using public API of `TxContext` (like extension method in C#), so their implementation should not be moved into specific implementation of `TxContext`.

pkg/auth/handler/wire.go
kiootic: For each handler inject, there'll two function in `wire.go`:
- `provideXXX`: instantiate handler using injected dependencies, returning `http.Handler`
- `newXXX`: specifies available dependencies and instantiate handler, returning `http.Handler`

Since each `newXXX` will specify only their corresponding `provideXXX` in the dependency set, there is only one provider for `http.Handler` (i.e. `provideXXX`), so there would be no conflict.

pkg/core/async/queue.go
kiootic: I'm following Go's convention here of defining the interface only on consumer side, to avoid coupling provider interface with what consumer need.

kiootic:
@louischan-oursky 
The authentication session mechanism is refactored. Only authentication work at the moment, need to fix session mechanism in next PR.

pkg/core/http/cookie.go
```diff
@@ -23,7 +23,7 @@ func (c *CookieConfiguration) WriteTo(rw http.ResponseWriter, value string) {
 		SameSite: http.SameSiteLaxMode,
 	}
 
-	cookie.Value = c.Name
```
louischan-oursky: Horrible typo😂

pkg/auth/dependency/authn/session.go
```diff
@@ -8,9 +8,14 @@ type SessionStep string
 
 const (
 	SessionStepIdentity SessionStep = "identity"
-	SessionStepMFA      SessionStep = "mfa"
+	SessionStepMFASetup SessionStep = "mfa.setup"
+	SessionStepMFAAuthn SessionStep = "mfa.authn"
```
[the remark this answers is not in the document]
kiootic: opened SkygearIO/skygear-SDK-JS#555

pkg/auth/dependency/authn/session_container.go
```diff
@@ -0,0 +1,7 @@
+package authn
+
+import "github.com/skygeario/skygear-server/pkg/auth/dependency/session"
+
+type SessionContainer interface {
```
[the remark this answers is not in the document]
kiootic: Yes the name is not good. I'm working on session, and realized that there should be two interfaces and three structs:
- structs
    - `auth/authn.Session`: authentication session, has attrs, stateless (no ID)
    - `auth/session.Session`: IdP session, has attrs & extra data (e.g. user agent), stateful (has ID)
    - OIDC offline grant: has attrs & extra data (e.g. user agent), stateful (has ID)
- interfaces
    - `core/authn.SessionAttributer` (go's convention is add -er): has attrs only, for use in authentication session steps
    - `core/authn.Session`: has ID, attrs & extra data (e.g. user agent), for general use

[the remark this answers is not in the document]
louischan-oursky: OK. Let's keep SessionContainer and rename it in another PR.

carmenlau:
Sorry that I don't have enough time to read the whole PR detailedly. If it looks good to you, feel free to merge it first, I will catch up this later when have time. Thanks a lot! @louischan-oursky @kiootic 🙇🏻‍♀️

Issue #1261, opened 2020-03-16
Auth UI Web App - Login and Signup


louischan-oursky:
- [x] Render templates with `html/template`
- [x] Support styling with developer-provided CSS
- [x] Derive `frame-ancestor` from all `redirect_uris`
- [x] Implement the root endpoint
  - [x] The sign in page
    - [x] Switching between different login ID **types**
      - [x] Email login ID
      - [x] Phone login ID
      - [x] Username login ID
      - [x] Raw login ID
    - [x] Validate
       - [x] Text login ID
       - [x] Calling code and phone number
  - [x] The sign in password page
    - [x] Error
    - [x] (JS) Toggle password visibility
  - [x] The sign up page
    - [x] Switching between different login ID **keys**
      - [x] Email login ID
      - [x] Phone login ID
      - [x] Username login ID
      - [x] Raw login ID
    - [x] Validate
  - [x] The sign up password page
    - [x] Toggle visibility of password
    - [x] Display a list of password requirements
      - [x] (JS) Show which requirement is fulfilled and which is not

PR #1263, opened 2020-03-17
Auth UI Web App - Part 1

louischan-oursky:
ref #1261


pkg/auth/dependency/webapp/csp_middleware_test.go
```diff
@@ -0,0 +1,79 @@
...
+				},
+			}
+			w := httptest.NewRecorder()
+			r, _ := http.NewRequest("GET", "/", nil)
+			h.ServeHTTP(w, r)
+
+			So(w.Result().Header.Get("Content-Security-Policy"), ShouldEqual, "frame-ancestors https://example.com https://app.com 'self';")
+		})
+
+		Convey("include only https? redirect URIs", func() {
```
kiootic: We should not allow HTTP origin to embed us as iframe, since it is not secure.

pkg/auth/dependency/webapp/validate_provider.go
```diff
@@ -0,0 +1,13 @@
+package webapp
+
+import (
+	"net/url"
+)
+
+// ValidateProvider validates url.Values with JSON Schema.
+type ValidateProvider interface {
+	// Prevalidate fills in default values.
+	Prevalidate(form url.Values)
```
kiootic: Maybe name it `SetDefaults` instead? `Prevalidate` is rather confusing.

louischan-oursky: But it also removes empty values.

pkg/auth/dependency/webapp/csp_middleware_test.go
louischan-oursky: Sorry. `http` should only be allowed under this condition. https://github.com/SkygearIO/features/pull/425/files#diff-9176bd53b9cb2e7662018c59f2cdc01aR169

For example `localhost`, `127.0.0.1` and `skygear.localhost` are allowed but `abclocalhost` is not. Any objections?

kiootic: I think that's okay.

pkg/auth/dependency/webapp/validate_provider.go
kiootic: How about `PrepareValues`? It prepares the form values for validation & using in provider.

Issue #1255, opened 2020-03-12
Implement basic OIDC function


kiootic:
- authorization & grant db schema
- authorize endpoint
- token endpoint
- userinfo endpoint
- revoke endpoint
- discovery documents

PR #1266, opened 2020-03-19
Add basic OIDC metadata & authorize endpoint

kiootic:
It is planned to integrate with Auth UI using this flow (suppose not authenticated):
1. user redirected to authorize endpoint
2. authorize endpoint check authentication, and redirect to Auth UI, with redirect_uri=original request URL
3. Auth UI perform authentication
4. Auth UI redirect to original request URL (i.e. retry authorize request)
5. authorize endpoint check authentication, and ask user consent if needed(NYI)
6. authorize endpoint generate code and redirect to app

@louischan-oursky please let me know if this fits your work.

ref #1255


pkg/auth/dependency/oauth/handler/handler_authz.go
```diff
@@ -168,3 +174,43 @@ func (h *AuthorizationHandler) parseScopes(scope string) ([]string, error) {
...
+	// Authorization of requested scopes not granted, requesting consent.
+	// TODO(oauth): request consent, for now just always implicitly grant scopes.
+	if authz == nil {
+		authz = &oauth.Authorization{
+			ID:        uuid.New(),
+			AppID:     h.AppID,
+			ClientID:  r.ClientID(),
+			UserID:    userID,
+			CreatedAt: h.Time.NowUTC(),
+			UpdatedAt: h.Time.NowUTC(),
```
louischan-oursky: Use a single `now`

pkg/auth/dependency/oauth/handler/handler_authz.go
```diff
@@ -87,7 +92,7 @@ func (h *AuthorizationHandler) doHandle(
 		SessionID:       session.SessionID(),
 
 		CreatedAt: h.Time.NowUTC(),
-		ExpireAt:  h.Time.NowUTC(),
+		ExpireAt:  h.Time.NowUTC().Add(1 * gotime.Minute),
```
louischan-oursky: 1 min seems too short. Maybe 3 or 5 minutes?

pkg/auth/handler/oauth/authorize.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+func AttachAuthorizeHandler(
+	router *mux.Router,
+	authDependency pkg.DependencyMap,
+) {
+	router.NewRoute().
+		Path("/oauth2/authorize").
+		Handler(pkg.MakeHandler(authDependency, newAuthorizeHandler)).
+		Methods("GET")
```
louischan-oursky: Should support both GET and POST https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

pkg/auth/handler/oauth/metadata.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (h *MetadataHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	meta := map[string]interface{}{}
+	for _, provider := range h.metaProviders {
+		provider.PopulateMetadata(meta)
+	}
+
+	encoder := json.NewEncoder(rw)
+	err := encoder.Encode(meta)
```
louischan-oursky: It's better to include content-length.

migrations/auth/1584511915_add_oidc_tables.down.sql
```diff
@@ -1 +1 @@
-DROP TABLE _auth_oauth_authorizations;
+DROP TABLE _auth_oauth_authorization;
```
louischan-oursky: Squash this commit with the very first commit.

migrations/auth/1584511915_add_oidc_tables.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE _auth_oauth_authorizations (
+  id TEXT PRIMARY KEY,
+  app_id TEXT NOT NULL,
+  client_id TEXT NOT NULL,
+  user_id TEXT NOT NULL REFERENCES _core_user(id),
+  created_at TIMESTAMP WITHOUT TIME ZONE NOT NULL,
+  updated_at TIMESTAMP WITHOUT TIME ZONE NOT NULL,
+  scopes JSONB NOT NULL,
+  UNIQUE (client_id, user_id)
```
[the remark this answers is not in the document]
kiootic: I think I'll include `app_id` in the unique constraint instead.

pkg/auth/dependency/oauth/store_grant.go
```diff
@@ -0,0 +1,25 @@
...
+
+type CodeGrantStore interface {
+	GetCodeGrant(codeHash string) (*CodeGrant, error)
+	CreateCodeGrant(*CodeGrant) error
+	UpdateCodeGrant(*CodeGrant) error
+	DeleteCodeGrant(*CodeGrant) error
+}
+
+type OfflineGrantStore interface {
+	GetOfflineGrant(id string) (*OfflineGrant, error)
```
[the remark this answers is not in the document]
kiootic: Offline grants are very similar to sessions, that session APIs are able to manipulate them. Therefore its implementation is very much like session's, that it can be get using ID (and token hash is verified separately by consumer). The refresh token would be encoded like session, i.e. offline grant ID + grant token hash.

pkg/auth/dependency/oauth/proto_authz.go
```diff
@@ -0,0 +1,25 @@
+package oauth
+
+type AuthorizationRequest map[string]string
```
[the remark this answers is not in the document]
kiootic: `AuthorizationRequest` is a model type for the OAuth, which treat the request parameters as simple map of keys and values. The wire format (`application/x-www-form-urlencoded`) allows duplicate keys, but OAuth does not specify how to handle it. Therefore, I think it is better to accept a simple map for OAuth, and let the handler handle the duplication case.

pkg/auth/dependency/oauth/result_authz.go
```diff
@@ -0,0 +1,69 @@
...
+	if a.RedirectURI != nil {
+		query := a.RedirectURI.Query()
+		for k, v := range a.Response {
+			query.Add(k, v)
+		}
+		a.RedirectURI.RawQuery = query.Encode()
+		http.Redirect(rw, r, a.RedirectURI.String(), http.StatusFound)
+	} else {
+		rw.WriteHeader(http.StatusBadRequest)
+		rw.Write([]byte("Invalid OAuth authorization request:\n"))
```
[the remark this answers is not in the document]
kiootic: I just tried locally and it automagically set a correct value for `Content-Length`.

pkg/auth/dependency/oauth/handler/handler_authz.go
kiootic: I'll use 5 minutes for now.

pkg/auth/handler/oauth/metadata.go
kiootic: I just tried and the Content-Length is also set automagically here...

pkg/auth/dependency/oauth/result_authz.go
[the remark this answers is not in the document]
louischan-oursky: From golang doc

```
    // If WriteHeader has not yet been called, Write calls
    // WriteHeader(http.StatusOK) before writing the data. If the Header
    // does not contain a Content-Type line, Write adds a Content-Type set
    // to the result of passing the initial 512 bytes of written data to
    // DetectContentType. Additionally, if the total size of all written
    // data is under a few KB and there are no Flush calls, the
    // Content-Length header is added automatically.
```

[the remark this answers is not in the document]
louischan-oursky: Very magical indeed.

pkg/auth/dependency/oauth/handler/handler_authz.go
```diff
@@ -167,7 +167,7 @@ func (h *AuthorizationHandler) validateRequest(r protocol.AuthorizationRequest)
 	if r.ResponseType() != "code" {
 		return protocol.NewError("unsupported_response_type", "only 'code' response type is supported")
 	}
-	if r.Scope() == "" {
+	if len(r.Scope()) == 1 && r.Scope()[0] == "" {
```
louischan-oursky: `r.Scope()` should return nil or empty slice if `scope` is absent or `scope=`. The current behavior is too surprising. This applies to `ui_locales` as well.

kiootic: Fixed.

PR #1270, opened 2020-03-20
Implement OIDC token endpoint & JWKS

kiootic:
Tested with AppAuth and everything seems fine.

ref #1255

### DIFF
--- a/pkg/auth/dependency/oauth/handler/handler_token.go
+++ b/pkg/auth/dependency/oauth/handler/handler_token.go
@@ -1,0 +1,167 @@
+package handler
+
+import (
+	"context"
+	"crypto/sha256"
+	"crypto/subtle"
+	"encoding/base64"
+	"errors"
+
+	"github.com/sirupsen/logrus"
+	"github.com/skygeario/skygear-server/pkg/auth/dependency/oauth"
+	"github.com/skygeario/skygear-server/pkg/auth/dependency/oauth/protocol"
+	"github.com/skygeario/skygear-server/pkg/auth/dependency/session"
+	"github.com/skygeario/skygear-server/pkg/core/config"
+	"github.com/skygeario/skygear-server/pkg/core/time"
+)
+
+type TokenHandler struct {
+	Context context.Context
+	Clients []config.OAuthClientConfiguration
+	Logger  *logrus.Entry
+
+	Authorizations oauth.AuthorizationStore
+	CodeGrants     oauth.CodeGrantStore
+	OfflineGrants  oauth.OfflineGrantStore
+	AccessGrants   oauth.AccessGrantStore
+	Sessions       session.Provider
+	GenerateToken  TokenGenerator
+	Time           time.Provider
+}
+
+func (h *TokenHandler) Handle(r protocol.TokenRequest) TokenResult {
+	client, errResp := h.resolveClient(r)
+	if errResp != nil {
+		return tokenResultError{Response: errResp}
+	}
+
+	result, err := h.doHandle(client, r)
+	if err != nil {
+		var oauthError *protocol.OAuthProtocolError
+		resultErr := tokenResultError{}
+		if errors.As(err, &oauthError) {
+			resultErr.Response = oauthError.Response
+		} else {
+			h.Logger.WithError(err).Error("authz handler failed")
+			resultErr.Response = protocol.NewErrorResponse("server_error", "internal server error")
+			resultErr.InternalError = true
+		}
+		result = resultErr
+	}
+
+	return result
+}
+
+var errInvalidAuthzCode = protocol.NewError("invalid_grant", "invalid authorization code")
+
+func (h *TokenHandler) doHandle(
+	client config.OAuthClientConfiguration,
+	r protocol.TokenRequest,
+) (TokenResult, error) {
+	if err := h.validateRequest(r); err != nil {
+		return nil, err
+	}
+
+	codeHash := hashToken(r.Code())
+	codeGrant, err := h.CodeGrants.GetCodeGrant(codeHash)
+	if errors.Is(err, oauth.ErrGrantNotFound) {
+		return nil, errInvalidAuthzCode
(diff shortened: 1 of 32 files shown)


pkg/auth/dependency/oauth/handler/handler_token.go
```diff
@@ -69,6 +67,76 @@ func (h *TokenHandler) doHandle(
...
+		}
+	}
+	if !allowed {
+		return nil, protocol.NewErrorResponse("invalid_request", "redirect URI is not allowed")
+	}
+
+	return client, nil
+}
+
+func (h *TokenHandler) validateRequest(r protocol.TokenRequest) error {
```
louischan-oursky: We should also see whether the client is allowed to specify such `grant_type`

pkg/auth/dependency/oidc/endpoints.go
```diff
@@ -0,0 +1,11 @@
+package oidc
+
+import "net/url"
+
+type JWKSEndpointProvider interface {
+	JWKSEndpointURI() *url.URL
+}
+
+type UserInfoEndpointProvider interface {
+	TokenEndpointURI() *url.URL
```
louischan-oursky: Typo here? The method function does not match the interface name.

pkg/auth/dependency/oidc/id_token.go
```diff
@@ -35,10 +35,13 @@ func (ti *IDTokenIssuer) IssueIDToken(client config.OAuthClientConfiguration, us
 		Nonce: nonce,
 	}
 
-	key, err := jwt.ParseRSAPrivateKeyFromPEM([]byte(ti.OIDCConfig.Keys[0].PrivateKey))
+	key := ti.OIDCConfig.Keys[0]
```
louischan-oursky: Check the length of `Keys` first to avoid panic.

pkg/auth/dependency/oauth/handler/handler_token.go
```diff
@@ -204,6 +204,10 @@ func (h *TokenHandler) handleRefreshToken(
 		return nil, err
 	}
 
+	if h.Time.NowUTC().After(offlineGrant.ExpireAt) {
```
louischan-oursky: In normal case, shouldn't the offline grant be evicted from redis? We do set expiry on it. Anyway it is good to have checking here. The underlying storage may not be redis.

pkg/auth/dependency/oidc/id_token.go
```diff
@@ -0,0 +1,44 @@
...
+const CodeGrantValidDuration = 5 * gotime.Minute
+
+func (ti *IDTokenIssuer) IssueIDToken(client config.OAuthClientConfiguration, userID string, nonce string) (string, error) {
+	now := ti.Time.NowUTC()
+	token := &IDToken{
+		StandardClaims: jwt.StandardClaims{
+			Issuer:    ti.URLPrefix.Value().String(),
+			Audience:  client.ClientID(),
+			IssuedAt:  now.Unix(),
+			ExpiresAt: now.Add(CodeGrantValidDuration).Unix(),
```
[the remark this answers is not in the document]
kiootic: I think `id_token_hint` should accepts expired ID tokens as well, since it is just a hint.

pkg/auth/dependency/oidc/metadata.go
```diff
@@ -1,13 +1,24 @@
...
+	meta["issuer"] = p.URLPrefix.Value().String()
 	meta["scopes_supported"] = AllowedScopes
 	meta["subject_types_supported"] = []string{"public"}
 	meta["id_token_signing_alg_values_supported"] = []string{"RS256"}
+	meta["claims_supported"] = []string{
+		"iss",
+		"aud",
+		"iat",
+		"exp",
+		"sub",
```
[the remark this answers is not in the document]
kiootic: I think `nonce` is not exactly a claim of ID token, since it is only present iif authorization request has nonce. Other case, including refreshing existing tokens, would not return a ID token containing nonce. Also, Google's OIDC discovery document also does not have it, yet seems to support nonce.

pkg/auth/dependency/oidc/id_token.go
kiootic: I think it's better to enforce non-emptiness at config schema level? Otherwise, the whole OIDC flow cannot work.

pkg/auth/dependency/oidc/id_token.go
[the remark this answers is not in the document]
louischan-oursky: Please add this assumption as comment here.

pkg/auth/dependency/oauth/handler/handler_token.go
kiootic: Yes indeed, just a double check.

pkg/auth/dependency/oauth/handler/handler_token.go
```diff
@@ -0,0 +1,389 @@
...
+	return client, nil
+}
+
+func (h *TokenHandler) validateRequest(r protocol.TokenRequest) error {
+	switch r.GrantType() {
+	case "authorization_code":
+		if r.Code() == "" {
+			return protocol.NewError("invalid_request", "code is required")
+		}
+		if r.CodeVerifier() == "" {
```
carmenlau: For Generic OIDC RP, code verifier is not required. I think we will do it when we introduce client secret later?

https://github.com/SkygearIO/features/blob/fefdd9934bb23d23e13e9b9dc2f340895008b9c7/features/370-auth-ui/oidc.md#generic-oidc-rp

louischan-oursky:
Can we also update `standalone-tenant-config.sample.yaml` to include `oidc` configuration with a signing key?

pkg/auth/dependency/oauth/handler/handler_token.go
kiootic: Yes, either PKCE or client secret is required.

PR #1271, opened 2020-03-24
Auth UI Web App - Part 4

louischan-oursky:
ref #1261


pkg/auth/dependency/audit/password_checker.go
```diff
@@ -332,6 +343,38 @@ func (pc *PasswordChecker) ValidatePassword(payload ValidatePasswordPayload) err
 	return PasswordPolicyViolated.NewWithCauses("password policy violated", violations)
 }
 
+// PasswordPolicy outputs a list of PasswordViolation to reflect the password policy.
+func (pc *PasswordChecker) PasswordPolicy() (out []PasswordViolation) {
```
kiootic: It seems now `PasswordViolation` means a specific configured password policy, instead of an instance of violation information. I think it's better to change its name to reflect the current role, e.g. `PasswordPolicy`?

pkg/auth/dependency/authn/provider.go
```diff
@@ -45,6 +45,10 @@ func (p *Provider) SignupWithLoginIDs(
 	return p.AuthnSession.StepSession(s)
 }
 
+func (p *Provider) ValidateSignUpLoginID(loginID loginid.LoginID) error {
```
kiootic: I think we're using `Signup` in the codebase (not `SignUp`).

pkg/auth/dependency/authn/signup.go
```diff
@@ -57,6 +57,11 @@ type SignupProcess struct {
 	TaskQueue                     async.Queue
 }
 
+func (p *SignupProcess) ValidateSignUpLoginID(loginID loginid.LoginID) (err error) {
```
kiootic: Same here, `Signup` instead of `SignUp`.

pkg/auth/dependency/webapp/authenticate_provider_impl.go
```diff
@@ -40,12 +51,18 @@ func (p *AuthenticateProviderImpl) ServeHTTP(w http.ResponseWriter, r *http.Requ
...
 	switch step {
-	case "submit_password":
-		writeResponse, err = p.SubmitPassword(w, r)
 	case "submit_login_id":
 		writeResponse, err = p.SubmitLoginID(w, r)
+	case "submit_password":
+		writeResponse, err = p.SubmitPassword(w, r)
 	case "choose_idp":
 		writeResponse, err = p.ChooseIdentityProvider(w, r)
+	case "sign_up":
```
kiootic: It seems the step names are getting complex. Maybe add a string enum and add structure to the step names? e.g. `signin:submit_login_id`, `signup:submit_password`.

louischan-oursky:
Let's use the name `Signup` and `Login`.

louischan-oursky:
@kiootic updated.

kiootic:
- the button 'Sign in!' and 'Use email instead' looks different on macOS Chrome, is that ok?
- when I tried sign up, it keeps giving 'Invalid email address' error.

kiootic:
> when I tried sign up, it keeps giving 'Invalid email address' error.

It turns out I entered a email in username field, and it gives invalid email address error. So it's a misleading message. Let's fix it in another PR.

PR #1272, opened 2020-03-24
Resolve access tokens for session

kiootic:


pkg/auth/dependency/auth/access_event_test.go
```diff
@@ -0,0 +1,128 @@
...
+			So(ip, ShouldEqual, "1.1.1.1")
+		})
+		Convey("should resolve with priority", func() {
+			ip := AccessEventConnInfo{
+				XRealIP:       "a",
+				XForwardedFor: "b",
+				Forwarded:     "for=c",
+				RemoteAddr:    "d",
+			}.IP()
+			So(ip, ShouldEqual, "a")
```
louischan-oursky: Just realized we have the priority

1. `x-real-ip`
2. `x-forwarded-for`
3. `forwarded`
4. remote addr

How did we come up with this priority? Any reference?

pkg/auth/dependency/auth/session_middleware.go
```diff
@@ -45,25 +49,48 @@ func (m *Middleware) resolve(rw http.ResponseWriter, r *http.Request) (AuthSessi
...
+
+func (m *Middleware) resolveSession(rw http.ResponseWriter, r *http.Request) (AuthSession, error) {
+	isInvalid := false
+
+	// IDP session in cookie takes priority over access token in header
+	for _, resolver := range []SessionResolver{m.IDPSessionResolver, m.AccessTokenSessionResolver} {
+		session, err := resolver.Resolve(rw, r)
+		if errors.Is(err, ErrInvalidSession) {
+			// Continue to attempt resolving session, even if one of the resolver reported invalid.
+			isInvalid = true
```
louischan-oursky: We never set it to `false` again. It is always true once it becomes true. So why don't we `return nil, ErrInvalidSession` here?

pkg/auth/dependency/auth/access_event_test.go
kiootic: It's arbitrary. My criteria is from non-standardized headers to standardized fields.

louischan-oursky: In my mind, they should be

1. `forwarded`
2. `x-forwarded-for`
3. `x-real-ip`
4. remote addr

From standardized headers to non-standardized headers.

kiootic: I don't have a strong opinion on this, so I'll change it to match yours.

PR #1278, opened 2020-03-24
Ensure cookie domain is ETLD+1 without port

### LINKED ISSUE #1275: Ensure cookie domain is always without port
In the function `NewSessionCookieConfiguration`, we use `publicsuffix.EffectiveTLDPlusOne(host)` to derive the domain. If `host` has `port`, `etldp1` also has port. But Cookie domain does not allow port.

`2020/03/24 18:20:47 net/http: invalid Cookie.Domain "accounts.localhost:8081"; dropping domain attribute`

We need to follow the similar logic in `isLocalhost` `csp_middleware.go` to strip the port.

louischan-oursky:
fix #1275

### DIFF
--- a/pkg/core/http/cookie_test.go
+++ b/pkg/core/http/cookie_test.go
@@ -103,3 +103,32 @@ func TestUpdateCookie(t *testing.T) {
 		})
 	})
 }
+
+func TestCookieDomainFromETLDPlusOneWithoutPort(t *testing.T) {
+	Convey("CookieDomainFromETLDPlusOneWithoutPort", t, func() {
+		check := func(in string, out string) {
+			actual := CookieDomainFromETLDPlusOneWithoutPort(in)
+			So(out, ShouldEqual, actual)
+		}
+		check("localhost", "")
+
+		check("accounts.localhost", "accounts.localhost")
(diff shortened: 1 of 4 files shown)


pkg/core/http/cookie_test.go
```diff
@@ -103,3 +103,32 @@ func TestUpdateCookie(t *testing.T) {
...
 	})
 }
+
+func TestCookieDomainFromETLDPlusOneWithoutPort(t *testing.T) {
+	Convey("CookieDomainFromETLDPlusOneWithoutPort", t, func() {
+		check := func(in string, out string) {
+			actual := CookieDomainFromETLDPlusOneWithoutPort(in)
+			So(out, ShouldEqual, actual)
+		}
+		check("localhost", "")
```
kiootic: I think this is a bug? It should be localhost. If we cannot derive eTLD+1, we may want to return the domain directly, since `localhost` is a TLD.

louischan-oursky: It is an issue of golang publicsuffix :(
However, returning empty string is part of the contract. If eTLD+1 cannot be derived, returning empty string means use host-only cookie.

Issue #1280, opened 2020-03-25
OIDC - support end_session_endpoint


carmenlau:
https://openid.net/specs/openid-connect-session-1_0.html#RPLogout

PR #1283, opened 2020-03-25
Fix logout endpoint

kiootic:
ref #1280


pkg/auth/dependency/oidc/handler/handler_end_session.go
```diff
@@ -6,27 +6,48 @@ import (
...
+
+		http.Redirect(rw, r, logoutURI.String(), http.StatusFound)
+		return nil
+	}
+
 	redirectURI := req.PostLogoutRedirectURI()
-	if !h.validateRedirectURI(s, redirectURI) {
-		// Invalid/empty redirect URI, redirect to settings to prompt user
-		http.Redirect(rw, r, h.SettingsEndpoint.SettingsEndpointURI().String(), http.StatusFound)
+	valid, client := h.validateRedirectURI(s, redirectURI)
```
louischan-oursky: passing `s` has no effect because `s` at this point must be `nil`.

pkg/auth/dependency/oidc/handler/handler_end_session.go
```diff
@@ -41,29 +62,23 @@ func (h *EndSessionHandler) Handle(s auth.AuthSession, req protocol.EndSessionRe
...
-			return err
-		}
-	}
-
 	http.Redirect(rw, r, redirectURI, http.StatusFound)
 	return nil
 }
 
-func (h *EndSessionHandler) validateRedirectURI(s auth.AuthSession, redirectURI string) bool {
+func (h *EndSessionHandler) validateRedirectURI(s auth.AuthSession, redirectURI string) (valid bool, client config.OAuthClientConfiguration) {
```
louischan-oursky: `s` is always nil in this function

pkg/auth/dependency/oidc/handler/handler_end_session.go
kiootic: Fixed.

PR #1287, opened 2020-03-26
Refactor user info

kiootic:


pkg/core/auth/authz/policy/authinfo.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 func requireAuthenticated(r *http.Request) error {
-	user := authn.GetAuthInfo(r.Context())
+	user := authn.GetUser(r.Context())
 	session := authn.GetSession(r.Context())
 	if user == nil || session == nil {
```
carmenlau: Maybe we only need to check if there is user in the context? In `RequiredAuthenticatedMiddleware` for web app, we only check user.

PR #1290, opened 2020-03-26
Fix Auth API behavior with tokens

kiootic:


pkg/auth/dependency/oauth/handler/result_authz.go
```diff
@@ -34,7 +34,7 @@ type (
 func (a authorizationResultRedirect) WriteResponse(rw http.ResponseWriter, r *http.Request) {
 	query := a.RedirectURI.Query()
 	for k, v := range a.Response {
-		query.Set(k, v)
+		query.Add(k, v)
```
louischan-oursky: Not sure why this change is related to the new test. Also this code snippet appears more than once. It is better to move this logic to `pkg/core/url/add_map.go`.

```golang
func AddMap(u *url.URL, m map[string]string) *url.URL {
  q := u.Query()
  for k, v := range m {
    q.Add(k, v)
  }
  out := *u
  out.RawQuery = q.Encode()
  return &out
}
```

kiootic: The expected behavior is that the existing query parameters should be preserved, so `Add` should be used instead of `Set`. The test is there to prevent regression in case we forget it.

PR #1293, opened 2020-03-27
Fix session APIs

kiootic:


pkg/auth/dependency/auth/session_manager.go
```diff
@@ -96,3 +111,51 @@ func (m *SessionManager) Logout(session AuthSession, rw http.ResponseWriter) err
...
+	if err != nil {
+		return nil, err
+	}
+
+	sessions := make([]AuthSession, len(idpSessions)+len(accessGrantSessions))
+	copy(sessions[0:], idpSessions)
+	copy(sessions[len(idpSessions):], accessGrantSessions)
+
+	sort.Slice(sessions, func(i, j int) bool {
+		return sessions[i].SessionID() < sessions[j].SessionID()
```
louischan-oursky: Sort by created at desc seems better.

pkg/auth/handler/session/revoke.go
```diff
@@ -89,98 +73,51 @@ const RevokeRequestSchema = `
...
-	if err := handler.BindJSONBody(r, w, h.Validator, "#SessionRevokeRequest", &payload); err != nil {
-		response.Error = err
-	} else {
-		result, err := h.Handle(r, payload)
-		if err != nil {
-			response.Error = err
-		} else {
-			response.Result = result
-		}
+	payload.CurrentSessionID = session.SessionID()
```
louischan-oursky: Is CurrentSessionID being respected?

kiootic: It is validated in `Validate()` function of the payload.

PR #1304, opened 2020-03-30
Disable auth API if not enabled

### LINKED ISSUE #1302: Disable Auth API if not explicitly enabled
Disallow using Auth API if `auth.enable_api` is not set to true.

kiootic:
fix #1302


kiootic:
- We do not want the restriction for existing admin API, and we do not want to introduce knowledge about individual handlers to middleware.
- This is a authorization restriction, so it is appropriate to be implemented as authorization policy.

PR #1312, opened 2020-04-01
Unlink not found oauth identity

### LINKED ISSUE #1308: Unlink non-existent SSO provider gives HTTP 500
The error got printed to stderr is `"principal not found"` which is should considered expected error instead of unexpected error.

louischan-oursky:
fix #1308

### DIFF
--- a/pkg/auth/dependency/principal/error.go
+++ b/pkg/auth/dependency/principal/error.go
@@ -10,5 +10,5 @@ var ErrNotFound = errors.New("principal not found")
 var ErrAlreadyExists = errors.New("principal already exists")
 var ErrMultipleResultsFound = errors.New("multiple principals found")
 
-var CurrentIdentityBeingDeleted = skyerr.Invalid.WithReason("CurrentIdentityBeingDeleted")
-var ErrCurrentIdentityBeingDeleted = CurrentIdentityBeingDeleted.New("must not delete current identity")
+// ErrCurrentIdentityBeingDeleted is shared by sso/unlink.go and loginid/remove.go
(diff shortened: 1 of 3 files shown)


pkg/auth/dependency/principal/oauth/error.go
```diff
@@ -0,0 +1,7 @@
+package oauth
+
+import (
+	"github.com/skygeario/skygear-server/pkg/core/skyerr"
+)
+
+var ErrIdentityNotFound = skyerr.NotFound.WithReason("OAuthIdentityNotFound").New("oauth identity not found")
```
kiootic, line 7: I think this error can be moved up to `principal` package, or even update the existing principal not found error, since there is no need for client to know the type of identity.

louischan-oursky: I updated the principal not found error to skyerr. However, we still have some error in `principal/password/error.go` that may overlap with those in `principal/error.go`

kiootic: Oh I forgot about the error in password package. If so, I think these skyerrs should belongs to the handler layer instead. So providers would return a simple error (`ErrNotFound`), and handlers would translate it into skyerrs for each group of APIs.

louischan-oursky: So revert to my original change?

kiootic: Yes, sorry for this. It would be better to also move the not found errors into the handler (i.e. into `loginid` and `sso` packages in handlers)

louischan-oursky: Updated.

PR #1318, opened 2020-04-03
Adopt Post/Redirect/Get pattern

### LINKED ISSUE #1288: [Auth UI] Adopt Post/Redirect/Get pattern
> I just found another potential problem: when I submit a wrong password multiple time when logging in, using the back button in page would result in CACHE_MISS Chrome error.

louischan-oursky:
fix #1288

@carmenlau I refactored `authenticate_provider_impl.go` greatly and it may have impact on how you implement SSO.


cmd/auth/main.go
```diff
@@ -242,9 +242,14 @@ func main() {
 	webappRouter.Use(auth.MakeMiddleware(authDependency, auth.NewCSRFMiddleware))
 	webappRouter.Use(webapp.PostNoCacheMiddleware)
 
-	webappAnonymousRouter := webappRouter.NewRoute().Subrouter()
-	webappAnonymousRouter.Use(webapp.RequiredAnonymousMiddleware{}.Handle)
-	webapphandler.AttachRootHandler(webappAnonymousRouter, authDependency)
+	webappAuthRouter := webappRouter.NewRoute().Subrouter()
+	webappAuthRouter.Use(webapp.RequiredAnonymousMiddleware{}.Handle)
```
carmenlau, line 246: Seems not related to this PR, but should we change this middleware name? 🙈

louischan-oursky: What about `RequiredNotAuthenticatedMiddleware`?

PR #1324, opened 2020-04-07
Refactor configuration

### LINKED ISSUE #1319: Refactor app configuration

kiootic:
fix #1319


pkg/core/config/message.go
```diff
@@ -0,0 +1,64 @@
+package config
+
+//go:generate msgp -tests=false
+
+type MessagesConfiguration struct {
+	Email       EmailMessageConfiguration `json:"email,omitempty" yaml:"email" msg:"email"`
+	SMSProvider SMSProvider               `json:"sms_provider,omitempty" yaml:"sms_provider,omitempty" msg:"sms_provider"`
+	SMS         SMSMessageConfiguration   `json:"sms,omitempty" yaml:"sms" msg:"sms"`
```
louischan-oursky, line 8: It is a bit strange we have both `sms_provider` and `sms` at the same level. I understand we need them separate because SMSMessageConfiguration could appear in feature configuration.

pkg/core/config/tenant.go
```diff
@@ -370,11 +413,11 @@ func (c *TenantConfiguration) AfterUnmarshal() {
...
 			config.Expiry = 3600 // 1 hour
 		}
-		if config.Sender == "" {
-			config.Sender = "no-reply@skygear.io"
-		}
-		if config.Subject == "" {
-			config.Subject = "Verification instruction"
+		if config.EmailMessage.Subject() == "" {
+			if config.EmailMessage == nil {
+				config.EmailMessage = EmailMessageConfiguration{}
```
carmenlau, line 418: Maybe we can use `default_zero_value:"true"` tag, so we don't need to check and assign zero value?

pkg/core/config/authentication.go
```diff
@@ -0,0 +1,18 @@
...
+	PrimaryAuthenticators       []string                    `json:"primary_authenticators,omitempty" yaml:"primary_authenticators" msg:"primary_authenticators"`
+	SecondaryAuthenticators     []string                    `json:"secondary_authenticators,omitempty" yaml:"secondary_authenticators" msg:"secondary_authenticators"`
+	SecondaryAuthenticationMode SecondaryAuthenticationMode `json:"secondary_authentication_mode,omitempty" yaml:"secondary_authentication_mode" msg:"secondary_authentication_mode"`
+	Secret                      string                      `json:"secret,omitempty" yaml:"secret" msg:"secret"`
+}
+
+type SecondaryAuthenticationMode string
+
+const (
+	SecondaryAuthenticationModeIfRequested SecondaryAuthenticationMode = "if_requested"
```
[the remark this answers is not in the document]
carmenlau: The original `off` is we don't ask user in the authentication flow, but user can still do mfa if they want?
How about call it `not_required`? 🙈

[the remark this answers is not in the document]
louischan-oursky: Not sure if the user can trigger MFA if they do not have authentication session token.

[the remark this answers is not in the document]
carmenlau: > The original `off` is we don't ask user in the authentication flow, but user can still do mfa if they want

But i think it is the original design?

> Not sure if the user can trigger MFA if they do not have authentication session token.

As I remember mfa api accept **authentication session token** or **valid session**?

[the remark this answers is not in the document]
kiootic: `if_requested` is same as `off`. I've updated the spec document to reflect this.

I think we're going to support step-up natively within authentication session. Therefore, it is possible to have an authentication session backed by an existing session. Existing MFA APIs would either accept an authentication session, or create a new authentication session with existing session, step it up, then finish the authentication session and update the session attributes.

pkg/core/config/tenant.go
kiootic: So `default_zero_value` works with maps too? I didn't know that.

pkg/core/config/authentication.go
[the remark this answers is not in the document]
louischan-oursky: By the way, ss `voluntary` a better name?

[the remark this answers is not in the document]
kiootic: If `voluntary` is used, I think `required` should be replaced with `mandatory` for contrast? Also, end-users cannot 'voluntarily' initiate a secondary authentication; it has to be 'requested' by the developer at authorize endpoint.

pkg/core/config/message.go
kiootic: I do think it looks weird... but I can't think of a better name. I had an idea of moving the message configuration one level deeper, so it would be `messages.email.message`, `messages.sms.message`, and `messages.sms.provider`, but it seems even weirder.

Any suggestions?

pkg/core/config/authentication.go
```diff
@@ -4,7 +4,7 @@ package config
 type AuthenticationConfiguration struct {
 	Identities                  []string                    `json:"identities,omitempty" yaml:"identities" msg:"identities"`
 	PrimaryAuthenticators       []string                    `json:"primary_authenticators,omitempty" yaml:"primary_authenticators" msg:"primary_authenticators"`
-	SecondaryAuthenticators     []string                    `json:"secondary_authenticators,omitempty" yaml:"secondary_authenticators" msg:"secondary_authenticators"`
+	SecondaryAuthenticators     *[]string                   `json:"secondary_authenticators,omitempty" yaml:"secondary_authenticators" msg:"secondary_authenticators"`
```
louischan-oursky, line 7: We have 2 nil cases here.

The pointer to `[]string` is nil.
The pointer to `[]string` is non-nil but `[]string` is nil.

Do we really need to change the type here? For example `LoginIDKeys` is still `[]LoginIDKeyConfiguration` but we explicitly check it against nil to see if it is specified by the developer.

pkg/core/config/message.go
louischan-oursky: No good idea :(

Just keep this.

pkg/core/config/authentication.go
kiootic: I'll change it back.

Go's `nil` handling is really confusing... It is convention to treat `nil` slice and empty slice as same, yet encoding tends to treat them differently.

louischan-oursky: golang-encoding impedance mismatch

PR #1326, opened 2020-04-08
Support localization in Auth UI

### LINKED ISSUE #1323: [Auth UI] Support localization in templates
- [x] Add `Translation config.TemplateItemType` to `template.Spec`. It expresses the template depends on another JSON file to provide the translation
- [x] Resolve all translations when a template has `Translation`
- [x] Support injecting custom function during render via `template.Funcs`.
- [x] Inject `localize` function when rendering Auth UI templates
- [x] Implement `localize` function
- [x] Extract all localizable texts from Auth UI templates.
- [x] Support `ui_locales` and `Accept-Language` (in that order)

louischan-oursky:
fix #1323


pkg/core/template/engine.go
```diff
@@ -1,98 +1,162 @@
...
+	// For example,
+	// {
+	//   "key1": {
+	//     "": "Hello",
+	//     "en": "Hello",
+	//     "en-US": "Hi!",
+	//     "zh": "你好"
+	//   }
+	// }
+	Translations map[string]map[string]string
```
kiootic, line 31: Would it be better to flatten it like this?
```json
{
    "key1": "Hello",
    "key1#en": "Hello",
    "key1#en-US": "Hi!",
    "key1#zh":  "你好"
}
```

In this ways, we can also resolve the other form of localization in configuration.

pkg/core/template/engine_test.go
```diff
@@ -191,3 +191,174 @@ func TestEngine(t *testing.T) {
...
+		}, []string{"ja-JP", "en-US", "zh-Hant-HK"}, "Hello from ja")
+
+		// Select the default because japanese and chinese is unknown to a korean speaker.
+		test(map[string]string{
+			"":   "Hello from default",
+			"ja": "Hello from ja",
+			"zh": "Hello from zh",
+		}, []string{"kr-KR"}, "Hello from default")
+
+		// Select english because english is an international language.
```
kiootic, line 325: Can you explain more on these three test cases? The term 'international language' and 'best variant' is not familiar to me, and looks rather 'magical'.

louischan-oursky: I don't know either. The comment given there is my guess. [This blog](https://blog.golang.org/matchlang) also just gives some examples. The blogpost says that [the implementation](https://github.com/golang/text/blob/master/language/match.go) takes CLDR data into account. So it is magical because it is driven by real world data.

For example, I guess there are more Taiwanese than Hongkongers. So `zh-Hant-TW` is preferred over `zh-Hant-HK`

kiootic: I suppose so... I just seems unintuitive that a missing language would resolve to English localization, rather than default localization.

I think we can omit the three test cases, since they're actually testing the underlying CLDR data, and I'm not sure if language/system upgrade would change the behavior. We can just test the most expected cases (e.g. `ja-JP` falls back to `ja`, `zh-Hant` would not falls back to `ja`, etc.).

pkg/core/template/engine.go
louischan-oursky: First it would impose a new restriction on the key naming. Key cannot have `#` sign otherwise it would be problematic.

Second the two-map approach does not require extra key manipulation to index. It is simpler.

Third the OIDC style localization is not yet done right now. I am going to support it in another PR (because your configuration refactoring introduces some configs that are not client metadata but use OIDC style localization, so I didn't do OIDC style localization in this PR). I am still thinking whether we should reuse MessageFormat for OIDC style localization. MessageFormat is powerful when we give it arguments to deal with plural and gender.

pkg/core/template/engine_test.go
louischan-oursky: Agree. Let me remove the test case related to the behavior of Matcher.

pkg/core/template/engine.go
kiootic: > I am still thinking whether we should reuse MessageFormat for OIDC style localization. MessageFormat is powerful when we give it arguments to deal with plural and gender.

I don't think we need to use it, since localization in configuration should be static strings without arguments, and I don't the OAuth/OIDC spec allows it.

louischan-oursky: > I don't think we need to use it, since localization in configuration should be static strings without arguments, and I don't the OAuth/OIDC spec allows it.

And that's why I don't think flattening the map here is helpful. The two localization processes are different. The data structure they use can also be different.

louischan-oursky:
conflicts resolved.

PR #1330, opened 2020-04-09
Configure country calling code

### LINKED ISSUE #1321: [Auth UI] Configurable country calling codes

louischan-oursky:
fix #1321


pkg/core/config/schema.go
```diff
@@ -166,7 +167,26 @@ var (
...
 		"properties": {
-			"css": { "type": "string" }
+			"css": { "type": "string" },
+			"country_calling_code": { "$ref": "#AuthUICountryCallingCodeConfiguration" }
+		}
+	},
+	"CountryCallingCode": {
+		"$id": "#CountryCallingCode",
+		"type": "string",
+		"enum": %s
```
kiootic, line 177: I think it's better not to embed all codes in the schema, since it would bloat the schema greatly. Also, I'm not sure if the codes are exhaustive and will not be changed. Maybe use a numeric regex instead?

louischan-oursky: Writing a regex for it is not an good idea. Maybe make it a format instead?

kiootic: I'm thinking of just a simple regex like `^\d+$`, since it seems likely that our list of calling codes is not always correct in future.

kiootic:
@louischan-oursky there is conflict here.

louischan-oursky:
rebased.

PR #1332, opened 2020-04-14
Support OIDC style localization - Part 1

### LINKED ISSUE #1329: Support OIDC style localization in configuration
- [x] Support OIDC style localization in request
- [x] Retain preferred language in async task

louischan-oursky:
ref #1329 

TODO:

- Make AsyncTask aware of the locale of the request

I am going to handle async task in the next PR.


pkg/auth/dependency/webapp/render_provider_impl.go
```diff
@@ -54,8 +53,11 @@ func (p *RenderProviderImpl) WritePage(w http.ResponseWriter, r *http.Request, t
 		return MakeURLWithPath(r.URL, path)
 	}
 
-	data["client_name"] = accessKey.Client["client_name"]
-	data["logo_uri"] = accessKey.Client["logo_uri"]
+	preferredLanguageTags := intl.GetPreferredLanguageTags(r.Context())
+
+	clientMetadata := accessKey.Client
+	data["client_name"] = intl.LocalizeOIDCJSONObject(preferredLanguageTags, clientMetadata, "client_name")
```
kiootic, line 59: It seems currently there is no way to trigger this, as access key is read from header only?

pkg/core/intl/oidc.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	"golang.org/x/text/language"
+)
+
+// LocalizeOIDCJSONObject returns the localized value of key in jsonObject according to preferredLanguageTags.
+func LocalizeOIDCJSONObject(preferredLanguageTags []string, jsonObject map[string]interface{}, key string) string {
```
kiootic, line 12: I think it's better not to use 'OIDC' in the names, since we're not just using it for OIDC, but also other localized strings in configuration.

pkg/auth/dependency/webapp/render_provider_impl.go
louischan-oursky: Auth UI now understand `redirect_uri`, `client_id` and `ui_locales`.

PR #1340, opened 2020-04-16
Support component template

### LINKED ISSUE #1338: Support component template
- [ ] Add component template loading
- [ ] Convert some Defines of Auth UI to component templates

louischan-oursky:
fix #1338

### DIFF
--- a/pkg/auth/dependency/webapp/template.go
+++ b/pkg/auth/dependency/webapp/template.go
@@ -6,6 +6,10 @@ import (
 )
 
 const (
+	TemplateItemTypeAuthUIHTMLHeadHTML config.TemplateItemType = "auth_ui_html_head.html"
+	TemplateItemTypeAuthUIHeaderHTML   config.TemplateItemType = "auth_ui_header.html"
+	TemplateItemTypeAuthUIFooterHTML   config.TemplateItemType = "auth_ui_footer.html"
+
 	TemplateItemTypeAuthUILoginHTML config.TemplateItemType = "auth_ui_login.html"
 	// nolint: gosec
 	TemplateItemTypeAuthUILoginPasswordHTML config.TemplateItemType = "auth_ui_login_password.html"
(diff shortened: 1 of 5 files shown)


pkg/auth/dependency/webapp/template.go
```diff
@@ -16,8 +20,11 @@ const (
 	TemplateItemTypeAuthUILogoutHTML         config.TemplateItemType = "auth_ui_logout.html"
 )
 
-const defineHead = `
-{{ define "HEAD" }}
+var TemplateAuthUIHTMLHeadHTML = template.Spec{
+	Type:   TemplateItemTypeAuthUIHTMLHeadHTML,
+	IsHTML: true,
+	Default: `
+{{ define "auth_ui_html_head.html" }}
```
kiootic, line 27: I think it's better to have a fallback for client_name, in case user entered the account page directly? We can do this in another issue.

pkg/core/template/engine.go
```diff
@@ -259,6 +278,44 @@ func (e *Engine) resolveTranslations(templateType config.TemplateItemType) (tran
 	return
 }
 
+func (e *Engine) resolveComponents(types []config.TemplateItemType, key string) (bodies []string, err error) {
+	return e.resolveComponents0(types, key, make(map[config.TemplateItemType]struct{}))
+}
+
+func (e *Engine) resolveComponents0(types []config.TemplateItemType, key string, seen map[config.TemplateItemType]struct{}) (bodies []string, err error) {
```
kiootic, line 285: I think this can be a local function in `resolveComponents` to simplify it?
e.g.
```go
func (e *Engine) resolveComponents(types []config.TemplateItemType, key string) ([]string, error) {
    templates := map[config.TemplateItemType]string
    resolve := func(types []config.TemplateItemType) error {
        // ...
    }
    err := resolve(types)
    if err != nil {
        return nil, err
    }
    var bodies []string
    for _, body := range templates {
        bodies = append(bodies, body)
    }
    return bodies, nil
}
```

pkg/auth/dependency/webapp/template.go
louischan-oursky: Actually the placement of `client_name` is tentative. It does not appear in the design. Let me open another issue to track this.

louischan-oursky:
Updated.

PR #1342, opened 2020-04-17
Auth UI Forgot Password - Part 1

louischan-oursky:
This PR rewrites the existing implementation

Changes:

- Able to send reset password link via SMS
- Remove SecureMatch
- Remove success and error redirect
- Remove the ad-hoc form


pkg/auth/dependency/forgotpassword/provider.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return
+	}
+
+	if len(prins) <= 0 {
+		err = ErrLoginIDNotFound
+		return
+	}
+
+	for _, prin := range prins {
```
carmenlau, line 69: After removing realm, is there any case that we need to handle multiple principals in forgot password?

louischan-oursky: Maybe `username` is configured to allow `@` and a username looks like an email address.
But the loop body does not handle this case correct. It should skip non-email and non-phone login ID.

pkg/auth/dependency/forgotpassword/provider.go
```diff
@@ -84,12 +73,16 @@ func (p *Provider) SendCode(loginID string) (err error) {
 
 		if email {
 			err = p.sendEmail(prin.LoginID, codeStr)
-			return
+			if err != nil {
+				return
+			}
```
carmenlau, line 78: I think we should return after sending even there is no error? I think creating and sending multiple code is not expected in forgot password context?

louischan-oursky:
Let me close this and continue in the PR of part 2.

Issue #1274, opened 2020-03-24
Auth UI Web App - Forgot Password


louischan-oursky:
- [x] Refactor forgot password so that Auth API and Auth UI can share the same implementation
- [x] Implement forgot password for Auth UI
- [ ] Update email template to match new design
- [ ] Use async task to send message

PR #1348, opened 2020-04-21
Auth UI Forgot Password - Part 2

louischan-oursky:
ref #1274

TODO:

Update the email template to match new design.
Use task to send message.


louischan-oursky:
> For UX, since forgot password page is in the third page. 

> put forgot password entry point in the first page?

Forgot password page is reachable from the following pages: login page, login password page, signup page. I think login page is the first page?

> maybe we will want to pre-fill value with user inputted login id?

Agree that if the user goes to forgot password page from login password page, the login ID is best pre-filled. However, it requires a form submission to achieve this. I would like to keep things simpler.

PR #1333, opened 2020-04-15
Add Interaction

kiootic:

PR #1352, opened 2020-04-21
Integrate interaction and login auth API

kiootic:
requires #1333

### DIFF
--- a/pkg/auth/dependency/interaction/provider.go
+++ b/pkg/auth/dependency/interaction/provider.go
@@ -1,16 +1,44 @@
 package interaction
 
+import (
+	gotime "time"
+
+	"github.com/skygeario/skygear-server/pkg/core/time"
+)
+
+// TODO(interaction): configurable lifetime
+const interactionIdleTimeout = 5 * gotime.Minute
+
 type Provider struct {
+	Store Store
+	Time  time.Provider
 }
 
 func (p *Provider) GetInteraction(token string) (*Interaction, error) {
-	// TODO(interaction): do something
-	return nil, nil
+	i, err := p.Store.Get(token)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(interaction): populate identity & authenticator infos
+	return i, nil
 }
 
 func (p *Provider) SaveInteraction(i *Interaction) (string, error) {
-	// TODO(interaction): do something
-	return "", nil
+	if i.Token == "" {
+		i.Token = generateToken()
+		i.CreatedAt = p.Time.NowUTC()
+		i.ExpireAt = i.CreatedAt.Add(interactionIdleTimeout)
+		if err := p.Store.Create(i); err != nil {
+			return "", err
+		}
+	} else {
+		i.ExpireAt = p.Time.NowUTC().Add(interactionIdleTimeout)
+		if err := p.Store.Update(i); err != nil {
+			return "", err
+		}
+	}
+	return i.Token, nil
 }
 
 func (p *Provider) Commit(i *Interaction) error {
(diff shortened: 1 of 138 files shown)


pkg/auth/model/session.go
```diff
@@ -31,14 +27,15 @@ const SessionSchema = `
 	"type": "object",
 	"properties": {
 		"id": { "type": "string" },
-		"identity_id": { "type": "string" },
+		"identity_type": { "type": "string" },
+		"identity_claims": { "type": "object" },
+		"acr": { "type": "string" },
+		"amr": { "type": "string" },
```
louischan-oursky, line 33: It should be array of string.

pkg/auth/dependency/interaction/provider.go
```diff
@@ -0,0 +1,95 @@
...
+	Authenticator AuthenticatorProvider
+	Config        *config.AuthenticationConfiguration
+}
+
+func (p *Provider) GetInteraction(token string) (*Interaction, error) {
+	i, err := p.Store.Get(token)
+	if err != nil {
+		return nil, err
+	}
+
```
louischan-oursky, line 40: Why do we need to fetch Identity/PrimaryAuthenticator/SecondaryAuthenticator?

pkg/auth/dependency/interaction/provider_intent.go
```diff
@@ -0,0 +1,23 @@
+package interaction
+
+import (
+	"fmt"
+
+	"github.com/skygeario/skygear-server/pkg/auth/dependency/auth"
+)
+
+func (p *Provider) NewInteraction(intent Intent, clientID string, session auth.AuthSession) (*Interaction, error) {
```
louischan-oursky, line 9: IMO it is simpler to have separate functions to create Interaction with a concrete Intent.

```golang
func InteractionWithIntentLogin(intent *IntentLogin, clientID string) *Interaction {
  return &Interaction{
    Intent: intent,
    ClientID: clientID,
  }
}
```

pkg/auth/dependency/interaction/provider.go
kiootic: We should be using the most up-to-date information of authenticator, in case of removal, password changes, etc.

pkg/auth/dependency/interaction/provider_action.go
```diff
@@ -0,0 +1,104 @@
...
+	} else if err != nil {
+		return nil, err
+	}
+
+	authen, err := p.Authenticator.Authenticate(userID, as, astate, secret)
+	if err != nil {
+		return nil, err
+	}
+
+	if step.Step == StepAuthenticateSecondary {
```
carmenlau, line 87: Why do we need to check if `step.Step == StepAuthenticateSecondary` only?

pkg/auth/dependency/webapp/template.go
```diff
@@ -260,6 +260,7 @@ var TemplateAuthUILoginPasswordHTML = template.Spec{
 <input type="hidden" name="x_calling_code" value="{{ .x_calling_code }}">
 <input type="hidden" name="x_national_number" value="{{ .x_national_number }}">
 <input type="hidden" name="x_login_id" value="{{ .x_login_id }}">
+<input type="hidden" name="x_interaction_token" value="{{ .x_interaction_token }}">
```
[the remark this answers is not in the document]
kiootic: Yes, that's the intention.

pkg/auth/dependency/interaction/provider_action.go
kiootic: `doAuthenticate` is used for `StepAuthenticatePrimary` or `StepAuthenticateSecondary`. For `StepAuthenticatePrimary`, we would not provide user with concrete authenticators, but available authenticators as configured, so there is no need to check for `StepAuthenticatePrimary`.

I'll add a comment explaining this later.

kiootic:
Merged to interaction branch.

PR #1355, opened 2020-04-22
Auth UI Forgot Password

### LINKED ISSUE #1274: Auth UI Web App - Forgot Password

louischan-oursky:
fix #1274


louischan-oursky:
Travis keeps complaining `make generate` produces uncommited changes. But I cannot produce it locally.

louischan-oursky:
> Should we remove the signin link in reset password successful page?

But if we remove the `/login` button, the reset password success page is a dead end.

carmenlau:
After offline discussion, the dead end in reset password is expected. Since user open the reset password link from the email, and auth ui doesn't have redirect uri for redirecting back to the app. So user should go back to the app themselves after reseting password.

PR #1359, opened 2020-04-24
Allow theming auth ui

louischan-oursky:


louischan-oursky:
@carmenlau Can you review this when you have time? I would like to have this fix so the demo can have a better theme experience.

PR #1361, opened 2020-04-24
Use interaction for signup

kiootic:


pkg/auth/dependency/interaction/provider_commit.go
```diff
@@ -0,0 +1,76 @@
...
+		UserID:         i.UserID,
+		IdentityType:   i.Identity.Type,
+		IdentityClaims: i.Identity.Claims,
+		// TODO(interaction): populate acr & amr
+	}
+	return attrs, nil
+}
+
+func (p *Provider) onCommitSignup(i *Interaction, intent *IntentSignup) error {
+	if intent.OnUserDuplicate == model.OnUserDuplicateAbort {
```
louischan-oursky, line 44: Add a TODO here to remember to handle Merge and CreateNew

pkg/auth/dependency/interaction/adaptors/identity.go
```diff
@@ -72,6 +73,24 @@ func (a *IdentityAdaptor) GetByClaims(typ authn.IdentityType, claims map[string]
...
+		}
+
+		userID, i, err = a.GetByClaims(authn.IdentityTypeOAuth, claims)
+		if err != nil && !errors.Is(err, identity.ErrIdentityNotFound) {
+			return "", nil, err
+		} else if err == nil {
+			return userID, i, nil
+		}
+
+		return "", nil, identity.ErrIdentityNotFound
```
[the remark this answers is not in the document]
kiootic: Since it is used for checking for duplicated identity, I think instead I'll add a `ExistsByClaims` function that returns a boolean indicating whether an identity with the claims exists.

pkg/auth/dependency/interaction/provider_commit.go
kiootic: CreateNew should not need special handling here.
We do not support Merge for signing up, it should be used by OAuth only. It should:
1. Get OAuth data when returning from provider.
2. Create new login interaction.
3. If failed to login (i.e. new user/need merge), create new signup interaction.
4. If failed to sign up (i.e. need merge), create new add identity interaction.

For Auth API, the three interaction flow would be performed automatically.
For Web UI, the final merge (add identity) interaction would be performed only after some kind of confirmation/authentication screen.

pkg/auth/dependency/interaction/provider_commit.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+
+			if _, exists := emailIdentities[email]; exists {
+				return ErrDuplicatedIdentity
+			}
+			emailIdentities[email] = struct{}{}
+		}
+
+		for email := range emailIdentities {
+			_, _, err := p.Identity.GetByClaims("", map[string]interface{}{string(metadata.Email): email})
```
carmenlau, line 59: If I understand correctly, the `GetByClaims` should accept new format of identity claims? Or `GetByClaims` will be updated to accpet both 

```
{
	"email": "user1@example.com"
}
```
or
```
{
	"https://auth.skygear.io/claims/login_id/key":   "email",
	"https://auth.skygear.io/claims/login_id/value": "user1@example.com"
}
```
?

pkg/auth/handler/signup.go
```diff
@@ -51,7 +50,8 @@ const SignupRequestSchema = `
 				},
 				"required": ["key", "value"]
 			},
-			"minItems": 1
+			"minItems": 1,
+			"maxItems": 1
```
louischan-oursky, line 54: The public API is changed.

pkg/auth/dependency/interaction/flows/webapp.go
```diff
@@ -63,11 +98,71 @@ func (f *WebAppFlow) AuthenticatePassword(token string, password string) (*WebAp
...
 	if err != nil {
 		return nil, err
 	}
 
+	switch i.Intent.Type() {
+	case interaction.IntentTypeSignup:
+		// New interaction for logging in after signup
+		i, err = f.Interactions.NewInteractionLogin(&interaction.IntentLogin{
+			Identity: i.Identity.ToSpec(),
+			AuthenticatedAs: &interaction.IntentLoginAuthenticatedAs{
```
carmenlau, line 144: I think having `AuthenticatedAs` and do the login again after signup is a good move! But seems the reason of session in `CreateSession` will become always `login`? We use the reason for hook, right?

pkg/auth/dependency/interaction/provider.go
```diff
@@ -12,14 +12,24 @@ import (
 type IdentityProvider interface {
 	Get(userID string, typ authn.IdentityType, id string) (*IdentityInfo, error)
 	GetByClaims(typ authn.IdentityType, claims map[string]interface{}) (string, *IdentityInfo, error)
+	ListByClaims(claims map[string]string) ([]*IdentityInfo, error)
```
carmenlau, line 15: `GetByClaims ` and `ListByClaims` treat the claims differently, better to add some comments here?

migrations/auth/1587982019_cleanup_columns.up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE _auth_authenticator_oob DROP COLUMN activated;
+ALTER TABLE _auth_authenticator_oob DROP COLUMN activated_at;
+DROP TABLE _auth_authenticator_oob_code;
+ALTER TABLE _auth_identity_login_id DROP COLUMN password;
+ALTER TABLE _auth_identity_login_id DROP COLUMN realm;
+ALTER TABLE _auth_identity_oauth DROP COLUMN token_response;
```
carmenlau, line 6: Since we updated some of the table schema here and the old code is not updated, so the api are broken. We need to unbreak it before we merge interaction back to master, I am ok for this. Just make sure we aware of it. cc @louischan-oursky 

And I found we drop `token_response` in oauth table here, seems we are not using it actually. Would like to confirm we don't need it?

pkg/auth/dependency/interaction/flows/webapp.go
kiootic: Now the original intent type is stored to determine session create reason.

migrations/auth/1587982019_cleanup_columns.up.sql
louischan-oursky: I think the token response is intended for supporting refreshing the User Info in the future if there is a refresh token in it.

louischan-oursky: If we want to remove it, I prefer we remove all related code. I saw in Carmen's PR we simply do not save it in the store implementation to unbreak things. Therefore keeping it instead of removing it would minimize the size of code changes.

kiootic: Okay. I'll keep it and set it to an empty object for the new identity provider.

kiootic:
@louischan-oursky Did you ran the new migration previously, and tried to run the fixed new migration? I just fix-up the commit that added the migration to avoid having broken migration in history.

louischan-oursky:
I `docker-compose down -v` everything and started from scratch.

carmenlau:
The PR looks good to me, but when I test with the JS SDK it fails to decode the identity as we changed the identity type to `login_id`. I remember we had discussed on it, but seems not only the gateway headers, sdk also depends on the identity type. Maybe we cannot change the identity type?:(

https://github.com/SkygearIO/skygear-SDK-JS/blob/e78b192dee8af7645b7b16927314bfda3f3f8530/packages/skygear-core/src/encoding.ts#L226

louischan-oursky:
We can change it if we introduce a new field such as `kind` to replace `type`. New SDK should read `kind` instead. The server writes both.

kiootic:
Having both `kind` and `type` with different value is really hacky IMO. I think as we discussed before we can just force all internal users to upgrade to fix it?

louischan-oursky:
SDK is bundled with the mobile application and cannot be force update :(

carmenlau:
Yes, per our discussion we can force internal users to update server code, but SDK we cannot :(

pkg/auth/dependency/interaction/infos.go
```diff
@@ -10,7 +10,11 @@ type IdentityInfo struct {
 }
 
 func (i *IdentityInfo) ToSpec() IdentitySpec {
-	return IdentitySpec{ID: i.ID, Type: i.Type, Claims: i.Claims}
+	return IdentitySpec{Type: i.Type, Claims: i.Claims}
+}
+
+func (i *IdentityInfo) ToRef() IdentityRef {
+	return IdentityRef{ID: i.ID}
```
carmenlau, line 17: Missing type when converting to ref

pkg/auth/dependency/interaction/infos.go
```diff
@@ -34,26 +38,20 @@ const (
...
-
-const (
-	AuthenticatorTypePassword     AuthenticatorType = "password"
-	AuthenticatorTypeTOTP         AuthenticatorType = "totp"
-	AuthenticatorTypeOOBOTP       AuthenticatorType = "oob_otp"
-	AuthenticatorTypeBearerToken  AuthenticatorType = "bearer_token"
-	AuthenticatorTypeRecoveryCode AuthenticatorType = "recovery_code"
-)
+func (i *AuthenticatorInfo) ToRef() AuthenticatorRef {
+	return AuthenticatorRef{ID: i.ID}
```
carmenlau, line 53: Missing type when converting to ref

PR #1367, opened 2020-04-28
OOB OTP

louischan-oursky:


pkg/auth/dependency/interaction/provider_action.go
```diff
@@ -103,5 +139,84 @@ func (p *Provider) doAuthenticate(i *Interaction, step *StepState, astate *map[s
...
+	if err != nil {
+		return
+	}
+
+	if i.State == nil {
+		i.State = map[string]string{}
+	}
+
+	// Check if we have a code already.
+	// The code remains unchanged through out the entire interaction once it was generated.
```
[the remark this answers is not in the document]
louischan-oursky: Common behavior I observe is that the code changes at certain interval. If the code changes on each resend. The user may receive the previous message first but the code there is no longer valid.

pkg/auth/dependency/interaction/flows/webapp.go
```diff
@@ -31,13 +31,25 @@ func (f *WebAppFlow) LoginWithLoginID(loginID string) (*WebAppResult, error) {
 		panic("interaction_flow_webapp: unexpected interaction state")
 	}
 
+	step := WebAppStepAuthenticatePassword
+	// Trigger OOB OTP if the first authenticator is OOB OTP
+	if len(s.Steps[0].AvailableAuthenticators) > 0 && s.Steps[0].AvailableAuthenticators[0].Type == interaction.AuthenticatorTypeOOBOTP {
```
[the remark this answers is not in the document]
louischan-oursky: Identity does not **have** primary authenticators. In the PR I define

- An authenticator is compatible with an identity
- An authenticator is related to an identity

in the RelateIdentityToAuthenticator method.

Authentication can have at most 1 primary authenticator and at most 1 secondary authenticator associated with it.

A user can have many authenticators. In database, we do not classify them as primary or secondary. Their classification is done in application level.

pkg/auth/dependency/interaction/oob.go
```diff
@@ -0,0 +1,123 @@
...
+	taskspec "github.com/skygeario/skygear-server/pkg/auth/task/spec"
+	"github.com/skygeario/skygear-server/pkg/core/async"
+	"github.com/skygeario/skygear-server/pkg/core/authn"
+	"github.com/skygeario/skygear-server/pkg/core/config"
+	"github.com/skygeario/skygear-server/pkg/core/mail"
+	"github.com/skygeario/skygear-server/pkg/core/sms"
+	"github.com/skygeario/skygear-server/pkg/core/template"
+)
+
+type OOBProviderImpl struct {
```
kiootic, line 15: I suppose this provider should belong to `authenticator/oob` package, and can accept `*oob.Authenticator` instead?

scripts/html-email/templates/oob_code_email.mjml
```diff
@@ -0,0 +1,26 @@
+<mjml>
```
kiootic, line 1: Any instructions on how to transpile this?

pkg/auth/dependency/interaction/adaptors/authenticator.go
```diff
@@ -157,6 +157,52 @@ func (a *AuthenticatorAdaptor) List(userID string, typ interaction.Authenticator
...
+		}
+
+		loginID := ii.Claims[interaction.IdentityClaimLoginIDValue]
+		var oas []*oob.Authenticator
+		oas, err = a.OOBOTP.List(userID)
+		if err != nil {
+			return
+		}
+		for _, oa := range oas {
+			if oa.Email == loginID || oa.Phone == loginID {
```
kiootic, line 195: I suppose we don't need to check the type of login ID key.

pkg/auth/dependency/interaction/provider.go
```diff
@@ -18,6 +18,21 @@ type IdentityProvider interface {
...
+	// OOB OTP authenticator requires email or phone.
+	//
+	// If authenticatorSpec is incompatible with identitySpec, nil is returned.
+	//
+	// Otherwise authenticatorSpec is further checked if it is related to identitySpec.
+	// If authenticatorSpec is related to identitySpec, authenticatorSpec.Props is mutated in-place.
+	// Currently on the following case mutation would occur.
+	//
+	//   - login ID identity of login ID type email or phone and OOB OTP authenticator.
+	RelateIdentityToAuthenticator(identitySpec IdentitySpec, authenticatorSpec *AuthenticatorSpec) *AuthenticatorSpec
```
kiootic, line 35: This function seems to have quite complex semantics. Can we refactor it to be simpler?

For example, it would be simpler to just take the identity and an authenticator type as parameters, and name it as `NewForIdentity`.

pkg/auth/dependency/interaction/provider_action.go
```diff
@@ -152,7 +156,19 @@ func (p *Provider) setupAuthenticator(i *Interaction, step *StepState, astate *m
 		return nil, ErrInvalidAction
 	}
 
-	// TODO(interaction): special handling for OTP
+	switch as.Type {
+	case AuthenticatorTypePassword:
+		// Nothing special needs to be done
+		break
+	case AuthenticatorTypeOOBOTP:
+		_, err := p.Authenticator.Authenticate( /* userID */ "", as, astate, secret)
```
kiootic, line 164: `i.UserID` should be passed in as userID.

pkg/auth/dependency/interaction/provider_action.go
```diff
@@ -157,3 +161,51 @@ func (p *Provider) setupAuthenticator(i *Interaction, step *StepState, astate *m
...
+
+	if spec.ID == "" {
+		panic("expected ActionTriggerOOBAuthenticator.Authenticator.ID to be present")
+	}
+
+	if spec.Type != AuthenticatorTypeOOBOTP {
+		panic("unexpected ActionTriggerOOBAuthenticator.Authenticator.Type: " + spec.Type)
+	}
+
+	now := p.Time.NowUTC()
```
kiootic, line 188: We should check the trigger cooldown here.

scripts/html-email/templates/oob_code_email.mjml
louischan-oursky: As a comment in the new make target.

pkg/auth/dependency/interaction/provider.go
louischan-oursky: This function is indeed complex because it is doing three things.

- Check compatible
- Check related
- If related, populate the authenticator spec

It must take an IdentitySpec because it needs the login ID to populate.
It can be refactored to take authenticator type instead because it does not read the claims. It writes only.

pkg/auth/dependency/interaction/provider_action.go
louischan-oursky: The user ID actually is unused because OOB OTP does not look at it. I will change it though.

pkg/auth/dependency/interaction/oob.go
louischan-oursky: We should accept AuthenticatorSpec because oob.Authenticator is generated after the code is verified.

pkg/auth/dependency/interaction/adaptors/authenticator.go
louischan-oursky: This if statement makes sure we only include OOB Authenticators that is related to the given identity.

louischan-oursky:
@carmenlau @kiootic 

resolved comments were addressed.

PR #1378, opened 2020-05-04
Default language and fallback language

### LINKED ISSUE #1363: default language and fallback language
- [ ] Define default language as `en`
- [ ] Introduce configurable fallback language, default to default language.

louischan-oursky:
fix #1363


pkg/core/intl/match.go
```diff
@@ -0,0 +1,56 @@
...
+	for i, tag := range supportedLanguageTags {
+		if tag == fallbackLanguageTag {
+			fallbackIdx = i
+			break
+		}
+	}
+
+	if fallbackIdx < 0 {
+		var s []string
+		s = append(s, fallbackLanguageTag)
```
carmenlau, line 22: I think `SortSupported` should only sort the list but not adding `fallbackLanguageTag`? If `fallbackLanguageTag` is not one of the supported language, and if the matched result return `fallbackLanguageTag`, render might have problem.

pkg/core/intl/oidc.go
```diff
@@ -18,66 +15,46 @@ func LocalizeJSONObject(preferredLanguageTags []string, jsonObject map[string]in
...
 
 // LocalizeStringMap returns the localized value of key in stringMap according to preferredLanguageTags.
-func LocalizeStringMap(preferredLanguageTags []string, stringMap map[string]string, key string) string {
+func LocalizeStringMap(preferredLanguageTags []string, fallbackLanguageTag string, stringMap map[string]string, key string) string {
 	prefix := fmt.Sprintf("%s#", key)
 	m := map[string]string{}
 	for k, stringValue := range stringMap {
 		if k == key {
-			m[""] = stringValue
+			m[fallbackLanguageTag] = stringValue
```
carmenlau, line 43: I found a bug that when I use the default config, this function will return empty string.

When `fallbackLanguageTag` is empty, the map `m` here store default value with empty key. `SortSupported` will populate default `en` to `fallbackLanguageTag` and the matched tag will be `en`.

pkg/core/intl/match.go
louischan-oursky: The property of `SortSupported` must uphold is that the returned slice is always non-empty and the first element must be `fallbackLanguageTag `. Maybe I should rename the function.

pkg/core/intl/oidc.go
louischan-oursky: I will add a if statement to set fallbackLanguageTag to DefaultLanguage if it is empty.

pkg/core/intl/match.go
carmenlau: But what if user provide `fallbackLanguageTag` not in the `supportedLanguageTags`?

Maybe what we should do is to ensure is default language is in `supportedLanguageTags` and `fallbackLanguageTag` default is default language?

louischan-oursky: > But what if user provide fallbackLanguageTag not in the supportedLanguageTags?

It is clearly a configuration problem introduced by the developer. Skygear only comes with english. If the developer change the `fallback_language` to something else, it is expected that they also have to provide ALL translation of that fallback language.

The purpose of `SortSupported` is to create a string slice that put the fallback language as the first element so that the matcher can use that language as fallback.

carmenlau: > It is clearly a configuration problem introduced by the developer. Skygear only comes with english. If the developer change the fallback_language to something else, it is expected that they also have to provide ALL translation of that fallback language.

ok, then keep it as it.

louischan-oursky:
@carmenlau 

Added FallbackLanguage and SupportedLanguages type alias to ensure better compile time checking of the intended invariants.

Fixed a one place still not using the new intl.Match

PR #1379, opened 2020-05-05
Do not mix SaveInteraction and Commit within a request

louischan-oursky:


pkg/auth/dependency/interaction/provider.go
```diff
@@ -60,6 +60,20 @@ type OOBProvider interface {
 // TODO(interaction): configurable lifetime
 const interactionIdleTimeout = 5 * gotime.Minute
 
+// NOTE(interaction): SaveInteraction and Commit are mutually exclusively within a request.
```
kiootic, line 63: I suppose it would be useful to add checks into `SaveInteraction` and `Commit`, since it would be hard to notice the misuse just by reading the code.

louischan-oursky: Added a runtime panic to guard this.

Issue #1375, opened 2020-05-04
Account settings Auth UI page


kiootic:
- [x] `/settings/identities`:
  - [x] Show a list of candidate identities
  - [x] Fetch the identities of the user and associate them with the candidates
  - [x] Show correct button according to the association
    - [x] Connect SSO
    - [x] Disconnect SSO
    - [x] Add login ID `/settings/identities/login_ids`
    - [x] Change login ID `/settings/identities/login_ids`

PR #1404, opened 2020-05-13
Identity settings - Part 1

louischan-oursky:
ref #1375

How to test:

1. Login with whatever method you want
2. Go to `/settings/identity` manually because `/settings` does not have a link to that page yet.

TODO:

Add/Update/Remove login ID with password/OOB OTP authenticator setup


louischan-oursky:
> Not sure if it is good to allow declaration for the template? Or better to declare all the variable e.g. has_oauth. To keep the template simple and restrict developer from writing complicated template. Open to discuss.

The RenderProvider is getting more and more long because it has to derive many data. So I put it in template. 

> Should translate error cannot remove last identity.

I am going to deal with error translation in another issue. Ideally we should extract all related errors (identified by Reason) and provide a translation for it. So the developer can customize the error message if they want.

> For the user facing path, now is /settings/identity. should we use /settings/accounts per our discussion?

We never use the term "account" in Skygear. To me, account is user. I would not say adding a email is adding an account.

Issue #1376, opened 2020-05-04
Anonymous user promotion sign up page


kiootic:
App may use OIDC flow to sign in as anonymous user, then redirect to a special promotion page (reusing signup page) for promoting user to normal user.

Promotion page have same UI as signup page, it allows adding one identity and removing existing anonymous identity. Promotion page can be used by anonymous users (i.e. users with anonymous identity) only and settings page cannot be used by normal users (i.e. users without anonymous identity).

After promotion, the page would redirect back to a URI specified in parameter:
```
/oidc/authorize?login_hint=foo&redirect_uri=app://on_promoted
/oidc/authorize -> check login hint -> redirect to /auth/promotion?interaction_token=foo&redirect_uri=app://on_promoted

/settings -> redirect to /auth/promotion?interaction_token=foo -> replace current session with new session -> redirect to /settings
/auth/promotion?interaction_token=foo&redirect_uri=app://on_promoted -> redirect to app://on_promoted?code=xxx -> SDK use OAuth API to get new session
```

- [x] Support anonymous promotion in server
- [x] Support anonymous promotion in React Native SDK

Signup pages should have SSO buttons. For promotion, the interaction should be created with a session ID.

PR #1405, opened 2020-05-13
Implement anonymous user promotion

kiootic:
requires #1407 
ref #1376

### DIFF
--- a/pkg/auth/dependency/interaction/flows/anonymous.go
+++ b/pkg/auth/dependency/interaction/flows/anonymous.go
@@ -104,7 +104,7 @@ func (f *AnonymousFlow) Authenticate(requestJWT string, clientID string) (*authn
 
 func (f *AnonymousFlow) DecodeUserID(requestJWT string) (string, anonymous.RequestAction, error) {
 	identity, request, err := f.Anonymous.ParseRequest(requestJWT)
-	if err != nil {
+	if err != nil || identity == nil {
 		return "", "", interaction.ErrInvalidCredentials
 	}
 
--- a/pkg/auth/dependency/interaction/flows/anonymous.go
+++ b/pkg/auth/dependency/interaction/flows/anonymous.go
@@ -101,3 +101,18 @@ func (f *AnonymousFlow) Authenticate(requestJWT string, clientID string) (*authn
 
 	return attrs, nil
 }
+
+func (f *AnonymousFlow) DecodeUserID(requestJWT string) (string, anonymous.RequestAction, error) {
+	identity, request, err := f.Anonymous.ParseRequest(requestJWT)
+	if err != nil {
+		return "", "", interaction.ErrInvalidCredentials
+	}
+
+	// Verify challenge token
+	purpose, err := f.Challenges.Consume(request.Challenge)
+	if err != nil || *purpose != challenge.PurposeAnonymousRequest {
+		return "", "", interaction.ErrInvalidCredentials
+	}
+
+	return identity.UserID, request.Action, nil
+}
--- a/pkg/auth/dependency/interaction/flows/webapp.go
+++ b/pkg/auth/dependency/interaction/flows/webapp.go
@@ -114,6 +114,59 @@ func (f *WebAppFlow) SignupWithLoginID(loginIDKey, loginID string) (*WebAppResul
 	}, nil
 }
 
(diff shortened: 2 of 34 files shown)


pkg/auth/dependency/webapp/authenticate_provider_impl.go
```diff
@@ -483,7 +492,7 @@ func (p *AuthenticateProviderImpl) LinkIdentityProvider(w http.ResponseWriter, r
 	q.Set("redirect_uri", r.URL.Path)
 	webappSSOState.SetRequestQuery(q.Encode())
 	state := sso.State{
-		Action: "link",
+		Action: action,
```
louischan-oursky, line 495: I think we need to fix the `userID` if it is promotion? The `userID` in this function is from `auth.GetSession`

pkg/auth/dependency/webapp/translation.go
```diff
@@ -25,10 +25,15 @@ var TemplateAuthUITranslationJSON = template.Spec{
...
 	"sign-in-apple": "Sign in with Apple",
+	"sign-up-apple": "Sign up with Apple",
 	"sign-in-google": "Sign in with Google",
+	"sign-up-google": "Sign up with Google",
 	"sign-in-facebook": "Sign in with Facebook",
+	"sign-up-facebook": "Sign up with Facebook",
 	"sign-in-linkedin": "Sign in with LinkedIn",
+	"sign-up-linkedin": "Sign up with LinkedIn",
 	"sign-in-azureadv2": "Sign in with Azure AD",
+	"sign-up-azureadv2": "Sign up with Azure AD",
```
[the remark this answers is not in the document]
kiootic: It is common to have different text to be consistent with the page (e.g. sign up with XXX for sign up page, log in with XXX for log in page), e.g. [Stack Overflow](https://stackoverflow.com/users/signup).

pkg/auth/dependency/webapp/authenticate_provider_impl.go
kiootic: Indeed, I didn't saw it when rebasing.

pkg/auth/dependency/webapp/translation.go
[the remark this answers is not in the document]
louischan-oursky: Fair enough.

pkg/auth/dependency/interaction/flows/webapp.go
```diff
@@ -114,6 +114,52 @@ func (f *WebAppFlow) SignupWithLoginID(loginIDKey, loginID string) (*WebAppResul
...
+	// Trigger OOB OTP if the first authenticator is OOB OTP
+	if len(s.Steps[0].AvailableAuthenticators) > 0 && s.Steps[0].AvailableAuthenticators[0].Type == authn.AuthenticatorTypeOOB {
+		step = WebAppStepSetupOOBOTP
+		err = f.Interactions.PerformAction(i, interaction.StepSetupPrimaryAuthenticator, &interaction.ActionTriggerOOBAuthenticator{
+			Authenticator: s.Steps[0].AvailableAuthenticators[0],
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
```
louischan-oursky, line 148: Line 134 - Line 148 can adjusted according to https://github.com/SkygearIO/skygear-server/pull/1407/commits/ac9091249d2b38c63173156a91a92c008df99add and https://github.com/SkygearIO/skygear-server/pull/1407/commits/0362d54743d12dd7031712c3b150b46ca427901a

pkg/auth/dependency/interaction/flows/anonymous.go
```diff
@@ -101,3 +101,18 @@ func (f *AnonymousFlow) Authenticate(requestJWT string, clientID string) (*authn
 
 	return attrs, nil
 }
+
+func (f *AnonymousFlow) DecodeUserID(requestJWT string) (string, anonymous.RequestAction, error) {
+	identity, request, err := f.Anonymous.ParseRequest(requestJWT)
```
carmenlau, line 106: Seems `Anonymous.ParseRequest` may return nil identity if kid not exists and key is provided in jwt? Maybe we should check and return `ErrInvalidCredentials` if identity is nil here.

pkg/auth/dependency/interaction/flows/webapp.go
louischan-oursky: Ping

PR #1407, opened 2020-05-14
Identity settings - Part 2

### LINKED ISSUE #1375: Account settings Auth UI page

louischan-oursky:
fix #1375

### DIFF
--- a/pkg/auth/dependency/interaction/flows/webapp.go
+++ b/pkg/auth/dependency/interaction/flows/webapp.go
@@ -29,13 +29,15 @@ func (f *WebAppFlow) LoginWithLoginID(loginID string) (*WebAppResult, error) {
 	s, err := f.Interactions.GetInteractionState(i)
 	if err != nil {
 		return nil, err
-	} else if s.CurrentStep().Step != interaction.StepAuthenticatePrimary {
+	}
+
+	if s.CurrentStep().Step != interaction.StepAuthenticatePrimary || len(s.CurrentStep().AvailableAuthenticators) <= 0 {
 		panic("interaction_flow_webapp: unexpected interaction state")
 	}
 
-	step := WebAppStepAuthenticatePassword
-	// Trigger OOB OTP if the first authenticator is OOB OTP
-	if len(s.CurrentStep().AvailableAuthenticators) > 0 && s.CurrentStep().AvailableAuthenticators[0].Type == authn.AuthenticatorTypeOOB {
+	var step WebAppStep
+	switch s.CurrentStep().AvailableAuthenticators[0].Type {
+	case authn.AuthenticatorTypeOOB:
 		step = WebAppStepAuthenticateOOBOTP
 		err = f.Interactions.PerformAction(i, interaction.StepAuthenticatePrimary, &interaction.ActionTriggerOOBAuthenticator{
 			Authenticator: s.CurrentStep().AvailableAuthenticators[0],
@@ -43,6 +45,10 @@ func (f *WebAppFlow) LoginWithLoginID(loginID string) (*WebAppResult, error) {
 		if err != nil {
 			return nil, err
 		}
+	case authn.AuthenticatorTypePassword:
+		step = WebAppStepAuthenticatePassword
+	default:
+		panic("interaction_flow_webapp: unexpected authenticator type")
 	}
 
 	token, err := f.Interactions.SaveInteraction(i)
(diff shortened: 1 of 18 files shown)


pkg/auth/dependency/interaction/flows/webapp.go
```diff
@@ -212,7 +240,7 @@ func (f *WebAppFlow) TriggerOOBOTP(token string, step interaction.Step) (*WebApp
 		panic("interaction_flow_webapp: unexpected interaction state")
 	}
 
-	err = f.Interactions.PerformAction(i, step, &interaction.ActionTriggerOOBAuthenticator{
+	err = f.Interactions.PerformAction(i, s.Steps[0].Step, &interaction.ActionTriggerOOBAuthenticator{
```
kiootic, line 243: I think it should be using `s.CurrentStep()` instead of `s.Steps[0]`?

pkg/auth/dependency/interaction/flows/webapp.go
```diff
@@ -230,6 +258,111 @@ func (f *WebAppFlow) TriggerOOBOTP(token string, step interaction.Step) (*WebApp
...
+
+		result = &WebAppResult{
+			Step: WebAppStepCompleted,
+		}
+
+		return
+	}
+
+	// Or have more steps to go through
+	step := WebAppStepSetupPassword
```
kiootic, line 343: Here we assume the next step is setting up authenticator, and the authenticator to setup is either password or OOB OTP.

It would be more robust (i.e. the interaction state would not be messed up) to assert `s.CurrentStep()` is in appropriate state (i.e. setup authenticator with correct available authenticator type).

pkg/auth/dependency/webapp/authenticate_provider_impl.go
```diff
@@ -410,6 +382,11 @@ func (p *AuthenticateProviderImpl) LinkIdentityProvider(w http.ResponseWriter, r
 
 	userID := auth.GetSession(r.Context()).AuthnAttrs().UserID
 
+	_, err = p.restoreState(r)
```
kiootic, line 385: I suppose this means an expired/invalid SID would fail the POST request and generate a new SID for the next request?

I think we need to reconsider whether to automatically re-generate the state in `persistState`, since now it has many important data that cannot be recovered if lost (e.g. interaction token, redirect URI).

louischan-oursky: We have a state middleware to detect expired/invalid SID and redirect the user to root if it is the case. Of course if we read the state twice we could have data race, e.g. the second read is considered expired.

pkg/auth/dependency/interaction/flows/webapp.go
louischan-oursky: Actually I would like to know by design, when would `getState` return more than 1 step?

If there are more than 1 step, the first step in the steps slice should be the next step the user should take, right? `CurrentStep()` return the last step in the steps slice.

Another question I encountered is that some of the code is now checking `len(s.Steps) != 1`, which may not be true in the future because `getState` might include more steps. But as long as the next step is expected, we should not panic.

kiootic: > Actually I would like to know by design, when would getState return more than 1 step?

`s.Steps` is supposed to be a list of executed steps + current step to execute. For example in a login interaction, it derives the steps like this:
1. If need primary authentication, add 'authenticate primary' to step list
2. If need primary authentication, and not yet performed it, return the current step list
3. If need secondary authentication, add 'authenticate secondary' to step list
4. If need secondary authentication, and not yet performed it, return the current step list
5. Add 'commit' to step list and return

So possible step list are:
- authenticate primary
- authenticate primary, authenticate secondary
- authenticate primary, authenticate secondary, commit

The last step in step list reflect `s.CurrentStep()`, i.e. the step that the flow should execute.

> If there are more than 1 step, the first step in the steps slice should be the next step the user should take, right? CurrentStep() return the last step in the steps slice.

The first step in the step list is the first step that the flow should execute. It may (length > 1) or may not (length = 1) been executed.

> Another question I encountered is that some of the code is now checking len(s.Steps) != 1, which may not be true in the future because getState might include more steps. But as long as the next step is expected, we should not panic.

These checks are mostly redundant, and should be replaced with `s.CurrentStep()`, since these checks occurs at the beginning of interaction, it is expected that there is exactly 1 step in the list.

pkg/auth/dependency/webapp/authenticate_provider_impl.go
kiootic: I suppose we can just read the state from store and inject it into context in middleware, so that `restoreState` would just read from the context and `persistState` would update state in context and write to store. Therefore, we would not re-generate expired/invalid SID anymore.

louischan-oursky: Let me open [a new issue](https://github.com/SkygearIO/skygear-server/issues/1409) to address this to avoid dragging this PR.

louischan-oursky:
commend addressed.

carmenlau:
> It seems removing last login ID would also remove the existing password authenticator. I suppose this behavior is okay.

Yes, removing identity will check and remove not used authenticator generically. So removing login id will cleanup related oob authenticator, and password authenticator will also be removed when removing last login id.

PR #1423, opened 2020-05-15
Fix API doc builds

### LINKED ISSUE #1422: Fix schemas
```
/home/travis/gopath/src/github.com/skygeario/skygear-server/pkg/auth/handler/oauth/challenge.go:58:1: invalid json schema: invalid character '"' after object key:value pair
```

We should build the API doc as part of CI pipeline for PR to avoid this in future.

kiootic:
fix #1422

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,7 @@ vendor:
 	go install github.com/golang/mock/mockgen
 	go install tools/nextimportslint.go
 	go install github.com/google/wire/cmd/wire
+	go install github.com/skygeario/openapi3-gen/cmd/openapi3-gen
 
 .PHONY: go-generate
 go-generate:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -51,6 +51,7 @@ vendor:
 	go install github.com/golang/mock/mockgen
 	go install tools/nextimportslint.go
 	go install github.com/google/wire/cmd/wire
+	go install github.com/skygeario/openapi3-gen/cmd/openapi3-gen
```
louischan-oursky, line 54: Could you also add `github.com/google/wire/cmd/wire` and `github.com/skygeario/openapi3-gen/cmd/openapi3-gen` to `devtools/devtools.go` so that the exact version is recorded in `go.mod`?

PR #1428, opened 2020-05-18
Do not implicitly re-generate web app state

### LINKED ISSUE #1409: Web App state expired/invalid handling
Original thread: https://github.com/SkygearIO/skygear-server/pull/1407#discussion_r424906280

---

Terry

> I suppose this means an expired/invalid SID would fail the POST request and generate a new SID for the next request?

> I think we need to reconsider whether to automatically re-generate the state in persistState, since now it has many important data that cannot be recovered if lost (e.g. interaction token, redirect URI).

Louis

> We have a state middleware to detect expired/invalid SID and redirect the user to root if it is the case. Of course if we read the state twice we could have data race, e.g. the second read is considered expired.

Terry

> I suppose we can just read the state from store and inject it into context in middleware, so that restoreState would just read from the context and persistState would update state in context and write to store. Therefore, we would not re-generate expired/invalid SID anymore.

louischan-oursky:
fix #1409


pkg/auth/dependency/webapp/state_provider.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// StateProvider does not implicitly generate new state for invalid or expired sid.
+// If sid is present, then it must be valid.
+type StateProvider interface {
+	// PersistError assumes id exists.
+	PersistError(id string, inputError error)
+	PersistState(r *http.Request, inputError error)
+	// MayRestoreState may return nil, nil.
+	MayRestoreState(r *http.Request) (state *State, err error)
```
kiootic, line 21: Actually, I wonder whether it's conventional to return `nil, nil` in Go? I seldom seen this behavior from my limitation experience with Go. I've written some code with this behavior before, and it seems getting confusing since we need to remember whether `nil` is possible for some function.

What's your opinion on this?

louischan-oursky:
Actually the entrypoint endpoint may also need to restore the state if the first step has error. Now I change the middleware to always generate a state if there is none. And StateProvider now requires sid to be present at all time.

kiootic:
> Actually the entrypoint endpoint may also need to restore the state if the first step has error.

Do we need to care about the error in the first step in subsequent steps? After the error occurred, user is prompt to correct the information provided and retry the action. At this point, the auth flow is conceptually 'restarted', so I suppose there should be no need to refer back to the first error?

louischan-oursky:
This is restricted by [the Post/Redirect/Get pattern](https://github.com/SkygearIO/skygear-server/issues/1288) we have to use. If the first POST fails, we have to generate a state to store the error and redirect the user to the same path. The GET request will restore the state and display the error. So the state has to be there even for the first step.

kiootic:
I see, let's consider 4 cases:
- GET entry step: may restore state for errors
- POST entry step: always regenerate state
- GET subsequent steps: must restore state
- POST subsequent steps: must restore state

Is that correct?

louischan-oursky:
Yes

pkg/auth/dependency/webapp/state_middleware.go
```diff
@@ -17,6 +17,23 @@ func (m *StateMiddleware) Handle(next http.Handler) http.Handler {
 
 		q := r.URL.Query()
 		sid := q.Get("x_sid")
+
+		if sid == "" && r.Method == "GET" {
```
kiootic, line 21: Are we going to always generate a SID on entry GET, not waiting for user clicking next button?

louischan-oursky: If we do it lazily, we need to bring back MayRestoreState.

GET: may restore state
POST entry step: always regenerate state
POST subsequent steps: must restore state

Doing it lazily is better IMO. Let's change it back?

kiootic: > Doing it lazily is better IMO. Let's change it back?

I agree. However, instead of having two function `MayRestoreStore` and `MustRestoreState`, can we have single `RestoreState` function and ignore not found error explicitly when needed?

pkg/auth/dependency/webapp/authenticate_provider_impl.go
```diff
@@ -122,7 +80,7 @@ func (p *AuthenticateProviderImpl) get(w http.ResponseWriter, r *http.Request, t
 		p.RenderProvider.WritePage(w, r, templateType, anyError)
 	}
 
-	state, err = p.restoreState(r)
+	state, err = p.StateProvider.RestoreState(r, true)
```
kiootic, line 83: I suppose we should check for `ErrStateNotFound` and ignore it.

PR #1450, opened 2020-05-25
Delete Auth APIs

### LINKED ISSUE #1434: Consider dropping Auth API
Since we are shifting the focus to Auth UI, Auth API is now a considerable maintenance burden. We are considering dropping Auth API to focus development effort on Auth UI.

kiootic:
fix #1434


kiootic:
`core/handler/factory.go` is still used by asset gear, so it cannot be removed for now.

PR #1468, opened 2020-06-01
Fix not handling OAuth errors

### LINKED ISSUE #1464: [Auth UI] Handle user quitting Facebook login flow
If user tries to connect Facebook, but clicks "Not Now" on Facebook login page, it will go to a page unable to load.

- Application Platform: Testing on client app

#### Expected Results
Should go back to connect account page.

#### Actual Results
It goes to a blank page.

#### Steps to reproduce
1. Connect Facebook
2. See Facebook login dialog, click "Not Now"
3. See it goes to a page unable to load

kiootic:
fix #1464

### DIFF
--- a/pkg/auth/dependency/webapp/authenticate_provider_impl.go
+++ b/pkg/auth/dependency/webapp/authenticate_provider_impl.go
@@ -559,7 +559,7 @@ func (p *AuthenticateProviderImpl) HandleSSOCallback(w http.ResponseWriter, r *h
 	writeResponse = func(err error) {
 		callbackURL := v.Get("redirect_uri")
 		if callbackURL == "" {
-			callbackURL = "/"
+			callbackURL = "/login"
 		}
 		sid := v.Get("x_sid")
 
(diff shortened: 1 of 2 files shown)


pkg/auth/dependency/webapp/authenticate_provider_impl.go
```diff
@@ -607,6 +605,12 @@ func (p *AuthenticateProviderImpl) HandleSSOCallback(w http.ResponseWriter, r *h
 		})
 	}
 
+	oauthError := r.Form.Get("error")
+	if oauthError != "" {
+		err = sso.NewSSOFailed(sso.SSOUnauthorized, "login failed")
```
louischan-oursky, line 610: I remember in the spec there is `error` and `error_description`. Maybe we should forward them to enable better diagnosis?

kiootic: `error` is an error code, therefore should not be displayed. I'll include `error_description` in the error message instead.

louischan-oursky: I guess it is not harmful to show the error code as well.

kiootic: The UX can be confusing to end-user, since the error code often contains technical terms.

louischan-oursky: OK. Let's revisit this later. I was thinking about using `<script>` to log the error code to console......

PR #1472, opened 2020-06-02
Allow logging in in anonymous user promotion flow

### LINKED ISSUE #1470: Allow logging in existing user in promotion flow
User may select an existing identity in promotion flow. Currently, it would fail with an identity exists error. Use case instead requires user to perform authentication and log in using the existing identity. The anonymous user is discarded in this case.

- [ ] Configuration flag (TBC): `on_identity_conflict.promotion`
    - `error`
    - `login`
- [ ] Hook events (TBC): `before_user_promote`, `after_user_promote`
    - `anonymous_user`
    - `user`
    - usually `anonymous_user` and `user` have same ID
    - when `on_identity_conflict.promotion` is `login`, `anonymous_user` and `user` may have different ID.

ref #1469

### LINKED ISSUE #1474: User is redirected to 404 in SSO promotion flow
If SSO promotion flow produce an error in OAuth flow, user would be redirected to 404 page.

SSO flow attempts to display the error by redirecting user to the callback URL. If user log in directly in Auth UI (not in OAuth flow), the callback URL is a path and redirection succeed. If user login using OAuth, the callback URL is a full URL (OAuth authorize endpoint) so the redirect URL is incorrect.

kiootic:
fix #1470 
fix #1474


pkg/auth/dependency/interaction/flows/webapp_sso.go
```diff
@@ -187,5 +187,5 @@ func (f *WebAppFlow) PromoteWithOAuthProvider(userID string, oauthAuthInfo sso.A
 		return nil, err
 	}
 
-	return f.afterAnonymousUserPromotion(result)
+	return f.afterAnonymousUserPromotion(i, result)
```
louischan-oursky, line 190: `PromoteWithOAuthProvider` should also handle `login` and `error` case.

PR #1476, opened 2020-06-03
Make login ID required error message more precise

### LINKED ISSUE #1466: [Auth UI] Error message for empty email field should not show "Username"
When clicking Next on email login without input, it will show an error message, but the error message says `Email or Username is required`, where the login field should be for email address only.

- Application Platform: Testing on client app

#### Expected Results
Should only show `Email is required`

#### Actual Results
See error message `Email or Username is required`, but user does not know what means by username

#### Steps to reproduce

1. Click connect email
2. Click Next directly without entering input 

#### Screenshot

IMG_0484

louischan-oursky:
fix #1466


louischan-oursky:
updated.

Issue #1482, opened 2020-06-04
Handle JWT time drift


kiootic:
We should gives a buffer for validation of `exp` and `iat` to account for small time drift (e.g. ~5s) between devices.

louischan-oursky:
Rick told me that browsers can detect such error when verifying TLS cert, showing message like "Your clock is skewed". If we need to follow this, we have wrap JWT verify function to translate the error.

PR #1484, opened 2020-06-04
Do not validate iat for JWT

kiootic:
ref #1482 

we may want to see if v4 of the JWT library supports handling time drifts.


kiootic:
`iat` should not be validated (jwt-go is going to remove that check in v4: https://github.com/dgrijalva/jwt-go/issues/314). For validation purpose, `nbf` should be used instead.

PR #1495, opened 2020-06-05
Remove old code

kiootic:


louischan-oursky:
Target another branch such as `authgear`? `master` should be kept for bug fix.

kiootic:
Oh yes, we should be using the `independence-refactor` branch for the refactor, or do you want to change the name?

louischan-oursky:
Any name is fine.

louischan-oursky:
I just rebase the base branch onto master and we have conflicts :(

kiootic:
Let me rebase...

.github/workflows/test.yaml
```diff
@@ -0,0 +1,32 @@
...
+          go-version: 1.13.3
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Get dependencies
+        run: make vendor
+      - name: Test
+        run: |
+          make generate
+          go mod tidy
+          (set +e; git status --porcelain | grep '.*'; test $? -eq 1)
```
louischan-oursky, line 30: Move line 28, 29 and 30 to Makefile as a new target.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,32 @@
...
+          make lint
+          make test
```
louischan-oursky, line 32: We should run in this order

- `make lint`
- `make test`
- `make generate; go mod tidy; (grep ...)`

And failing one should not prevent following commands from running. However, any failure should fail the whole job.

kiootic:
I'll add back a Dockerfile and make target, but I think image building should be part of deployment flow, so no need to build in workflow now.

PR #1496, opened 2020-06-05
Do not show 0s for cooldown label

### LINKED ISSUE #1493: [Auth UI] Should not show 0s for resend count down
Upon reaching 0s, the resend verification code countdown will just stay as 'Resend (0s)' but not 'Resend'.
- Application Platform: Tested on Rees

#### Expected Results
Should be just 'Resend' when reached 0s

#### Actual Results
It shows Resend (0s)' but not 'Resend'

#### Steps to reproduce
1. Open signup / login page
2. Select phone number / email to reach verification page
3. Wait until the count down to become 0s

#### Screenshot
image

louischan-oursky:
fix #1493

### DIFF
--- a/static/authui/js/main.js
+++ b/static/authui/js/main.js
@@ -144,6 +144,8 @@ window.addEventListener("load", function() {
 
     var scheduledAt = new Date();
     var cooldown = parseInt(el.getAttribute("data-cooldown"), 10) * 1000;
+    var label = el.getAttribute("data-label");
+    var labelUnit = el.getAttribute("data-label-unit");
     var token;
     var tick = function() {
       var now = new Date();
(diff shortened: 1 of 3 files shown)


static/authui/js/main.js
```diff
@@ -158,9 +160,14 @@ window.addEventListener("load", function() {
...
-      el.textContent = (el.textContent || "").replace(/\d+/, String(displaySeconds));
+      if (displaySeconds === 0) {
+        el.textContent = label;
+      } else {
+        el.textContent = labelUnit.replace("%d", String(displaySeconds));
+      }
     }
-    token = setInterval(tick, 500);
+    // 24fps => 1 / 24 * 1000 = 42ms
+    token = setInterval(tick, 42);
```
kiootic, line 170: Actually, why do we need to use smaller interval?

louischan-oursky: 0.5s sometimes does not look smooth.

kiootic: Let's use `requestAnimationFrame` instead?

louischan-oursky: `requestAnimationFrame` is a bit over kill.

kiootic: I don't think it's overkill. The issue we have at hand is 'countdown animation is not smooth', so using a proper animation API to update it is quite appropriate.

PR #1504, opened 2020-06-09
OAuth response mode

### LINKED ISSUE #1503: OAuth response mode
Client may use XHR to interact with Auth UI. Custom URL scheme redirect does not work with XHR. Therefore, we need an extra endpoint to use HTTP 302 redirect. That endpoint can also support `response_mode=form_post`.

louischan-oursky:
fix #1503

### DIFF
--- a/pkg/auth/dependency/oauth/handler/response_mode.go
+++ b/pkg/auth/dependency/oauth/handler/response_mode.go
@@ -1,0 +1,106 @@
+package handler
+
+import (
+	"fmt"
+	"html/template"
+	"net/http"
+	"net/url"
+
+	coreurl "github.com/skygeario/skygear-server/pkg/core/url"
+)
+
+const htmlRedirectTemplateString = `<!DOCTYPE html>
+<html>
+<head>
+<meta http-equiv="refresh" content="0;url={{ .redirect_uri }}" />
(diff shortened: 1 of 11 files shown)


pkg/auth/dependency/oauth/handler/response_mode.go
```diff
@@ -0,0 +1,105 @@
...
+window.location.href = "{{ .redirect_uri }}"
+</script>
+</body>
+</html>
+`
+
+const formPostTemplateString = `<!DOCTYPE html>
+<html>
+<head>
+<title>Submit this form</title>
```
kiootic, line 28: For better UX, we can use localized app name?

louischan-oursky: 90% people wont see this. When they see this, what they need to do is to submit this form. But localizing this string has some overhead.

kiootic: I think at least this message should be moved to body (and may need localization in future), since users seldom pay attention to title.

louischan-oursky: Move the message as `<noscript>` to `<body>`

Issue #1497, opened 2020-06-05
Migrate migration tool


kiootic:
We want a migration tool that can keep history of applied migrations. Let's investigate existing migration tools features.

kiootic:
some candidates:
- https://github.com/golang-migrate/migrate (used currently)
- https://github.com/pressly/goose
    - support hybrid versioning to handle merging
    - has migration history
    - does not apply migrations in gap and no error produced
- https://github.com/rubenv/sql-migrate
    - has migration history
    - does not apply migrations in gap and no error produced
- https://github.com/amacneil/dbmate
    - has migration history
    - applies all un-applied migrations, even in past gap

Issue #1477, opened 2020-06-03
authgear


louischan-oursky:
- Gateway
  - Remove gateway code
- Asset Gear
  - Move its code to a new repository
  - Remove the dependency on tenant config
  - Introduce master key as an env var
  - Introduce secret as an env var
  - Extract the upload code as a recipe (instead of making a new SDK)
- Auth Gear
  - API changes
    - Rename `/_auth/session/resolve` to something else
    - Remove `skygear_user` in ID token claims
    - Add the new claim in full URL form to indicate `is_anonymous`
    - Add the new claim in full URL form to indicate `metadata`
    - Remove `session` in hook context
    - Remove the feature disable user
    - Remove the feature verification
    - Remove master key
  - Internal changes
    - Remove asset gear loader in template engine
    - Remove the dependency on libvips
    - Fix last login at
    - Remove master key from tenant config
    - Remove irrelevant files from the repository
    - Migrate CI/CD to GitHub Actions
  - Database schema
    - Consolidate migrations
    - Rename _core_user to _auth_user
    - Merge _auth_user_profile into _auth_user
    - Consider switch to a migration tool that keeps full history of executed migrations, not just the latest version.
  - Relocation
    - Move it to authgear/authgear
  - Documentation
    - Move skygeario/features to /features
    - Move skygeario/guides to authgear/docs
    - Rewrite authgear/docs to be more relevant
  - SDK
    - Rename the packages
    - Move it to authgear/authgear-sdk-js
    - Delete generation of IIFE bundle
    - Delete examples
    - Delete observer
    - Delete asset gear container
    - Delete node client
    - Delete the existing docs site
    - DocFX v3 supports versioning but it is not released yet
    - Revamp the current API
    - Emit `onAccessTokenExpired` (single handler)
    - Emit `onRefreshTokenExpired` (single handler)
    - Register default `onAccessTokenExpired` handler to refresh access token
  - Future works
    - Support k8s CRD config backend
    - Introduce authgear/helm-chart
    - Write an example app and use authgear/helm-chart to deploy
    - Secret Management (Needs design)
    - Multi-tenant by hostname
    - Move database connection URI to envvar?
    - Integrate OpenTelemetry

kiootic:
- Do we want to cleanup the `core` packages and remove irrelevant stuff/merge into `auth` package/new repo?
- Do we want to do #1433 as part of refactor? (i.e. use library instead of hand-rolled OAuth implementation)
[...]

PR #1505, opened 2020-06-10
Merge user & profile tables

kiootic:
ref #1477 
ref #1497


migrate/migrations/20200610183245-initial_tables.sql
```diff
@@ -0,0 +1,132 @@
...
+CREATE TABLE _auth_oauth_authorization
+(
+    id         text PRIMARY KEY,
+    app_id     text                        NOT NULL,
+    client_id  text                        NOT NULL,
+    user_id    text                        NOT NULL REFERENCES _auth_user (id),
+    created_at timestamp without time zone NOT NULL,
+    updated_at timestamp without time zone NOT NULL,
+    scopes     jsonb                       NOT NULL,
+    UNIQUE (app_id, user_id, client_id)
```
louischan-oursky, line 33: Explicitly create **EACH** unique index with `CREATE UNIQUE INDEX` so that the name is deterministic. 

We should introduce the down migration as well. In the down migration we can drop the unique index first.

migrate/migrations/20200610183245-initial_tables.sql
```diff
@@ -0,0 +1,132 @@
...
+
+CREATE SEQUENCE _auth_event_sequence;
+
+CREATE TABLE _auth_password_history
+(
+    id        text PRIMARY KEY,
+    app_id    text                        NOT NULL,
+    user_id   text                        NOT NULL REFERENCES _auth_user (id),
+    password  text                        NOT NULL,
+    logged_at timestamp without time zone NOT NULL
```
louischan-oursky, line 21: Maybe rename `logged_at` to `created_at`.

cmd/auth/main.go
```diff
@@ -135,7 +133,6 @@ func main() {
 		ReservedNameChecker:      reservedNameChecker,
 	}
 
-	task.AttachVerifyCodeSendTask(asyncTaskExecutor, authDependency)
```
louischan-oursky, line 138: Remove the configuration as well.

kiootic: I planned to cleanup & refactor the configuration (config/tenant.go) as a separate task.

migrate/migrations/20200610183245-initial_tables.sql
kiootic: I don't think we need to drop the unique indexes explicitly, since dropping the table would drop them too?

pkg/core/config/tenant_test.go
```diff
@@ -113,7 +113,7 @@ func makeFullTenantConfig() TenantConfiguration {
 		AppName:    "myapp",
 		AppID:      "66EAFE32-BF5C-4878-8FC8-DD0EEA440981",
 		DatabaseConfig: &DatabaseConfiguration{
-			DatabaseURL:    "postgres://user:password@localhost:5432/db?sslmode=disable",
+			DatabaseURL:    "postgres://user:password@localhost:5432/migrate?sslmode=disable",
```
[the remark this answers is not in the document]
kiootic: IDE auto-refactor is too powerful -_-

migrate/migrations/20200610183245-initial_tables.sql
louischan-oursky: Let's add `down` command first and we can verify whether unique indexes are dropped along with the tables

kiootic: I just tried and indexes are dropped along with the tables.

Issue #1506, opened 2020-06-11
Refactor configuration


kiootic:
- remove msgpack
- use https://github.com/kubernetes-sigs/yaml for yaml
- use https://github.com/iawaknahc/jsonschema for validation
- app_id optional (default constant in single teanant)
- hostnames in config (merge from VALID_HOSTS)
- watch config (directory)
- delete version
- move db config
    - db, redis -> credentials config
- delete master key & access key
- delete removed feature config
- local development mode flag (use viper & cobra?)
- delete default smtp etc config
- enable file loader by default
- ```go
   type PrivateConfig struct {
	Secrets []SecretDef
	Credentials []CredentialDef
    }
   type AppConfig struct {
	ID string
	Hosts []string
	// ...
   }
   ```

PR #1509, opened 2020-06-15
Add refactored config struct shapes

kiootic:
ref #1506

### DIFF
--- a/pkg/auth/config/authenticator.go
+++ b/pkg/auth/config/authenticator.go
@@ -1,0 +1,57 @@
+package config
+
+type AuthenticatorConfig struct {
+	Password     *AuthenticatorPasswordConfig     `json:"password,omitempty"`
+	TOTP         *AuthenticatorTOTPConfig         `json:"totp,omitempty"`
+	OOB          *AuthenticatorOOBConfig          `json:"oob_otp,omitempty"`
+	BearerToken  *AuthenticatorBearerTokenConfig  `json:"bearer_token,omitempty"`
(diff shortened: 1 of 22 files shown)


pkg/auth/config/ui.go
```diff
@@ -0,0 +1,16 @@
+package config
+
+type UIConfig struct {
+	CustomCSS          string                      `json:"custom_css,omitempty"`
+	CountryCallingCode *UICountryCallingCodeConfig `json:"country_calling_code,omitempty"`
+	Localization       *UILocalizationConfig       `json:"localization,omitempty"`
```
louischan-oursky, line 6: Move Localization to the root.

pkg/validation/schema.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (s *SimpleSchema) ValidateReader(r io.Reader) error {
+	return convertErrors(validateSchema(s.col, r))
+}
+
+type MultipartSchema struct {
+	mainPartID string
+	parts      map[string]interface{}
+	col        *jsonschema.Collection
```
louischan-oursky, line 31: We are not utilizing `$anchor` and the first argument of `Apply`. So we have to build a MultipartSchema for each schema we want to validate against, which may become tedious.

pkg/auth/config/authenticator.go
```diff
@@ -0,0 +1,57 @@
+package config
+
+type AuthenticatorConfig struct {
+	Password     *AuthenticatorPasswordConfig     `json:"password,omitempty"`
```
louischan-oursky, line 4: How do we deal with default values?

PR #1513, opened 2020-08-12
Escape asset name in S3 asset store

Steven-Chan:


pkg/server/asset/s3.go
```diff
@@ -136,17 +137,18 @@ func (s *s3Store) PutFileReader(
 // GeneratePostFileRequest return a PostFileRequest for uploading asset
 func (s *s3Store) GeneratePostFileRequest(name string, contentType string, length int64) (*PostFileRequest, error) {
 	return &PostFileRequest{
-		Action: "/files/" + name,
+		Action: "/files/" + url.PathEscape(name),
```
louischan-oursky: I am not sure whether we should pass the path-escaped key when we interact with the Go SDK of AWS. The SDK may expect unescaped key.

Line 140 and line 149 definitely need manual escape because they are not interacting with the AWS SDK.

louischan-oursky (commented):
Not sure whether we should merge it into the `v1-branch`. I suggest we create a new branch, say `v1.7.3-fix`, based on the tag `v1.7.3` and use that branch as the target branch of this PR.

pkg/server/asset/s3.go
louischan-oursky: Second thought. If line 140 used to have unescaped URL, file with space still can be uploaded successfully. Maybe we need only to change line 149.

Steven-Chan: Why not merge into `v1-branch`?

louischan-oursky: Seems like `v1-branch` is identical to `v1.7.3`. Sorry, forgot what I just said.

Steven-Chan: I don't think the Go AWS SDK handles key escaping, the whole repo does not use `url.PathEscape` once.

Steven-Chan: > Seems like v1-branch is identical to v1.7.3. Sorry, forgot what I just said.

I was thinking we should release a `v1.7.4`.

louischan-oursky: > the whole repo does not use url.PathEscape once.

We did use `url.PathEscape` in `pkg/server/asset/cloud.go` but we did not use Go AWS SDK there so no conclusion can be made. 

> I don't think the Go AWS SDK handles key escaping

Then I wonder why we can upload file with name having spaces. The observed problem now is incorrectly generated URL, which should be related to line 149.

Sorry I am a bit conservative because we do not have tests to verify our changes :(

Steven-Chan: > We did use url.PathEscape in pkg/server/asset/cloud.go but we did not use Go AWS SDK there so no conclusion can be made.

I meant the AWS Go SDK repo, not ours.

louischan-oursky: Not sure if it is related. https://github.com/aws/aws-sdk-go/issues/1789

